PR #24, opened 2026-04-18
docs(adr): ADR-0003 — provider-refresh substrate

madeye:
## Summary
- Unifies the fetch + cache + interval-refresh machinery shared by rule-providers (M1.D-5) and proxy-providers (M1.H-1) behind one `ProviderSource` primitive plus a minimal `RefreshTarget` trait.
- Resolves the `// TODO: unify with M1.D-5` marker that `docs/specs/proxy-providers.md` §Non-goals calls out, before any provider code is written.
- Codifies the failure policy (last-good retention, atomic cache write, panic-abort on the refresh task) against ADR-0002 Class A/B rows.
- Keeps payload-specific storage deliberately un-unified: rule-providers keep `Arc<ArcSwap<RuleSet>>` for wait-free match reads, proxy-providers keep `Arc<RwLock<Vec<Arc<dyn ProxyAdapter>>>>` for multi-step rebuilds (see §Alternatives A.2).

Blocks M1.D-5 (task #16) and M1.H-1 (task #15) engineer work.

## Test plan
- [ ] §Decision splits "shared substrate" vs "per-type" correctly — rule-providers' `ArcSwap` and proxy-providers' health-map prune stay inside the concrete impls, only fetch/cache/interval/panic-abort are unified.
- [ ] §5 failure-policy table matches the dispositions already in `docs/specs/rule-provider-upgrade.md` and `docs/specs/proxy-providers.md`. Class A/B cites per ADR-0002.
- [ ] §6 ETag deferral is acceptable for M1 (recommended intervals are 1h–24h; bandwidth cost is bytes-per-day).
- [ ] §7 single-file placement in `mihomo-config/src/provider_source.rs` is acceptable vs a new crate.
- [ ] §Migration describes how M1.D-5 consumes today's `rule_provider.rs` fetch logic (extract → promote to atomic write → wrap in `RuleProvider`) without breaking the spec-approved `mrs` parser scope.

### DIFF
--- a/docs/adr/0003-provider-refresh-substrate.md
+++ b/docs/adr/0003-provider-refresh-substrate.md
@@ -1,0 +1,393 @@
+# ADR 0003: Unified provider-refresh substrate for rule and proxy providers
+
+- **Status:** Proposed (architect 2026-04-18, awaiting pm + engineer review)
+- **Date:** 2026-04-18
+- **Author:** architect
+- **Supersedes:** —
+- **Related:** roadmap M1.D-5 (rule-provider upgrade), M1.H-1 (proxy-providers),
+  M1.G-5 / M1.G-6 (providers REST endpoints),
+  [`docs/specs/rule-provider-upgrade.md`](../specs/rule-provider-upgrade.md),
+  [`docs/specs/proxy-providers.md`](../specs/proxy-providers.md),
+  [ADR-0002](0002-upstream-divergence-policy.md) (divergence classification)
+
+## Context
+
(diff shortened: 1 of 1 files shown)


madeye:
✅ Solid unification of fetch+cache+refresh substrate. Separation of shared (fetch/cache/loop) vs per-type (ArcSwap/RwLock) is correct. Failure table is well-classified and panic-abort requirement is explicit. Migration path clear for both M1.D-5 and M1.H-1. LGTM.

PR #25, opened 2026-04-18
docs(adr): ADR-0005 — runtime event broadcaster

madeye:
## Summary
- Ratifies `tokio::sync::broadcast` (not `watch` / `mpsc` / a shared enum bus) as the fan-out primitive for WebSocket streaming endpoints.
- Clarifies that `/memory` is a per-connection sampler and `DELETE /connections` is a direct call — neither lives on the broadcaster. Avoids speculative "runtime event bus" plumbing.
- Codifies subscriber lifecycle, drop-oldest-per-subscriber backpressure (matches the already-approved `api-logs-websocket.md`), and — most importantly — the panic-abort contract: a process-wide `std::panic::set_hook` that aborts on any panic (sync tracing layer, async task, blocking helper). Forbids `tower::CatchPanicLayer` per the QA invariant recorded in memory `feedback_api_no_catch_panic.md`.

Blocks engineer task #17 (M1.G-3/4/7/8/9 WebSocket log stream + memory + DELETE /connections + /dns/query + cache flush).

## Test plan
- [ ] §2/§3 scope: agree that `/memory` and `DELETE /connections` are NOT broadcaster consumers. Would flattening them into a unified bus (Alternative A.1) buy anything I missed?
- [ ] §5 backpressure: drop-oldest over slow-client-disconnect matches dashboard UX needs (debugger-paused clients don't get kicked). QA to confirm this doesn't break any planned test for lag behaviour.
- [ ] §6 panic-abort: the `std::panic::set_hook` → `std::process::abort()` is broader than tokio's `unhandled_panic = Shutdown` — required because tracing layers and blocking helpers run outside tokio tasks. Confirm this matches QA's soak-test panic-abort invariant.
- [ ] §8 divergence classes: lag-vs-disconnect = Class B; no-CatchPanic = Class A (silent bug > loud crash).

### DIFF
--- a/docs/adr/0005-runtime-event-broadcaster.md
+++ b/docs/adr/0005-runtime-event-broadcaster.md
@@ -1,0 +1,371 @@
+# ADR 0005: Runtime event broadcaster for WebSocket streaming endpoints
+
+- **Status:** Proposed (architect 2026-04-18, awaiting pm + engineer + qa review)
+- **Date:** 2026-04-18
+- **Author:** architect
+- **Supersedes:** —
+- **Related:** roadmap M1.G-3 (`GET /logs` websocket), M1.G-4
+  (`GET /memory` websocket), M1.G-7 (`DELETE /connections` bulk),
+  [`docs/specs/api-logs-websocket.md`](../specs/api-logs-websocket.md)
+  (approved), `docs/specs/api-delay-endpoints.md`,
+  [ADR-0002](0002-upstream-divergence-policy.md),
+  memory `feedback_api_no_catch_panic.md` (QA panic-abort invariant)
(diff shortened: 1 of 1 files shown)


madeye:
✅ Clean architectural split: broadcast for /logs, per-connection sampler for /memory, direct call for DELETE /connections. Panic-abort via std::panic::set_hook covers sync tracing layers — approved. Lag→visible-frame (Class B) + panic-abort (Class A) well-classified. No CatchPanicLayer. QA audit: confirm 128-msg capacity from api-logs-websocket.md is M1 plan and lag-visibility backpressure shape matches soak-test expectations.

PR #26, opened 2026-04-18
docs(adr): ADR-0004 — metrics cardinality policy

madeye:
## Summary
- Introduces Class I (static enum) / II (config-bounded, capped at 1024) / III (unbounded, forbidden) classification so reviewers can approve or reject label sets with a three-way lookup instead of re-deriving the policy per metric.
- Settles the three questions PM flagged as blockers on test plan #30: cardinality cap (1024 distinct values per Class II label per scrape, with `mihomo_metric_truncated_total` observable overflow counter), label-value sanitisation (reject empty, replace control chars, no case normalisation, no length cap), and scrape auth (same `require_auth` Bearer as all REST — no `?token=` shortcut).
- Ratifies the per-scrape fresh-registry exposition model from `docs/specs/metrics-prometheus.md` and forbids global static registries. Exposition format is text 0.0.4 only for M1; OpenMetrics and gzip deferred to M2.
- Adds four operational counters (`mihomo_metric_truncated_total`, `_skipped_total`, `_sanitised_total`, `_conflict_total`) so metric-pipeline degradation is visible.

Unblocks PM task #30 (metrics test plan) and engineer task #18 (M1.H-2).

## Test plan
- [ ] §1 classification: `proxy_name` on `mihomo_proxy_alive` is Class II; `rule_type`/`action` on `mihomo_rules_matched_total` are Class I. Any metric in the approved catalog that I've misclassified?
- [ ] §1 Class II: 1024 cap with truncation counter — does this match what you want to test for? If you'd prefer a different number (4096?), flag and I'll amend.
- [ ] §2 sanitisation: empty label → skip + `mihomo_metric_skipped_total`. Control char → `<sanitised>` literal + `mihomo_metric_sanitised_total`. Confirm these don't break any planned test.
- [ ] §4 auth: no `/metrics` secret bypass. Consistent with ADR-0005 §4 WS `?token=` being WS-upgrade-only.
- [ ] §7 checklist: reviewers use this at PR-time. QA may want test-plan bullets for the truncation / sanitisation / conflict counters — fine by me, those are Class I metrics with a known shape.

### DIFF
--- a/docs/adr/0004-metrics-cardinality.md
+++ b/docs/adr/0004-metrics-cardinality.md
@@ -1,0 +1,370 @@
+# ADR 0004: Prometheus metrics cardinality and exposition policy
+
+- **Status:** Proposed (architect 2026-04-18, awaiting pm + qa review)
+- **Date:** 2026-04-18
+- **Author:** architect
+- **Supersedes:** —
+- **Related:** roadmap M1.H-2 (Prometheus `/metrics`),
+  [`docs/specs/metrics-prometheus.md`](../specs/metrics-prometheus.md) (approved),
+  [ADR-0002](0002-upstream-divergence-policy.md) (divergence policy)
+
+## Context
+
+`docs/specs/metrics-prometheus.md` is approved and defines the concrete
+catalog for M1 (`mihomo_traffic_bytes`, `mihomo_connections_active`,
+`mihomo_proxy_alive`, `mihomo_proxy_delay_ms`, `mihomo_rules_matched_total`,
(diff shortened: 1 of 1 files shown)


madeye:
✅ Answers all three blocking questions: 1024-cap + truncation, sanitisation rules, Bearer-token auth. Class I/II/III classification is clear and enforceable. Per-scrape fresh registry + &'static str-only hot path correct. Four diagnostic counters (truncated/skipped/sanitised/conflict) well-placed. Checklist for new metrics is solid. Minor: expand §1 Class II unique-constraint to show dedup/last-write-wins code shape for collision handling.

Issue #8, opened 2026-04-11
CI: wire v2ray_plugin_integration tests into test.yml


madeye:
The test suite `crates/mihomo-proxy/tests/v2ray_plugin_integration.rs` (2 tests, covers the built-in v2ray-plugin WS+TLS transport added in b3e3b81) is not executed in CI.

`cargo test --lib` does not pick up `tests/*.rs` integration binaries — each needs an explicit `cargo test --test <name>` step, the same way `trojan_integration` and `shadowsocks_integration` are wired.

**Fix:** add to `.github/workflows/test.yml` under the `test` job:

```yaml
- name: v2ray-plugin integration tests
  run: cargo test --test v2ray_plugin_integration -- --nocapture
```

Priority: P0 — the feature ships untested in CI.

See docs/ci-status.md for the full gap report.

madeye:
Already wired in `.github/workflows/test.yml` — step "v2ray-plugin integration tests" in both the `test` and `macos` jobs.

Issue #9, opened 2026-04-11
CI: wire pre_resolve_test into test.yml


madeye:
The test suite `crates/mihomo-tunnel/tests/pre_resolve_test.rs` (2 tests, covers the DNS pre-resolve path used by rule matching) is not executed in CI. `cargo test --lib` does not pick up integration binaries.

**Fix:** add to `.github/workflows/test.yml` under the `test` job:

```yaml
- name: Pre-resolve tests
  run: cargo test --test pre_resolve_test -- --nocapture
```

Priority: P0.

See docs/ci-status.md for the full gap report.

madeye:
Already wired in `.github/workflows/test.yml` — step "DNS pre-resolve tests" in both the `test` and `macos` jobs.

PR #31, opened 2026-04-18
feat(rules): complete rule parser for M1.D-1/D-3/D-6

madeye:
## Summary

- Wires eight previously-unregistered rule types into `parse_rule`: `IN-PORT`, `DSCP`, `UID`, `SRC-GEOIP`, `PROCESS-PATH`, `DOMAIN-WILDCARD`, `IP-SUFFIX`, `IP-ASN` (plus `SRC-IP-SUFFIX` / `SRC-IP-ASN`). Each silently hit the `unknown rule type` arm before — configs relying on them were silently misrouted (Class A per ADR-0002).
- New rule modules: `domain_wildcard`, `ip_suffix`, `ip_asn`. `ParserContext` grows `asn: Option<Arc<maxminddb::Reader<Vec<u8>>>>`; missing reader hard-errors at parse time with a discovery-path hint.
- `mihomo-config` scans rules for `GEOIP`/`SRC-GEOIP` and `IP-ASN`/`SRC-IP-ASN`, lazy-loading both MMDBs via the standard discovery chain (`$XDG_CONFIG_HOME/mihomo/`, `$HOME/.config/mihomo/`, `./mihomo/`). `process_path` prefix mode now matches on path-component boundaries (fixes a latent test; aligns with spec intent).

Implements task #8. Follows `docs/specs/rules-parser-completion.md` (approved) and `docs/specs/rules-parser-completion-test-plan.md`.

Upstream citations per ADR-0002 are inline in each new rule module. Class B divergences (PROCESS-PATH prefix match, UID non-Linux always-false) are documented in the module-level doc comments.

Out of scope (tracked separately):
- SRC-GEOIP and IP-ASN fixture-DB-backed integration tests — flagged in test plan §Issue 2; fixtures not yet in repo. Present coverage exercises the parser dispatch and missing-reader error paths.
- GEOSITE (M1.D-2), rule-provider upgrade (M1.D-5), SUB-RULE (M1.D-7) — separate specs.

## Test plan

- [x] `cargo test -p mihomo-rules --lib` — 65 passed (up from 18)
- [x] `cargo test -p mihomo-rules --test rules_test` — 95 passed (up from 79)
- [x] `cargo test -p mihomo-config --lib` — 32 passed (up from previous; adds ASN discovery-chain tests)
- [x] `cargo build --workspace` — clean
- [x] `cargo clippy --workspace --all-targets` — clean
- [ ] QA to verify with fixture MMDBs (see Issue 2 in test plan) once committed

### DIFF
--- a/crates/mihomo-config/src/lib.rs
+++ b/crates/mihomo-config/src/lib.rs
@@ -208,36 +208,51 @@ pub fn rebuild_from_raw_with_cache_dir(
     Ok((proxies, rules))
 }
 
-/// Scan `raw.rules` for any `GEOIP` entry; if present, lazy-load the Country
-/// MMDB from the default path and build a `ParserContext` carrying the reader.
-/// Fail-fast (returning an error that names the offending rule and the path
-/// we tried) when the scan matches but the load fails.
+/// Scan `raw.rules` for any GeoIP-backed entry (`GEOIP`, `SRC-GEOIP`) or any
+/// ASN-backed entry (`IP-ASN`, `SRC-IP-ASN`); if present, lazy-load the
+/// corresponding MMDB from the default path and build a `ParserContext`
+/// carrying the readers. Fail-fast (returning an error that names the
+/// offending rule and the path we tried) when the scan matches but the
+/// load fails.
 fn build_parser_context(
     raw: &raw::RawConfig,
 ) -> Result<mihomo_rules::ParserContext, anyhow::Error> {
-    build_parser_context_at(raw, default_geoip_path())
+    build_parser_context_at(raw, default_geoip_path(), default_asn_path())
 }
 
 /// Same as [`build_parser_context`] but lets the caller override the mmdb
-/// path — used by tests to avoid depending on the user's `$HOME`.
+/// paths — used by tests to avoid depending on the user's `$HOME`.
 fn build_parser_context_at(
     raw: &raw::RawConfig,
-    path: PathBuf,
+    geoip_path: PathBuf,
+    asn_path: PathBuf,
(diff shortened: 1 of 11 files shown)


madeye:
Post-merge architectural review (second pass, per team-lead request):

No trait-level changes in this PR — it stays within the existing `Rule` contract. Three architectural notes worth capturing:

**1. `ParserContext` now carries two `Option<Arc<maxminddb::Reader<Vec<u8>>>>` fields (GEOIP + ASN).** That's fine for M1, but the discovery-chain logic is duplicated — once for Country.mmdb in `mihomo-config/src/lib.rs` and once for ASN.mmdb. When GEOSITE lands (M1.D-2, task #9) a third `Option<Arc<...>>` joins the party. I'd prefer not to ship three duplicated discovery helpers — suggest engineer on task #9 extracts a single `fn locate_mmdb(basename: &str, required: bool) -> Option<PathBuf>` helper when they add the geosite path. No churn on this PR needed.

**2. `process_path` prefix mode now matches on path-component boundaries.** Good fix. Worth confirming the boundary rule for Windows paths (`\` separator) if we ever port there — but that's out of scope for M1 (Windows is in `docs/vision.md` §Non-goals territory: transparent-proxy doesn't target Windows anyway).

**3. Class B divergences documented in module docs.** Checked `uid.rs` and `process_path.rs`: both cite upstream + note the divergence inline, matching the ADR-0002 convention. Consistent with the rest of the codebase.

No architectural concerns. The only cleanup suggestion is the mmdb discovery-helper dedup, which is M1.D-2 engineer's call when that task starts.

PR #33, opened 2026-04-18
feat(rules): SUB-RULE named rule subsets (M1.D-7)

madeye:
## Summary
- Implements SUB-RULE dispatch (M1.D-7): named rule subsets referenced via `SUB-RULE,<name>`.
- Adds `SubRuleRule` in mihomo-rules wrapping `Arc<Vec<Box<dyn Rule>>>` so blocks share a single Arc.
- Extends `Rule` trait with `match_and_resolve(...) -> Option<String>`; match_engine switches to the new API so dispatch into sub-rules is transparent to tunnel.
- mihomo-config `sub_rules_parser`: DFS cycle detection with path-aware errors, topological build order, hard-error on undefined block (Class A), warn-once on empty block (Class B).

Originally opened as PR #32 (stacked on PR #31). Closed automatically when #31 merged with --delete-branch; this PR is the rebased-onto-main replacement.

## Test plan
- [x] 13 unit tests in sub_rule.rs (A1-A6, B1-B3)
- [x] 7 sub_rules_parser tests
- [x] 9 end-to-end config tests (undefined, cycles, diamond-negative, inner-match, fall-through, forward-ref, empty-block)
- [x] Local gates: cargo fmt --check, clippy -D warnings, cargo test --lib

### DIFF
--- a/crates/mihomo-api/tests/api_test.rs
+++ b/crates/mihomo-api/tests/api_test.rs
@@ -12,30 +12,13 @@ use tower::ServiceExt;
 
 fn test_raw_config() -> RawConfig {
     RawConfig {
-        port: None,
-        socks_port: None,
         mixed_port: Some(7890),
-        allow_lan: None,
-        bind_address: None,
         mode: Some("rule".into()),
-        log_level: None,
-        ipv6: None,
-        external_controller: None,
-        secret: None,
-        dns: None,
-
-        proxies: None,
-        proxy_groups: None,
         rules: Some(vec![
             "DOMAIN,example.com,DIRECT".into(),
             "MATCH,REJECT".into(),
         ]),
-        rule_providers: None,
-        subscriptions: None,
-        tproxy_port: None,
-        tproxy_sni: None,
-        routing_mark: None,
-        hosts: None,
+        ..Default::default()
     }
 }
 
--- a/crates/mihomo-app/tests/systemd_config_test.rs
+++ b/crates/mihomo-app/tests/systemd_config_test.rs
@@ -4,29 +4,13 @@ use std::os::unix::fs::PermissionsExt;
 
 fn minimal_raw_config() -> RawConfig {
     RawConfig {
-        port: None,
-        socks_port: None,
         mixed_port: Some(7890),
-        allow_lan: None,
-        bind_address: None,
         mode: Some("rule".into()),
-        log_level: None,
-        ipv6: None,
-        external_controller: None,
-        secret: None,
-        dns: None,
-        proxies: None,
-        proxy_groups: None,
         rules: Some(vec![
             "DOMAIN,example.com,DIRECT".into(),
             "MATCH,REJECT".into(),
         ]),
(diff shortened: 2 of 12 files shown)


madeye:
Post-merge architectural review (second pass, per team-lead request):

The `Rule::match_and_resolve` addition is **sound** — default delegates to the existing `match_metadata` + `adapter()` pair, so every pre-existing rule works unchanged and only `SubRuleRule` needs the override. Avoiding a breaking signature change on `Rule::match_metadata` (which has many impls) was the right call.

Three observations worth capturing here for the next engineer, no amendment required:

**1. Trait doc could flag the SUB-RULE-inside-logic interaction.** `AndRule`/`OrRule`/`NotRule` do NOT override `match_and_resolve`, so a SUB-RULE used as a child of a logic combinator falls through the default path: the combinator's own `match_metadata` calls the SubRule's `match_metadata` (which discards the inner adapter) and the combinator returns its own `adapter()`. This is consistent with upstream `rules/logic/logic.go` (logic combinators own the target, SUB-RULE doesn't), but the `match_and_resolve` doc in `mihomo-common/src/rule.rs:93-101` says *"override only when the resolved target must come from some other source — notably SUB-RULE"* and doesn't note that wrapping a SUB-RULE in AND/OR/NOT collapses back to the outer combinator's adapter. One sentence added to the doc would prevent a future engineer from "fixing" this by plumbing `match_and_resolve` through the logic combinators — which would silently diverge from upstream.

**2. Cycle detection is top-level-only (and that's fine).** `sub_rules_parser::parse_sub_rule_reference` only catches `SUB-RULE,<name>` lines at column 0 of a sub-rule block. A SUB-RULE nested inside an `AND,((SUB-RULE,X),...)` expression would bypass it — but `parse_one_rule_or_subrule` is not wired to the `sub_rules` resolver inside `mihomo_rules::parse_rule`'s logic path, so that nesting is not currently parseable. Noted for when/if logic rules gain SUB-RULE awareness (doesn't need to happen in M1).

**3. `SubRuleRule::match_metadata` vs `match_and_resolve` semantics.** The impl keeps `match_metadata` as "did anything match?" (line 67) for "API compatibility". That's the correct choice — callers who only want boolean membership (e.g. rule-ordering or stats counters keyed on top-level rule type) stay unsurprised. If a future `RuleMatchCounters` instrumentation (ADR-0004 §7 worked example) counts at `match_metadata` time, it will count SUB-RULE matches as type `SUB-RULE`, not the matched inner rule's type. That is the correct behaviour — ADR-0004 §1 Class I constraints rule out nested inner type labels — but worth noting in the M1.H-2 test plan.

No action required on this PR. Flagging for follow-up:
- One-line doc amendment to `Rule::match_and_resolve` (trivial, can ride a future PR that touches `mihomo-common/src/rule.rs`).
- Engineer on M1.H-2 (task #18) to confirm rule-match counter keys by top-level `rule_type()` not inner SUB-RULE match type.

Approving architectural shape.

PR #138, opened 2026-05-24
feat(snell): add Snell v4/v5 outbound adapter

missuo:
## Summary

- New `meow-proxy::snell` module porting the client side of [missuo/opensnell](https://github.com/missuo/opensnell): Argon2id KDF + AES-128-GCM AEAD framing with padding interleave and burst-aware payload-limit ramp-up, optional http/tls simple-obfs (reusing the SS codec), UDP-over-TCP datagram framing exposed as a `ProxyPacketConn`, and a bounded LIFO reuse pool for `CommandConnectV2` sessions (10 idle / 15 s max age / 2 uses per conn / 500 ms tail drain).
- YAML parser accepts mihomo-compatible `type: snell` blocks (`server`, `port`, `psk`, `version`, `udp`, `reuse`, `obfs-opts.{mode,host}`).
- New `snell` cargo feature on `meow-proxy` + `meow-config`, enabled by default.

Wire format implemented = snell v4 / v5 (client-side identical, matching opensnell's scope). Mihomo's legacy v1 / v2 / v3 wires are intentionally rejected at parse time with a Class-A error per ADR-0002.

Closes the Snell row in `docs/gap-analysis.md` (was "Medium priority Gap").

## Test plan

- [x] `cargo fmt --all --check`
- [x] `cargo clippy --all-targets -- -D warnings` — default features, `--no-default-features --features snell`, and the touched-crate set all clean.
- [x] `cargo test -p meow-proxy --features snell --lib snell::` — 9 unit tests pass (KDF determinism, AEAD frame round-trip through a duplex, bit-balanced padding, padding swap reversibility, nonce wrap, zero-chunk half-close detection, UDP request frame layout v4/v6).
- [x] Live end-to-end smoke test (`crates/meow-proxy/tests/snell_smoke.rs`, gated behind `SNELL_SMOKE=1` so CI never dials anyone): `SNELL_SMOKE=1 SNELL_SERVER=… SNELL_PSK=… cargo test -p meow-proxy --features snell --test snell_smoke -- --nocapture` — proxied `GET /ip` to `httpbin.org` and got back `HTTP/1.1 200 OK` with the `origin` field showing the proxy's IP, confirming the full handshake / AEAD / response-decode path is correct.

## Reference

Built by porting [`missuo/opensnell`](https://github.com/missuo/opensnell)'s `components/snell` client package to Rust + tokio. The reuse-pool semantics (max 2 sessions per TCP, 500 ms drain-before-reuse) and the v4 payload-limit ramp + padding-interleave specifics all follow that reference verbatim.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,7 +118,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -129,7 +129,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -174,6 +174,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6a3a1fd6f75306b68087b831f025c712524bcb19aad54e557b1129cfa0a2b207"
 dependencies = [
  "rustversion",
+]
+
+[[package]]
+name = "argon2"
+version = "0.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3c3610892ee6e0cbce8ae2700349fcf8f98adb0dbfbee85aec3c9179d29cc072"
+dependencies = [
+ "base64ct",
+ "blake2",
+ "cpufeatures 0.2.17",
+ "password-hash",
 ]
 
 [[package]]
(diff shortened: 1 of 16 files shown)


madeye:
Thanks for the contribution! This PR currently has merge conflicts with main and no CI results. Would you be able to rebase it onto the latest main? Quite a few perf-focused changes landed recently (including in `meow-proxy` and `meow-transport`), so a rebase will also let CI run against the current codebase. Happy to review once it's green.

missuo:
Thanks for taking a look! Rebased onto latest main (`bae6bdb`) — the branch is conflict-free now and CI should be able to run.

Notes from the rebase:

- **Feature merges**: main added `vmess` to the default feature sets while this branch adds `snell`; both are kept. The duplicate `aes-gcm` workspace dependency is deduped to the one main introduced for vmess (snell only needs `Aes128Gcm`, fully covered).
- **Feature wiring**: main moved `meow-app` to explicit per-protocol feature forwarding (the `full` bundle) instead of inheriting crate defaults, so I added `snell = ["meow-config/snell", "meow-proxy/snell"]` and included it in `full` — without this the binary parsed configs but rejected `type: snell`.
- **One rustdoc fix**: an intra-doc link in `snell/pool.rs` pointed at a private const, which trips `RUSTDOCFLAGS="-D warnings"` on current main.

Full regression bar is green locally (fmt, three-way clippy with `-D warnings`, rustdoc, `cargo test --lib`). Also re-verified end-to-end against a live snell v5 server after the rebase: HTTP CONNECT + SOCKS5 inbounds through the adapter, plus a 5 MB download, all routed and completed cleanly.

PR #224, opened 2026-06-16
feat(proxy): implement hysteria2 outbound in-tree on Quinn

gengjiawen:
本实现经 Composer 2.5 与 GPT-5.5 xhigh 审计。 Cursor 里没额度跑 claude 了 :)

## Summary

- 在 `meow-proxy` 内实现 Hysteria2 出站客户端（基于 Quinn），覆盖 HTTP/3 认证、TCP 流帧、UDP datagram、Salamander 混淆与端口跳跃。
- 认证后保持 HTTP/3 session 存活，避免 `poll_close` 过早关闭 QUIC 连接导致代理流失败。
- 新增 `hysteria2_integration` 集成测试：从官方 Docker 镜像提取 hysteria server 二进制并以宿主机子进程运行（Quinn 在嵌套容器 `docker --network container:…` 环境下无法完成握手）。
- 扩展 `meow-config` 解析与 README/CI 说明。

## Test plan

- [x] `cargo test --lib`
- [x] `cargo test --test hysteria2_integration --features hysteria2`
- [x] CI green on `codex/hysteria2`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -113,6 +113,11 @@ jobs:
       - name: Trojan integration tests
         run: cargo test --test trojan_integration -- --nocapture
 
+      - name: Hysteria2 Docker integration tests
+        env:
+          MEOW_REQUIRE_DOCKER: "1"
+        run: cargo test -p meow-proxy --features hysteria2 --test hysteria2_integration -- --nocapture
+
       - name: v2ray-plugin integration tests
         run: cargo test --test v2ray_plugin_integration -- --nocapture
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1293,6 +1293,34 @@ dependencies = [
 ]
 
 [[package]]
+name = "h3"
+version = "0.0.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "10872b55cfb02a821b69dc7cf8dc6a71d6af25eb9a79662bec4a9d016056b3be"
+dependencies = [
+ "bytes",
+ "fastrand",
+ "futures-util",
+ "http",
+ "pin-project-lite",
+ "tokio",
+]
+
+[[package]]
+name = "h3-quinn"
+version = "0.0.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b2e732c8d91a74731663ac8479ab505042fbf547b9a207213ab7fbcbfc4f8b4"
+dependencies = [
+ "bytes",
+ "futures",
+ "h3",
+ "quinn",
+ "tokio",
+ "tokio-util",
+]
+
+[[package]]
 name = "half"
 version = "2.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2179,12 +2207,16 @@ dependencies = [
  "argon2",
  "async-trait",
  "base64",
+ "blake2",
  "bytes",
  "chacha20poly1305",
  "crc32fast",
  "futures",
+ "h3",
+ "h3-quinn",
  "hex",
  "hmac",
+ "http",
  "libc",
  "md-5",
  "meow-common",
@@ -2202,6 +2234,7 @@ dependencies = [
  "smol_str",
  "socket2 0.5.10",
  "tempfile",
+ "thiserror 2.0.18",
  "tokio",
  "tokio-rustls",
  "tracing",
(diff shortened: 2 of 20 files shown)


gengjiawen:
@madeye Can you take a look, thx

madeye:
## Review notes: QUIC/TLS backend choice

Reviewed the QUIC and crypto-stack design. Summary: **the library and crypto choices are sound — keep `quinn` + rustls/`ring`.** Details and a couple of follow-ups below.

### Crypto stack — correctly does *not* use `boring-tls`
`hysteria2/tls.rs` wires Quinn to **rustls + the `ring` provider** (`rustls::crypto::ring::default_provider()` → `quinn::crypto::rustls::QuicClientConfig`), and `grep` for `boring`/`aws_lc` across the whole `hysteria2/` module returns nothing. That's the right call:

- `ring` is exactly the provider the rest of the repo standardizes on (`install_default()` in `main.rs`, `meow-config`, Trojan, AnyTLS, DNS). This PR adds **no new crypto backend**.
- The `boring-tls` feature exists for `meow-transport`'s **TCP** TLS path (ECH / uTLS ClientHello fingerprinting via `tokio-boring`). There is no `boring`-based Quinn crypto backend to reuse, so routing QUIC through BoringSSL isn't an option anyway.
- The `let _ = …install_default()` idempotent pattern matches the existing call sites, so no panic when `ring` + `aws-lc-rs` are both compiled in.

### quinn vs quiche
quinn is the better fit here, and switching to quiche would be a net regression for this codebase:

- **Async fit (decisive):** the whole project is tokio-async. quinn is tokio-native; quiche is sans-I/O and would force us to hand-roll the UDP loop, timers, and retransmit scheduling and bridge them into tokio.
- **Crypto consistency:** quiche vendors *its own* BoringSSL — separate from `boring-sys`. We can't drop rustls/`ring` (Trojan/AnyTLS/DNS depend on it), so quiche would leave us with **three** crypto stacks (rustls+ring, boring-sys BoringSSL, quiche BoringSSL) → worse binary size and build complexity.
- **Custom CC:** if Brutal congestion control is ever wanted, quinn's `Controller`/`ControllerFactory` traits make it implementable; quiche's CC is more internal.

quiche would only win for a Cloudflare-scale *server* or a runtime-agnostic library — neither applies to a tokio proxy client.

### Follow-ups (non-blocking)
1. **Binary-size delta (ADR-0007):** Quinn + `h3` + `h3-quinn` are a meaningful size add. Please include a before/after stripped-binary delta in the commit body per the ADR-0007 gate.
2. **`fingerprint` semantics:** it's consumed as a SHA-256 **certificate pin** (`PinVerifier`), not a uTLS ClientHello fingerprint. That matches mihomo's hysteria2 (QUIC has no uTLS), but worth a doc note so users don't expect fingerprint-spoofing behavior.
3. **Brutal CC:** the `up`/`down` fields are currently bandwidth hints only (default CC). Fine to land as-is — just confirm Brutal CC is explicitly out of scope for this PR.

madeye:
## Binary-size delta (ADR-0007 follow-up)

Measured the stripped-release size cost of this PR. Both binaries built identically with the release profile (`lto = true`, `opt-level = "z"`, `codegen-units = 1`, `panic = "abort"`, `strip = true`), default features:

| Build | `meow` size |
|-------|-------------|
| `main` (baseline, no hysteria2) | 7,572,544 B (7.22 MiB) |
| this PR (hysteria2 in default features) | 8,039,904 B (7.67 MiB) |
| **delta** | **+467,360 B (+456 KiB, +6.17%)** |

That ~456 KiB is the Hysteria2 module + the new `quinn` / `h3` / `h3-quinn` dependencies. No new crypto backend is linked (reuses the existing rustls + `ring` provider), so the cost is QUIC/HTTP-3 code, not a second TLS stack.

**Caveat on target:** this was built on **macOS arm64 (Mach-O)**, which ADR-0007 explicitly lists as out of scope for the size *caps*. The hard caps apply to `aarch64-unknown-linux-musl` (default 18 MiB) and `x86_64-unknown-linux-musl` (default 20 MiB), measured stripped on Linux. The +456 KiB delta is representative of the feature's footprint cost across targets, and there's ample headroom under the default caps — no breach risk. If you want the exact gating numbers for the CI size check, I can build the two musl targets via `cargo zigbuild` and post those too.

PR #232, opened 2026-06-18
Fix Snell UDP response frame parsing

gengjiawen:
## Summary

- Fix Snell UDP response parsing by reading one complete decrypted datagram frame before splitting address header and payload.
- Preserve datagram truncation semantics by copying at most the caller-provided output buffer length.
- Document the generic `/tmp` single-proxy config + app-level curl smoke flow, with Snell v3 as an example proxy block.

## Why

The previous parser read the address family/header and payload through separate `read` calls. Real Snell servers return each UDP response as one Snell frame whose decrypted body contains both the source address and payload, matching mihomo/opensnell behavior. Splitting it across reads can leave the payload unavailable to the caller and cause UDP DNS timeouts on real nodes.

## Tests

- `cargo fmt --check`
- `cargo test -p meow-proxy --features snell snell::udp --lib`
- `cargo test -p meow-proxy --features snell --test snell_integration`
- Manual real-node smoke:
  - Snell v3 TCP HTTPS curl returned `http_code=204` for default curl and `--http1.1`.
  - Snell v3 UDP DNS returned a response from `1.1.1.1:53`.
  - Snell v4 TCP HTTPS curl returned `http_code=204` for default curl and `--http1.1`.
  - Snell v4 UDP DNS returned a response from `1.1.1.1:53`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -118,6 +118,11 @@ jobs:
           MEOW_REQUIRE_DOCKER: "1"
         run: cargo test -p meow-proxy --features hysteria2 --test hysteria2_integration -- --nocapture
 
+      - name: Snell v3 Docker integration tests
+        env:
+          MEOW_REQUIRE_DOCKER: "1"
+        run: cargo test -p meow-proxy --features snell --test snell_server_docker_integration -- --nocapture
+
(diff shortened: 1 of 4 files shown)


gengjiawen:
@madeye PTAL

PR #252, opened 2026-06-26
fix: support Snell connect_over for dialer-proxy

gengjiawen:
## Summary
- implement `SnellAdapter::connect_over` so Snell can be used as the final hop in relay/dialer-proxy chains
- reuse Snell obfs + AEAD wrapping for fresh TCP and existing relay streams
- add a focused v3 in-memory test that verifies CONNECT header emission over an existing stream

## Verification
- `cargo test -p meow-proxy --lib`
- config validation against a private Clash-compatible subscription
- real private-subscription smoke for a Snell node using `dialer-proxy`:
  - default HTTPS curl returned `http_code=204` with HTTP/2
  - HTTP/1.1 HTTPS curl returned `http_code=204`

Runtime path confirmed generically as:
```text
relay: dial_tcp hop 0 through dialer proxy
relay: connect_over final hop 1 through Snell proxy
snell connecting to target over existing stream
```

### DIFF
--- a/crates/meow-proxy/src/snell/adapter.rs
+++ b/crates/meow-proxy/src/snell/adapter.rs
@@ -136,15 +136,22 @@ impl SnellAdapter {
             .await
             .map_err(MeowError::Io)?;
         let _ = tcp.set_nodelay(true);
+        let inner: Box<dyn TransportStream> = Box::new(tcp);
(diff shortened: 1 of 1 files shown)


gengjiawen:
@madeye PTAL

Issue #262, opened 2026-06-29
meow-proxy v0.15.1 won't compile with the `anytls` feature — crates.io `anytls-rs` 0.5.4 has no `Stream::close()`


madeye:
## Summary

`meow-proxy` fails to compile when the `anytls` feature is enabled, because it calls `anytls_rs::Stream::close()` but the dependency it resolves from crates.io — `anytls-rs = "0.5.4"` — does **not** provide that method.

The crates.io crate named `anytls-rs` (version 0.5.4) is **jxo-me's upstream crate** (`repository = https://github.com/jxo-me/anytls-rs`, author `Mickey <mickey@jxo.me>`), which has no `Stream::close()`. The `close()` method only exists in **madeye's fork** (`github.com/madeye/anytls-rs`), where it was added by the fd/stream-leak fix (madeye/anytls-rs#1). So a clean build against the registry cannot find it.

This affects anyone building meow-proxy with `anytls` from a clean registry (i.e. without the exact `Cargo.lock` shipped in this repo — and even that lock pins the same registry `0.5.4`).

## Reproduction

```sh
cargo build -p meow-proxy --features anytls
# or any downstream crate enabling meow-proxy/anytls or meow-config/anytls
```

## Error

```
error[E0599]: no method named `close` found for struct `Arc<anytls_rs::Stream>` in the current scope
   --> crates/meow-proxy/src/anytls_adapter.rs:259:21
    |
259 |         self.stream.close();
    |                     ^^^^^
...
   --> crates/meow-proxy/src/anytls_adapter.rs:270:21
    |
270 |         self.stream.close();
    |                     ^^^^^
error: could not compile `meow-proxy` (lib) due to 2 previous errors
```

## Root cause

`Cargo.toml` (workspace) declares:

```toml
# crates.io release of the madeye fork. 0.5.4 carries the `SocketProtector`
# registry + `connect_tcp` / `bind_udp` helpers (Android protect hook) and the
# fd/stream-leak fix for issue #201 item 4 (madeye/anytls-rs#1). ...
anytls-rs = { version = "0.5.4" }
```

[...]

PR #264, opened 2026-06-29
fix(anytls): patch anytls-rs to madeye fork commit with Stream::close (#262)

madeye:
Fixes the `anytls`-feature build break described in #262.

## Problem
`meow-proxy`'s anytls adapter calls `self.stream.close()` (`anytls_adapter.rs:259,270` — the fd/stream-leak fix for #201 item 4, madeye/anytls-rs#1). But crates.io `anytls-rs@0.5.4` is **jxo-me's upstream crate** (verified: `repository = github.com/jxo-me/anytls-rs`, published by `jxo-me`), which has **no `Stream::close()`**. That method only exists in the **madeye fork**, where it merged (`e6134889`) *after* the fork's own `v0.5.4` tag — so neither the registry crate nor that tag compiles.

Result: any build with `--features anytls` fails with `E0599: no method named close`, across v0.14.0 / v0.15.0 / v0.15.1. This is also why `clippy --all-features` is red in CI (e.g. on #263).

## Fix
Pin the registry crate to the exact fork merge commit via `[patch.crates-io]`:

```toml
[patch.crates-io]
anytls-rs = { git = "https://github.com/madeye/anytls-rs", rev = "e6134889d3abbfaa2cb7439969dbda2ef6930611" }
```

- The patched crate is itself `anytls-rs@0.5.4`, so it satisfies the existing `version = "0.5.4"` requirement — **no version bump**.
- The dep requirement stays a **registry version**, so `meow-proxy` remains publishable to crates.io. `anytls` is opt-in and off by default, so `cargo publish`'s verify build never compiles the anytls path.

## Caveat / follow-up
`[patch]` is workspace-local and stripped on publish, so it does **not** propagate to downstream consumers who pull `meow-proxy` from crates.io with `anytls` enabled — they'd still hit the broken upstream crate (and need their own patch). The permanent fix is to **publish the fork under a distinct name** (e.g. `anytls-meow`) and depend on that. That requires a crates.io publish from the fork repo, so I've left it as a follow-up rather than doing it unilaterally — happy to prep the fork rename if you want to go that route. Tracking stays in #262.

## Verification
All green locally, including the previously-failing path:
- ✅ `cargo build -p meow-proxy --features anytls` (the #262 repro)
- ✅ `cargo fmt --all -- --check`
- ✅ `cargo clippy --all-targets -- -D warnings`
- ✅ `cargo clippy --all-targets --no-default-features -- -D warnings`
- ✅ `cargo clippy --all-targets --all-features -- -D warnings` ← was failing before
- ✅ `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps`
- ✅ `cargo test --lib`

Refs #262

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -141,12 +141,12 @@ checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 [[package]]
 name = "anytls-rs"
 version = "0.5.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9fa494bfdf944c403ffb86d3387cb8a15422ca9a59af66fad50b941340aded56"
+source = "git+https://github.com/madeye/anytls-rs?rev=e6134889d3abbfaa2cb7439969dbda2ef6930611#e6134889d3abbfaa2cb7439969dbda2ef6930611"
 dependencies = [
  "anyhow",
  "bytes",
  "chrono",
+ "libc",
  "md5",
  "notify",
  "once_cell",
@@ -157,12 +157,13 @@ dependencies = [
  "serde",
  "sha2",
  "socket2 0.6.3",
- "thiserror",
+ "thiserror 2.0.18",
  "tokio",
  "tokio-rustls",
  "tokio-util",
  "tracing",
  "tracing-subscriber",
+ "trust-dns-resolver",
  "x509-parser",
 ]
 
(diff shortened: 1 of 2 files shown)


madeye:
Superseded by #265, which fixes #262 by **vendoring the anytls fork in-tree** (`crates/meow-anytls`) instead of patching the registry crate.

Why #265 is the better fix:
- The `[patch.crates-io]` here is workspace-local and stripped on publish, so it does **not** help downstream consumers who pull `meow-proxy` from crates.io with `anytls` enabled. Vendoring removes the external/registry dependency entirely, so it works for them too.
- meow-rs stays publishable, and `anytls` remains optional/off-by-default.

Closing in favor of #265.

PR #283, opened 2026-07-02
dns: preserve encrypted upstream labels

gengjiawen:
## Summary
- preserve the configured upstream label for encrypted DNS resolvers used by DNS results
- include the `tls://` / `https://` scheme for DoT and DoH fallback labels instead of reporting only the hostname
- omit default encrypted DNS ports from display labels while preserving explicit non-default ports and SNI overrides

## Motivation
When querying through a config that uses DoT upstreams such as `tls://dns.google:853`, `/dns/results` reported `from_server` as `dns.google`. That made encrypted DNS look indistinguishable from a plain upstream. The resolver now keeps a display label derived from the configured nameserver URL and attaches it to the `DnsClient`.

## Tests
- `cargo fmt`
- `cargo test -p meow-dns --lib`
- `cargo test -p meow-api --test api_test dns_results_returns_searchable_cache_entries`

### DIFF
--- a/crates/meow-dns/src/client.rs
+++ b/crates/meow-dns/src/client.rs
@@ -107,6 +107,7 @@ pub struct DnsClient {
     transport: Transport,
     timeout: Duration,
     proxy: Option<DnsProxy>,
+    label: Option<Arc<str>>,
 }
 
 enum Transport {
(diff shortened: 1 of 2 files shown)


madeye (approved):
Reviewed the encrypted DNS label preservation path. The label is captured from NameServerUrl before bootstrap resolution, default ports are canonicalized, IPv6 authorities are bracketed, explicit SNI/path information is retained, and proxy-routed labels still receive the existing #PROXY suffix. Existing focused tests and the full CI matrix pass. I found no blocking correctness issues.

Issue #288, opened 2026-07-03
perf(rule-ir): native lowering for GEOSITE / RULE-SET / GEOIP / logic fallback ops


madeye:
Pass 4 from the rule-engine review (follow-up to #285/#286, complements #287).

## Problem

GEOSITE, RULE-SET, GEOIP/SRC-GEOIP, IP-ASN, IP-SUFFIX, and AND/OR/NOT stay on the `RuleOp::Fallback` path: per connection that reaches them, they pay `dyn Rule` dispatch plus their own internal lookups. GEOSITE is the worst: `GeositeDB::lookup` re-splits attribute categories and — for `base@attr` payloads — allocates `format!("{base}@{attr}")` **per connection**, plus a category `HashMap` hash per probe.

## Proposed lowering

`meow-tunnel` already depends on `meow-rules`, so the IR can downcast (add `fn as_any(&self) -> Option<&dyn Any>` to `Rule`, default `None`, overridden by the lowerable rule types):

- **GeoSite** → resolve the category once at compile time to per-bucket handles (restructure `GeositeDB` to `HashMap<String, Arc<CategoryData>>` where `CategoryData` = trie + keywords + compiled regexes); op = `Box<[Arc<CategoryData>]>`, match = all buckets contain host. Kills the per-connection allocation, category hashing, and attr split.
- **RuleSet** → op holding the concrete set variant (Domain/IpCidr/Classical) resolved once.
- **GeoIp/IpAsn/IpSuffix** → ops over their existing shared indexes (`country_index`, ASN index).
- **And/Or/Not** → recursive `RuleOp` expression tree (`All/Any/Not(Box<[RuleOp]>)`); children are parser-produced simple predicates, so they lower fully; any non-lowerable child keeps the whole rule on fallback.

## Notes

- Demand flags from #286 must be preserved per lowered op (no-resolve semantics).
- `never_matches` pruning (#285) can then also cover unknown geosite categories.
- Expected effect: near-zero `dyn` dispatch on the match hot path; GEOSITE attr rules stop allocating per connection.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

madeye:
[...]

PR #291, opened 2026-07-03
perf(rule-ir): native lowering for GEOSITE / RULE-SET / GEOIP / IP-ASN / IP-SUFFIX / logic rules

madeye:
## Summary

Implements #288 (pass 4 of the rule-engine review). Stacked on #286 → #285; merge in order and re-target.

The rule types that always fell back to `dyn Rule` dispatch now lower to native opcodes when their match state is cheap to share (`Arc` handles or `Copy` matchers).

### What changed

- **`Rule::as_any()` hook** (default `None`) — the IR downcasts to concrete rule types; `match_metadata` remains the semantic source of truth, lowering is optimization only.
- **GEOSITE** → pre-resolved bucket handles via new `GeositeDB::resolve_keys` / `lookup_resolved`. The per-connection costs it used to pay — attribute splitting, case-fold key scans, and a `format!("{base}@{attr}")` **heap allocation per connection** for attribute categories — all move to compile time. Bonus: `resolve_keys` returning `None` upgrades `never_matches`, so GEOSITE rules naming unknown categories are pruned from the plan entirely.
- **RULE-SET** → frozen `Arc<dyn RuleSet>` handle. Safe to freeze: rule-provider refresh flows through `Tunnel::update_rules`, which rebuilds the IR.
- **GEOIP / SRC-GEOIP / IP-ASN** → one shared `IpRanges` op over their existing Arc'd Patricia tries; **IP-SUFFIX** → its `Copy` matcher (boxed — the inline `u128` V6 masks would have doubled `RuleOp` from 32 B to 64 B; a `size_of` guard test now pins `RuleOp ≤ 32 B`).
- **AND / OR / NOT** → recursive `AllOf`/`AnyOf`/`NotOp` expression trees over lowered children; a non-lowerable child keeps the whole logic rule on fallback.

Duplicate elimination stays restricted to payload-pure ops; state-carrying ops are exempt like `Fallback`.

### Verification

- Full bar green: fmt, three-way clippy `-D warnings`, doc `-D warnings`, 12 lib suites, `rules_test` ×102.
- The seeded randomized differential test now exercises logic rules through the lowered path.
- New tests: GEOSITE attribute lowering, unknown-category pruning, GEOIP/IP-SUFFIX lowered ops (hit+miss), `AND(suffix, NOT(port))` expression tree, opaque-child fallback preservation, `RuleOp` size guard.
- `rule_ir_footprint` (release): hot loop remains **zero-allocation**.

ADR-0011: no tracked types touched.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2348,6 +2348,7 @@ dependencies = [
  "criterion",
  "dashmap",
  "ipnet",
+ "iprange",
  "itoa",
  "meow-common",
  "meow-config",
--- a/crates/meow-common/src/rule.rs
+++ b/crates/meow-common/src/rule.rs
@@ -138,6 +138,16 @@ pub trait Rule: Send + Sync {
         false
     }
 
+    /// Concrete-type escape hatch for the rule IR compiler: rule types that
+    /// can be lowered to native opcodes (their match state is cheap to share
+    /// — `Arc` handles or `Copy` matchers) override this to return `self`.
+    /// The default `None` keeps the rule on the virtual-dispatch fallback
+    /// path. Lowering is an optimization only; `match_metadata` remains the
+    /// source of truth for semantics.
+    fn as_any(&self) -> Option<&dyn std::any::Any> {
+        None
+    }
+
     /// Match against metadata and, on match, return the routing target.
     ///
     /// Default: `Some(self.adapter())` when `match_metadata`
--- a/crates/meow-rules/src/geoip.rs
+++ b/crates/meow-rules/src/geoip.rs
@@ -31,6 +31,12 @@ impl GeoIpRule {
     }
 }
 
+impl GeoIpRule {
+    pub fn ranges(&self) -> &CountryRanges {
+        &self.ranges
+    }
+}
+
 impl Rule for GeoIpRule {
     fn rule_type(&self) -> RuleType {
         RuleType::GeoIp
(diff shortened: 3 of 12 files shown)


madeye:
Pushed a follow-up commit (`a2a897f`) after benchmarking caught a codegen regression the test suite could not see.

**What happened:** `rules_bench` showed `synthetic_10k_wildcard_miss` at 52.6 µs vs 22.7 µs on main (+2.3x). Bisect across the stack pinned it to this PR's lowering commit; a symbolized `sample` profile showed the workspace's `opt-level = "z"` + fat LTO had stopped inlining the grown `matches_op` into the scan loop, and the machine outliner had shredded the glob matcher's compare loops into `OUTLINED_FUNCTION_*` fragments.

**Fix:** keep `matches_op` slim (`#[inline(always)]`, one-comparison arms only) with all fat arms behind a single `#[inline(never)] matches_native_op`; force-inline the wildcard matcher leaves; monomorphize the scan's demand-stop branch via const generic.

**Final numbers (median, IR vs legacy engine):**

| case | legacy | compiled IR |
|---|---|---|
| synthetic_10k_domain_suffix_hit | 58.9 µs | **4.15 µs** |
| synthetic_10k_match_fallthrough | 66.3 µs | **5.49 µs** |
| synthetic_10k_wildcard_miss | 77.2 µs | **22.76 µs** (= main's 22.7, regression gone) |
| fixture_geosite_github | 206.9 ns | **160.8 ns** |
| fixture_geoip_cn | 130.8 ns | **115.4 ns** |
| fixture_match_fallthrough | 7.02 µs | **6.49 µs** |

Stripped release binary: 8355936 → 8339536 B (**−16.4 KB** vs base branch, ADR-0007 clean). Full bar re-run green.

PR #293, opened 2026-07-04
perf(rule-ir): fuse GEOSITE / RULE-SET domain tries into the global index

### LINKED ISSUE #287: perf(rule-ir): fuse GEOSITE / RULE-SET(domain) tries into the global domain index
Pass 3 from the rule-engine review (follow-up to #285/#286). The headline win for realistic configs.

## Problem

Real Clash-style configs are dominated by `GEOSITE,cn`-style rules. These are already tries internally (`GeositeDB` = `HashMap<category, DomainTrie<()>>`, `DomainRuleSet` = `DomainTrie<()>`), but the IR treats them as opaque fallback slots: a `GEOSITE` rule at index 500 forces every connection through 500 slots of linear scan before the rule's own private probe, and never participates in trie early-exit.

## Proposed pass

At `CompiledRuleSet::build`, for each live GEOSITE / RULE-SET(domain) slot, merge the referenced category/rule-set trie into the global `DomainIndex` with the slot's rule index as the value (min-index on collisions — the infrastructure from #285 already provides this).

- Requires domain enumeration over `DomainTrie` (add a sealed-node DFS iterator reconstructing `pattern → (kind, labels)`), or storing pattern lists alongside the tries at parse time.
- Categories with keyword/regex entries can't fully fuse: fuse the trie part, keep a residual op that evaluates only the keyword/regex tail (slot not trie-owned; hit-path still benefits).
- Attribute categories: fuse only single-attribute buckets (`base@attr` is its own bucket); multi-attr lookups are intersections and stay on the fallback path.
- RULE-SET fusion must be gated on provider immutability — if rule providers hot-swap contents at runtime, either rebuild the IR on refresh (current `update_rules` path) or exclude refreshable sets.

## Expected effect

One trie probe answers domain routing for essentially the whole config; the scan only covers non-domain residual slots below the min-index ceiling.

Measure with `rules_bench` (fixture is GEOSITE-heavy) and `rule_ir_footprint` — the merged trie duplicates category tries, so the memory delta must be reported (ADR-0011 discipline).

madeye:
## Summary

Implements #287 with the **relaxed size gate** decided there: GEOSITE and domain-behavior RULE-SET slots merge their pattern tries into the compiled plan's `DomainIndex`, so one min-index trie probe answers them together with every plain DOMAIN/DOMAIN-SUFFIX rule.

### Semantics (first-match-wins preserved exactly)

- Only **single-bucket** GEOSITE rules fuse — multi-attribute categories are intersections, which a union-style index cannot express (a fused hit would falsely claim a match).
- Buckets with keyword/regex **residuals** fuse their trie part (hits gain early-exit) but stay on the scan path; residual-free buckets and domain rule-sets become fully **trie-owned** — skipped in every scan, enrichment demands cleared.
- New `DomainTrie::for_each_pattern` emits canonical insertable patterns; round-trip match-semantics tested (sealed + building). Ascending insertion + first-write-wins keeps min-index per pattern.
- **Gate:** 500k pattern slots per rule (real geosite giants like `cn` are ~200–300k); skips are logged, never silent.
- Plan selection now weighs fused pattern counts (review pass 8): two heavyweight GEOSITE rules select the indexed plan even below the 64-rule threshold.

### Codegen hardening

Third measured incident of workspace-wide `opt-level = "z"` shredding the scan loop (this time `Option::is_some_and` compiled as an outlined call per slot). Fix: per-package `opt-level = 3` for `meow-tunnel` + `meow-trie` — fat LTO honors per-function optsize attrs, so the rest of the binary stays size-optimized. This also ends the per-PR codegen lottery on the hot path.

### Numbers

| case | pre-fusion (#291 tip) | this PR |
|---|---|---|
| fixture_match_fallthrough | 6.49 µs | **4.50 µs** |
| fixture_geoip_cn | 115 ns | **46 ns** |
| fixture_geosite_github | 161 ns | **156 ns** |
| synthetic_10k_wildcard_miss | 22.8 µs | **15.4 µs** |
| synthetic_10k_suffix_hit | 4.15 µs | **4.02 µs** |
| fixture domain-suffix probe | 24 ns | 151 ns (real mega-trie fanout cost; one probe/connection) |

Footprint: fixture RSS after IR build 47.1 → 72.1 MiB (+25 MiB fused patterns — the accepted trade). Hot loop stays **zero-allocation**. Binary size (ADR-0007): default 8 388 992 B (+49 KB, vs 18 MiB cap), minimal 4 637 344 B (vs 8 MiB hard cap).

Verification: full bar green (fmt, three-way clippy `-D warnings`, doc, 12 lib suites, `rules_test` ×102); new tests for owned fusion, residual buckets, multi-attr exclusion, rule-set fusion, min-index interplay; randomized differential now generates GEOSITE rules.

## Full rules_bench results (this branch, medians)

All three engines measured from the same run; the per-package `opt-level = 3` override lifts the legacy columns too, so these are apples-to-apples at shipped codegen settings.

| case | legacy linear | legacy indexed | compiled IR | speedup |
|---|---|---|---|---|
| fixture_domain_suffix_maxlv | 21.8 ns | 41.8 ns | 142.5 ns | 0.2× |
| fixture_geosite_github | 207.6 ns | 223.2 ns | **154.4 ns** | 1.3× |
| fixture_geoip_cn | 125.3 ns | 140.9 ns | **45.8 ns** | 2.7× |
| fixture_match_fallthrough | 6.78 µs | 6.74 µs | **4.40 µs** | 1.5× |
| synthetic_10k_domain_suffix_hit | 24.20 µs | 23.95 µs | **4.00 µs** | 6.0× |
| synthetic_10k_match_fallthrough | 34.84 µs | 35.85 µs | **5.32 µs** | 6.6× |
| synthetic_10k_wildcard_miss | 44.57 µs | 45.02 µs | **15.24 µs** | 2.9× |

The first row is the documented fusion trade-off: the fixture's `DOMAIN-SUFFIX,maxlv.net` sits at rule index 0, so the legacy linear scan matches in one comparison, while the IR always pays a single probe of the fused mega-trie (~142 ns, once per connection). In exchange the fall-through row — where rule engines actually spend their time — is 1.5× faster, and the wildcard row drops further to 5.19 µs with the stacked #294.

Footprint from the same run: fixture RSS after IR build 47.1 → 72.1 MiB (+25 MiB fused patterns, inside the relaxed 500k-slot gate); hot matching loop remains zero-allocation.

Closes #287.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,18 @@ codegen-units = 1
 panic = "abort"
 opt-level = "z"
 
+# Hot-path exception to the size-first default: the rule-match engine and
+# the domain trie run per connection, and at `opt-level = "z"` their scan
+# and probe loops get shredded by the -Oz inliner/machine-outliner (2-3x
+# swings measured on rules_bench across otherwise-unrelated code growth).
+# Fat LTO honors per-function optsize attrs, so exempting just these two
+# crates buys back the hot path while the rest of the binary stays "z".
+[profile.release.package.meow-tunnel]
+opt-level = 3
+
+[profile.release.package.meow-trie]
+opt-level = 3
+
 # Profile for dhat heap profiling — keeps debug symbols, no strip, no LTO.
 [profile.dhat]
 inherits = "release"
--- a/crates/meow-rules/src/geosite.rs
+++ b/crates/meow-rules/src/geosite.rs
@@ -154,6 +154,20 @@ impl GeositeDB {
             .collect()
     }
 
+    /// The domain trie behind one pre-resolved bucket key from
+    /// [`Self::resolve_keys`], if the bucket has one.
+    pub fn bucket_domain_trie(&self, key: &str) -> Option<&DomainTrie<()>> {
+        self.categories.get(key)
+    }
+
+    /// True iff the bucket has non-trie match sources (keywords / regex)
+    /// alongside its domain list — such a bucket cannot be fully handed
+    /// over to an external domain index.
+    pub fn bucket_has_residuals(&self, key: &str) -> bool {
+        self.keywords.get(key).is_some_and(|k| !k.is_empty())
+            || self.regex_compiled.get(key).is_some_and(|r| !r.is_empty())
+    }
+
     /// Match `domain` against one pre-resolved bucket key from
     /// [`Self::resolve_keys`]. Skips the per-lookup attribute splitting and
     /// case-fold scan that [`Self::lookup`] pays.
--- a/crates/meow-rules/src/rule_set.rs
+++ b/crates/meow-rules/src/rule_set.rs
@@ -76,6 +76,13 @@ pub trait RuleSet: Send + Sync {
     fn is_empty(&self) -> bool {
         self.len() == 0
     }
+
+    /// Concrete-type escape hatch for the rule IR compiler (mirrors
+    /// `Rule::as_any`); set variants that can be fused into an external
+    /// index override this.
+    fn as_any(&self) -> Option<&dyn std::any::Any> {
+        None
+    }
 }
 
 /// Build a rule-set of the given behavior from already-parsed entries.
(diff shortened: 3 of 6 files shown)


madeye:
Dropping per maintainer decision after reviewing the full benchmark table: the realistic win (1.5× on fixture fall-through) doesn't justify +25 MiB RSS and the 6× slower per-connection domain probe from mega-trie fanout. The branch stays as reference; #287 remains open with the measurements. The stacked #294 (wildcard indexing) will be reworked onto main — it needs none of the fusion machinery.

PR #309, opened 2026-07-10
fix(lint): satisfy clippy 1.97 question_mark in reverse_lookup

madeye:
## Summary

The `lint` job on main is red: CI uses `dtolnay/rust-toolchain@stable`, which picked up Rust 1.97, and its clippy newly flags the `if let / else return None` block in `DnsCache::reverse_lookup` (`crates/meow-dns/src/cache.rs:198`) under `clippy::question_mark`. Both post-merge Test runs on main (`0609fed`, `f54a7d6`) failed on this single error; the code itself dates from April and is unrelated to those commits.

Rewritten with the `?` operator. Behavior is unchanged:
- clean miss → `None` without popping (preserves the "skip useless pop on clean miss" optimization from 8007a33)
- fresh hit → domain returned
- expired hit → entry popped, `None`

## Verification

Full regression bar run locally:
- `cargo clippy --all-targets -- -D warnings` (default / `--no-default-features` / `--all-features`)
- `cargo test --lib` — 64 passed

No struct layout changes (control flow only).

### DIFF
--- a/crates/meow-dns/src/cache.rs
+++ b/crates/meow-dns/src/cache.rs
@@ -195,12 +195,9 @@ impl DnsCache {
         let shard = &self.reverse[shard_ip(ip)];
         let mut reverse = shard.lock();
         let now = Instant::now();
-        if let Some(entry) = reverse.get(&ip) {
-            if entry.expire_at > now {
-                return Some(SmolStr::from(entry.domain.as_ref()));
-            }
-        } else {
-            return None;
+        let entry = reverse.get(&ip)?;
+        if entry.expire_at > now {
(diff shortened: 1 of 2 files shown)


madeye:
Pushed a second commit: the first CI run surfaced another clippy 1.97 `question_mark` site in `crates/meow-proxy/src/health.rs` (`ParsedUrl::parse`) that was hidden behind the meow-dns error. Installed Rust 1.97.0 locally and swept the whole workspace with all three clippy variants (`-D warnings`) — these two were the only sites. Full regression bar re-run: fmt, three-way clippy under 1.97, rustdoc `-D warnings`, `cargo test --lib` all green.

PR #310, opened 2026-07-10
feat: add Windows platform support

xmh0511:
- Add Windows as a supported platform via Win32 API process lookup
- Gate SIGTERM signal handler for Unix only; Windows uses ctrl_c() only
- Add Windows service management stubs (not yet implemented)
- Implement Windows process lookup using Win32 APIs (GetExtendedTcpTable/GetExtendedUdpTable + QueryFullProcessImageNameW) for PROCESS-NAME/PROCESS-PATH rules support on Windows
- Read Windows hosts file from C:\Windows\System32\drivers\etc\hosts
- Add windows-sys crate as Windows-only dependency in meow-common
- Add windows-latest CI job (clippy + build + smoke tests)
Details
Process Lookup (meow-common/src/process_lookup.rs)
Windows process lookup queries the kernel TCP/UDP owner-PID tables via GetExtendedTcpTable / GetExtendedUdpTable, then resolves the process image path with QueryFullProcessImageNameW (documented to work with PROCESS_QUERY_LIMITED_INFORMATION handles). Supports:
- IPv4 and IPv6 (AF_INET + AF_INET6)
- TCP and UDP
- UDP wildcard binds (0.0.0.0 / ::)
- Buffer-size race retry on ERROR_INSUFFICIENT_BUFFER
- Sound pointer handling via read_unaligned
TProxy on Windows
TProxy compiles on Windows via existing #[cfg(not(any(target_os = "macos", target_os = "linux")))] fallback PlatformGuard in firewall.rs and orig_dest.rs — returns runtime errors with clear messages. Default features unchanged.
Hosts File (meow-config/src/dns_parser.rs)
merge_system_hosts now reads C:\Windows\System32\drivers\etc\hosts on Windows (via SystemRoot env var) instead of no-op warning.

可以正常代理了，windows的powershell的curl是alias的，不走`$env`的设置

图片

图片

 `curl -x socks5h://127.0.0.1:7890 https://www.google.com -v` 是可以通信的

图片

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,6 +29,25 @@ jobs:
             # aarch64: hard gate — release fails if minimal binary exceeds 8 MiB
             minimal_budget_bytes: 8388608
             minimal_soft_gate: false
+            # Static musl binary → one build serves several OpenWrt arch
+            # labels; the .ipk differs only in the Architecture field (#284).
+            openwrt_archs: >-
+              aarch64_generic aarch64_cortex-a53 aarch64_cortex-a72
+              aarch64_cortex-a76
+          - target: armv7-unknown-linux-musleabihf
+            # armv7: soft gate only for now (no ADR-0007 budget yet)
+            minimal_budget_bytes: 0
+            minimal_soft_gate: true
+            # boring-sys is broken on 32-bit musl (bindgen emits time64
+            # `time_t: i64` per musl 1.2, while boring passes i32), so build
+            # without the boring-tls default. Trade-off: no ECH/uTLS
+            # fingerprinting on armv7 — everything else in `full` is present.
+            cargo_flags: --no-default-features --features full
+            # Rust's armv7 musleabihf baseline is vfpv3-d16 without NEON, so
+            # the same binary is safe on every core listed here (#284).
+            openwrt_archs: >-
+              arm_cortex-a7_neon-vfpv4 arm_cortex-a8_vfpv3
+              arm_cortex-a9_vfpv3-d16 arm_cortex-a15_neon-vfpv4
           # mipsel-unknown-linux-musl: temporarily disabled — `rust-std` is no
           # longer published for this target on stable. Re-enable when stable
           # ships the component again, or when ADR-0007 §1 promotes it to a
@@ -52,8 +71,8 @@ jobs:
 
       - run: cargo install --locked cargo-zigbuild
 
-      - name: Build release binary (static musl, default features)
-        run: cargo zigbuild --release --locked --target ${{ matrix.target }} --bin meow
+      - name: Build release binary (static musl)
+        run: cargo zigbuild --release --locked ${{ matrix.cargo_flags }} --target ${{ matrix.target }} --bin meow
 
       - name: Build minimal release binary (stripped, for size check)
         if: matrix.minimal_budget_bytes != 0
@@ -90,7 +109,10 @@ jobs:
         id: pkg
         run: |
           set -euo pipefail
-          ref="${GITHUB_REF_NAME:-${{ github.event.inputs.tag }}}"
+          # On workflow_dispatch GITHUB_REF_NAME is the *branch* name (which
+          # may contain slashes and break the tarball path) — use the tag
+          # input instead.
+          ref="${{ github.event_name == 'workflow_dispatch' && github.event.inputs.tag || github.ref_name }}"
           name="meow-${ref}-${{ matrix.target }}"
           staging="dist/${name}"
           mkdir -p "${staging}"
@@ -101,6 +123,21 @@ jobs:
           echo "archive=${name}.tar.gz" >> "$GITHUB_OUTPUT"
           echo "checksum=${name}.tar.gz.sha256" >> "$GITHUB_OUTPUT"
 
+      - name: Package OpenWrt ipks
+        if: matrix.openwrt_archs != ''
+        run: |
+          set -euo pipefail
+          ref="${{ github.event_name == 'workflow_dispatch' && github.event.inputs.tag || github.ref_name }}"
+          version="${ref#v}"
+          version="${version//\//-}"
+          for arch in ${{ matrix.openwrt_archs }}; do
+            bash openwrt/build-ipk.sh meow \
+              --binary "target/${{ matrix.target }}/release/meow" \
+              --version "${version}-1" --arch "${arch}" --outdir dist-ipk
+          done
+          cd dist-ipk
+          for f in *.ipk; do sha256sum "$f" > "$f.sha256"; done
+
       - name: Upload artifacts
         uses: actions/upload-artifact@v4
         with:
@@ -108,6 +145,8 @@ jobs:
           path: |
             ${{ steps.pkg.outputs.archive }}
             ${{ steps.pkg.outputs.checksum }}
+            dist-ipk/*.ipk
+            dist-ipk/*.ipk.sha256
           if-no-files-found: error
           retention-days: 14
 
@@ -117,10 +156,21 @@ jobs:
     needs: build
     if: startsWith(github.ref, 'refs/tags/v')
     steps:
+      - uses: actions/checkout@v4
+
       - uses: actions/download-artifact@v4
         with:
           path: dist
           merge-multiple: true
+
+      - name: Package luci-app-meow ipk (arch-independent)
+        run: |
+          set -euo pipefail
+          version="${GITHUB_REF_NAME#v}"
+          bash openwrt/build-ipk.sh luci --version "${version}-1" --outdir dist
+          cd dist
+          f="luci-app-meow_${version}-1_all.ipk"
+          sha256sum "$f" > "$f.sha256"
 
       - name: Publish release
         uses: softprops/action-gh-release@v2
@@ -128,5 +178,7 @@ jobs:
           files: |
             dist/*.tar.gz
             dist/*.tar.gz.sha256
+            dist/**/*.ipk
+            dist/**/*.ipk.sha256
           generate_release_notes: true
           fail_on_unmatched_files: true
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -269,17 +269,41 @@ jobs:
       - name: Build
         run: cargo build --tests
 
-      - name: Core smoke tests
-        run: |
-          cargo test --lib
-          cargo test --test common_test -- --nocapture
-          cargo test --test config_test -- --nocapture
-          cargo test --test rules_test -- --nocapture
-          cargo test --test api_test -- --nocapture
-          cargo test --test trojan_integration -- --nocapture
-          cargo test -p meow-dns --lib -- --nocapture
-          cargo test -p meow-transport --test tls_test -- --nocapture
-          cargo test -p meow-transport --test ws_test -- --nocapture
+      - name: Unit tests (windows)
+        shell: bash
+        run: cargo test --lib
+
+      - name: Common types tests (windows)
+        shell: bash
+        run: cargo test --test common_test -- --nocapture
+
+      - name: Config parsing tests (windows)
+        shell: bash
+        run: cargo test --test config_test -- --nocapture
+
+      - name: Rules tests (windows)
+        shell: bash
+        run: cargo test --test rules_test -- --nocapture
+
+      - name: API tests (windows)
+        shell: bash
+        run: cargo test --test api_test -- --nocapture
+
+      - name: Trojan integration tests (windows)
+        shell: bash
+        run: cargo test --test trojan_integration -- --nocapture
+
+      - name: DNS unit tests (windows)
+        shell: bash
+        run: cargo test -p meow-dns --lib -- --nocapture
+
+      - name: Transport TLS tests (windows)
+        shell: bash
+        run: cargo test -p meow-transport --test tls_test -- --nocapture
+
+      - name: Transport WS tests (windows)
+        shell: bash
+        run: cargo test -p meow-transport --test ws_test -- --nocapture
 
   macos:
     runs-on: macos-latest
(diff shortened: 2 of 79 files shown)


madeye:
Thanks for the Windows work! One request before this can merge: please add **Windows CI coverage** to this PR, otherwise none of the new `#[cfg(windows)]` code is exercised by our checks and it will rot immediately.

Concretely, in `.github/workflows/test.yml`:

1. Add a `windows-latest` job that runs at least:
   ```
   cargo clippy --all-targets -- -D warnings
   cargo test --lib
   ```
   (the same three-way clippy matrix as Linux/macOS if feasible)
2. If a full Windows runner is too heavy, the minimum bar is a cross-check from the Linux job:
   ```
   rustup target add x86_64-pc-windows-msvc
   cargo check --workspace --target x86_64-pc-windows-msvc
   ```
   — but a real `cargo test --lib` on `windows-latest` is strongly preferred, since this PR adds runtime behavior (Win32 process lookup, hosts parsing, signal handling), not just compilation.

Note: as written, `process_lookup::tests::finds_self_via_tcp_listener` will fail on Windows — it asserts `info.uid.is_some()` but the Win32 impl returns `uid: None`. Adding the Windows test job will surface that immediately; please gate that assertion to Unix (or populate a Windows equivalent) as part of the same change.

xmh0511:
@madeye I have added the Windows CI coverage and gated that assertion to Unix

madeye (changes requested):
Thanks for taking on Windows support — the overall direction here is right: the `#[cfg]` gating strategy is correct, the Windows CI job mirrors this repo's three-way clippy gate, and enabling the existing process-lookup test on Windows instead of skipping it is exactly what we want. A few things need to change before this can merge.

## Branch structure

This PR overlaps heavily with #313 (which contains most of these commits plus the mihomo-compat work, but is missing the `ech_dns.rs` gating, the `systemd_config_test.rs` gate, the `firewall.rs` import gating, and the CI job). Please keep **all** Windows work in this PR and rebase #313 down to compat-only changes — as-is, merging either one conflicts with the other. Also consider working from a feature branch rather than your fork's `main`; it makes rebasing much easier.

## Blocking: default-feature changes regress other platforms

Replacing `default = ["full", "boring-tls"]` in `meow-app/Cargo.toml` with a hand-picked list (and dropping `listener-tproxy` from `meow-listener`'s defaults) changes Linux builds too:

- Linux users building with default features silently lose the tproxy listener.
- The hand-picked list is already stale: it omits `vless-encryption`, which entered `full` when #317 merged — the inherent hazard of a list that has to be manually kept in sync with `full`.

Cargo features can't be platform-conditional, so please keep the defaults as they are and instead make the tproxy listener compile to a runtime-error stub on Windows (`firewall.rs` is already platform-gated internally — this PR even patches it to compile everywhere), or document `--no-default-features` for Windows builds.

## `process_lookup.rs` Windows implementation

1. **Unaligned pointer casts (soundness).** The table walk casts a `Vec<u8>` buffer to `*const u32` / `*const MIB_TCPROW_OWNER_PID` and dereferences. `Vec<u8>` only guarantees 1-byte alignment, so this is technically UB (it works because allocators over-align, but `unsafe` blocks should be sound on paper). Allocate a `Vec<u32>` of `(size + 3) / 4` elements, or use `std::ptr::read_unaligned`.
2. **IPv4 only.** Both lookups pass `AF_INET`, so PROCESS-NAME/PROCESS-PATH rules silently never match IPv6 connections. mihomo queries both families. At minimum add a comment and a note in the PR description; ideally query the `AF_INET6` tables too.
3. **UDP wildcard binds won't match.** UDP client sockets typically bind `0.0.0.0:port`; the row's `dwLocalAddr` is 0 while `target.ip()` is concrete, so the equality check fails and UDP process rules mostly won't work. Match on port plus (address equal **or** row address unspecified), as mihomo does.
4. **Buffer-size race.** The table can grow between the size query and the fetch; the second call then fails and the lookup returns `None`. Standard fix: retry 2–3 times on `ERROR_INSUFFICIENT_BUFFER`.
5. **Process access rights.** `OpenProcess(PROCESS_QUERY_INFORMATION | PROCESS_VM_READ)` fails for elevated/system processes when meow runs unprivileged. `QueryFullProcessImageNameW` with `PROCESS_QUERY_LIMITED_INFORMATION` needs fewer rights and is the modern replacement for `GetModuleFileNameExW`.
6. Nit: `(row.dwLocalPort as u16).swap_bytes()` reads better (and is endian-correct by construction) as `u16::from_be(...)`.

## Unrelated changes — please split out

- `PROBE_TIMEOUT` 5s → 10s is a global behavior change (slower dead-node detection on every platform) with no justification given. If your nodes need a longer probe timeout, that should be a config option, not a constant bump buried in a Windows PR.
- The health-check error-detail logging (`probe_many_bounded_detailed`) is a nice improvement — worth its own small PR that can merge immediately.
- `.mimocode/` in `.gitignore` is contributor-local tooling; please drop it or put it in your global gitignore.

## Minor

- `#[allow(clippy::unnecessary_wraps)]` in `firewall.rs` needs a `reason = "…"` per this repo's lint policy (no silent allows).

I verified `probe_many_bounded_detailed` and the `ech_dns.rs` gating are consistent with current `main` (the only `parse_nameserver_token` caller is already inside `#[cfg(unix)]`), so the code should compile — but CI hasn't run on this PR yet, so fmt/clippy/tests are all unverified. Once the branch is rebased and de-overlapped we'll approve the workflow run.

xmh0511:
@madeye  Please review the change for your previous requirements.

madeye:
# Review — feat: add Windows platform support

## Overview

Adds Windows as a supported platform: a Win32-API process lookup (`GetExtendedTcpTable`/`GetExtendedUdpTable` + `GetModuleFileNameExW`) so PROCESS-NAME/PROCESS-PATH rules work on Windows, Windows hosts-file support, Unix-gating of SIGTERM handling / service management / resolv.conf parsing, and a new `windows-latest` CI job. The overall structure is good — cfg boundaries are placed sensibly, the unsafe FFI is bounds-checked, and enabling the existing `process_lookup` test on Windows gives the new code real coverage in CI.

## Correctness issues

- **`GetModuleFileNameExW` with `PROCESS_QUERY_LIMITED_INFORMATION` is contrary to its documented contract** (`process_lookup.rs`, `get_process_info`). MSDN requires `PROCESS_QUERY_INFORMATION | PROCESS_VM_READ` for `GetModuleFileNameExW`; the API that is *documented* to work with a LIMITED handle is `QueryFullProcessImageNameW`. The doc comment even names `QueryFullProcessImageNameW` but the code calls the other function. If the OS enforces the documented rights, every lookup returns `None` and PROCESS rules silently never match. Please switch to `QueryFullProcessImageNameW` (feature `Win32_System_Threading`, already enabled) — it's strictly better here: works with LIMITED handles, returns a Win32 path, and reports truncation. The enabled-on-Windows unit test will confirm either way, but there's no green Windows CI run attached to this PR yet, so this needs verification before merge.

- **The Windows CI test step can mask failures.** `windows-latest` steps default to `pwsh`, which only propagates the *last* command's exit code — if `cargo test --lib` fails but the final `ws_test` passes, the step goes green. The macOS/Linux jobs don't have this problem (`bash -e`). Split the nine commands into separate steps, or check `$LASTEXITCODE` after each.

- **PR description doesn't match the diff.** "Disable listener-tproxy in default features" — there's no `Cargo.toml` feature change in the diff; `meow-app` default still pulls `full` → `listener-tproxy`, and the `firewall.rs` changes suggest the tproxy module compiles on Windows via the fallback `PlatformGuard`. Either the bullet is stale or a commit is missing; please reconcile. (Also note default features include `boring-tls`, so the Windows CI job builds BoringSSL — needs CMake/NASM on the runner and will be slow; worth watching.)

## Code quality

- **Hand-rolled `MIB_UDP6ROW_OWNER_PID` duplicates a type windows-sys already exports** under `Win32_NetworkManagement_IpHelper`. The layout is correct, but import it instead — a hand-copied struct is a silent-UB risk if the definition ever drifts.

- **Three near-identical table-walk functions.** `find_pid_in_udp_table` and `find_pid_in_udp6_table` differ only in address family and row type, and both duplicate `find_pid_in_table`'s probe/allocate/retry loop. One generic walker (or just parameterizing the UDP pair) would cut ~80 lines.

- **Two trait designs for the same job.** `TableRow` uses never-constructed tag structs (`TcpRow(MIB_TCPROW_OWNER_PID)` wrapping an associated `type Row`), while `UdpRow` is implemented directly on the row types. The `UdpRow` shape is simpler and works for TCP too — use it for both.

- **Misleading safety comment** in `parse_table`: "the allocator over-aligns for common types" is not a guarantee (`Vec<u8>` has alignment 1). `read_unaligned` alone is what makes it sound; the comment should say only that.

- Minor: the 1024-`u16` path buffer silently truncates long paths (`GetModuleFileNameExW` doesn't error on truncation — another point for `QueryFullProcessImageNameW`, which does). And `addr_matches` (TCP) accepts v4-mapped-v6 while the UDP path doesn't — probably fine, but the asymmetry is undocumented.

## Dependencies

- **`windows-sys = "0.59"` adds a third copy** — `Cargo.lock` already carries 0.52.6 and 0.60.2. Depend on `0.60` to reuse the existing tree instead of adding a duplicate.

## What's solid

- The cfg-gating in `ech_dns.rs` is correct: `parse_nameserver_token`'s only caller is inside the existing `#[cfg(unix)]` resolv.conf block, Windows falls through to the public-resolver fallback, and the `Ipv6Addr` keep-alive const still covers the import.
- `dns_parser.rs` hosts-path handling (respecting `SystemRoot`) is right, and the error message now includes the actual path.
- FFI hygiene is careful: buffer-size retry on `ERROR_INSUFFICIENT_BUFFER` (122), per-row bounds checks before `read_unaligned`, port byte-order handling (`u16::from_be(dwLocalPort as u16)`) is the standard correct idiom, `CloseHandle` is never skipped, and `OpenProcess` failure is correctly checked against null (not `INVALID_HANDLE_VALUE`).
- The `#[allow(clippy::unnecessary_wraps, reason = …)]` follows the project's no-silent-allows lint policy, and gating `systemd_config_test.rs` with `#![cfg(unix)]` is the right fix.

## Test coverage

- Enabling the existing lookup test on Windows (with the `uid` assertion gated) is good, but it only exercises TCP/IPv4. The UDP tables, the IPv6 row parsing, and the wildcard-match logic have no coverage — a UDP variant of the listener test would be cheap and would catch the byte-order/layout mistakes this kind of code invites.

## Verdict

Direction is right and the unsafe code is more careful than most first Windows ports. Before merge: (1) switch to `QueryFullProcessImageNameW` or attach a green Windows CI run proving `GetModuleFileNameExW` works with a LIMITED handle, (2) fix the pwsh failure-masking in the CI step, (3) reconcile the "disable listener-tproxy by default" claim with the actual diff, (4) dedupe `windows-sys` to 0.60.

---

madeye:
# Re-review (after `7838e334`…`6957d465`)

All four blocking items from the previous review are addressed, and the code-quality feedback was taken too. This is now in good shape — one cosmetic nit and two open items that are about verification, not code.

## Feedback resolution

- **`QueryFullProcessImageNameW`** — done. `get_process_info` now uses it with `PROCESS_NAME_WIN32`, handles the in/out `size` parameter correctly (capacity in, chars-written out, excluding the NUL), and checks the return against 0. The unused `Win32_System_ProcessStatus` feature was dropped from `Cargo.toml`. The undersized-buffer case correctly falls through to `None`.
- **pwsh failure masking** — done. The nine test commands are now separate steps, each with `shell: bash` (GitHub runs bash with `-e -o pipefail`, and one command per step means nothing can be masked either way).
- **PR description** — rewritten and now matches the diff: the tproxy bullet is gone, replaced by an accurate explanation that tproxy compiles on Windows via the existing fallback `PlatformGuard` with default features unchanged.
- **`windows-sys` dedup** — done, now `0.60`, and `Cargo.lock` reuses the existing 0.60.2 tree instead of adding a third copy.
- **Code quality** — the hand-rolled `MIB_UDP6ROW_OWNER_PID` is replaced by the windows-sys import, the two trait designs collapsed into one `RowAccess` trait implemented directly on the four row types, `parse_table` is now shared between TCP and UDP, and the misleading alignment comment was fixed. The refactor cut the module from ~360 to ~240 lines.

## Remaining items

- **Nit: `is_udp_target` is misnamed and its doc comment is stale.** After unifying `parse_table`, this one function now serves both TCP and UDP, but the name says UDP-only and the comment describes two behaviors ("For TCP: … For UDP: same plus…") when there's only one implementation. Rename to something like `addr_matches` and collapse the comment. Side note: this unification means UDP matching now also accepts v4-mapped-v6 addresses, which it didn't before — that's a harmless (arguably correct) widening, not a regression.
- **UDP test coverage** — still only the TCP/IPv4 path is exercised by the enabled-on-Windows test. A UDP-socket variant would be cheap and is the only automated check on the UDP table layout. Not blocking, but worth a follow-up.
- **CI** — the new `windows` job hasn't reported yet; it gates everything above, including the `QueryFullProcessImageNameW` behavior via the process-lookup unit test. I'd wait for a green `windows` job before merging; that's the only real remaining gate.

**Verdict: approve once the Windows CI job goes green** — the rename nit can ride along with any other change or be ignored.

---

madeye (approved):

madeye (approved):

madeye (approved):
All review feedback addressed (QueryFullProcessImageNameW, CI step splitting, windows-sys 0.60 dedup, RowAccess unification, UDP/IPv6 test coverage, addr_matches rename) and all 9 CI checks are green on 22d12200, including the new windows job. Approving.

xmh0511:
@madeye  can you merge this PR, then I can rebase https://github.com/madeye/meow-rs/pull/313

Issue #221, opened 2026-06-15
REGRESSION: parking-mutex swap (lwip 1dd61a5b) deadlocks tunnel data path on background sleep/wake


madeye:
## Root cause found: regression from the parking-mutex swap (lwip `1dd61a5b`)

This is a **regression**, not a fresh recurrence. The build from 5 days prior (lwip `dab00c54`, the spin-then-`yield_now` lock) did **not** freeze; the freeze appeared only after lwip was bumped to `1dd61a5b` "make LWIP_MUTEX a real parking mutex."

**Mechanism.** `LWIP_MUTEX.lock()` changed from a never-blocking bounded-spin-then-`yield_now()` loop to `parking_lot::Mutex::lock()`, which **parks (blocks) the tokio worker OS thread** on contention. That global lock is taken on every lwIP entry point and, critically, is **held across `poll_recv()`** in the poll paths and **re-acquired inside `Drop`** impls that run on worker threads:

- `rust/stack_impl.rs:100` `NetStackImpl::poll_next` — across `rx.poll_recv`
- `rust/tcp_stream_impl.rs:166` `poll_read` — across `read_rx.poll_recv` loop
- `rust/udp.rs:131` `UdpSocket::poll_next` — across `rx.poll_recv`
- `rust/stack_impl.rs:84`, `rust/tcp_stream_impl.rs:209`, `rust/tcp_listener_impl.rs:80` — `Drop` impls

On the 4-worker iOS runtime, a post-sleep/wake replay burst (the ingress backlog draining through `poll_flush → input_fn` holds the lock for a while) makes the other workers **park** on `lock()`, collapsing the pool and wedging the data path — while the process/runtime stay alive (VPN connected, RSS flat). The old lock never removed a worker from the scheduler, so it couldn't produce this.

The livelock the parking swap was meant to fix had a **different** root cause — a herd of leaked `sys_check_timeouts` timer tasks — already fixed independently by aborting the timer task in `NetStackImpl::drop`. With a single timer task the spin-then-yield lock has near-zero contention and does not livelock.

## Fix

Revert the lock mechanism to spin-then-`yield_now` (keep the timer-abort fix), drop the `parking_lot` dep: madeye/lwip `78566b68` on branch `fix/revert-parking-mutex-deadlock`. iOS re-pinned and rebuilt; on-device confirmation in progress.

[...]

PR #313, opened 2026-07-11
Improve compatibility with mihomo

xmh0511:
- select_proxy_member_async: return 400 for non-Selector groups (matching mihomo behavior)
- Built-in UI: add API secret input field with localStorage persistence
- api() function: include Authorization Bearer header when secret is set
- Update test expectations for non-Selector group PUT requests

改善与mihomo的兼容性，可以直接通过替换clash-nyanpasu的mihomo内核进行图形化使用

图片

图片


madeye (changes requested):
Thanks — getting meow-rs to work as a drop-in mihomo core for clash-nyanpasu is a great goal, and several changes here are close to mergeable: the 400-for-non-Selector PUT (with the `SelectResult` enum replacing the ambiguous `Option<bool>` — nice cleanup), the `v`-prefixed `/version`, the JSON hello body, the `delay` field on `ProxyInfo`, and the API-secret input in the built-in UI all match mihomo behavior and come with updated tests.

## Branch structure (blocking)

This PR currently contains most of #310's Windows work (Win32 `process_lookup`, service stubs, signal handling, hosts parsing, `Cargo.lock`, default-feature changes) — but not all of it, so it likely doesn't build on Windows by itself, and it conflicts with #310 in either merge order. Please rebase this branch onto upstream `main` with **only** the mihomo-compat commits (API routes, built-in UI, GLOBAL selector, CLI flags) and leave all Windows work to #310.

The default-feature changes to `meow-app`/`meow-listener` also regress Linux builds (dropping `listener-tproxy`, and the hand-picked list already omits `vless-encryption` from #317) — detailed feedback on that is in #310; it should not be in this PR at all.

## Silently-ignored CLI flags are a security problem (blocking)

The new flags — `--secret`, `--ext-ctl`, `--ext-ui`, `--config-string`, `--ext-ctl-*`, `--post-up/down`, `--age-secret-key`, `-m` — are parsed and then **ignored**. The worst case is `--secret`: a frontend passing `--secret X` believes the REST API is now authenticated, but meow ignores it and the controller runs with whatever the config file says (possibly no auth) while the GUI reports it's secured. At minimum `--secret` and `--ext-ctl` must actually override the config; the others should either work or hard-error, never silently no-op.

## CLI identity

`#[command(name = "mihomo", disable_version_flag = true)]` changes `--help` output for *all* users and removes `--version`/`-V` entirely (breaking existing scripts). Clap's `name` is cosmetic — nyanpasu invokes the binary by path — so please keep `name = "meow"`, keep the standard version flag, and add `-v` as an additional flag for the mihomo-style output line.

## GLOBAL selector

- `proxies` is a `HashMap<SmolStr, _>`, so GLOBAL's member order is nondeterministic — it reshuffles on every run/rebuild, which frontends render as a randomly reordering list. Sort the names, or preserve config order.
- clash/mihomo's GLOBAL includes DIRECT; excluding the built-ins is a small deviation worth at least a comment.
- Placement is fine — all load paths (initial load included) funnel through `rebuild_from_raw_impl`, so GLOBAL appears consistently.

## `/configs` hardcoded values

`sniffing: true`, `tcp_concurrent: false`, and `find_process_mode: "off"` (while process rules actually work) are reported regardless of actual config, and frontends will display them as truth. Please read the real values where they exist, or omit fields meow doesn't have, rather than fabricating constants.

## Minor

- The secret `<input>` uses inline styles where the rest of the file uses the `<style>` block.
- `hello()` allocates a `String` per request; a `&'static str` field would do.
- No test covers GLOBAL beyond one count assertion, and none covers the new CLI flags — a test asserting `--secret` actually protects the API would lock in the fix for the blocking issue above.

CI hasn't run on this PR yet; once it's rebased to compat-only we'll approve the workflow run.

xmh0511:
@madeye I have made changes according to these requirements

madeye (approved):
## Review

### Overview

This PR improves mihomo API/CLI compatibility so meow-rs can be dropped in as the kernel for clash-nyanpasu and similar dashboards:

- `GET /` returns `{"hello":"meow"}` JSON; `GET /version` prefixes `v`
- `PUT /proxies/{name}` on a non-Selector group now returns **400** instead of 404 (matches mihomo), via a new `SelectResult` enum replacing the `Option<bool>` tri-state
- `/proxies` gains a `delay` field from the adapter's delay history
- `GET /configs` gains `redir-port`, `tproxy-port`, `allow-lan`, `bind-address`, `ipv6`
- Auto-creates a `GLOBAL` selector containing all proxies when the user config doesn't define one
- mihomo-style CLI flags (`-v`, `--secret`, `--ext-ctl`, `--ext-ui`, `--config-string`, plus accepted-but-unsupported ones that hard-error)
- Built-in UI: API-secret input persisted in `localStorage`, sent as `Authorization: Bearer`

### Verification

Ran locally on this branch (`bad880c`): `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test --lib` (913 passed), `cargo test -p meow-api --test api_test` (95 passed), `cargo test -p meow-config --test config_persistence_test` (14 passed), `cargo test -p meow-app --test systemd_config_test` (9 passed) — all green.

**The failing `test` CI check is a runner-infra flake**, not a code failure: it died at the "Systemd config tests" step with `No space left on device` while unpacking crates; every PR-affected suite (API integration, config persistence) passed in the same run, and the systemd suite passes locally on this branch.

### Issues

1. **CLI override ordering bug** (`crates/meow-app/src/main.rs`): `--config-string` replaces the whole `config` *after* `--secret` / `--ext-ctl` / `--ext-ui` overrides are applied, silently discarding them when combined. mihomo applies flag overrides on top of whichever config source is used. Fix: load `--config-string` first, then apply the field overrides. Worth a fast-follow.

2. **`-v` pre-parse scan matches anywhere**: `std::env::args().any(|a| a == "-v")` fires even when `-v` appears as an option *value* (e.g. a file literally named `-v`), and the flag is invisible in `--help`. A clap-level solution (e.g. `short_flag` on a version arg) would be cleaner. Low risk in practice.

3. **Inconsistent unsupported-flag policy**: `--post-up`/`--post-down`/`--age-secret-key` hard-error, while `-m/--geodata-mode`, `--ext-ctl-tls/-unix/-pipe/-routing-mark` are silently ignored. If the silent ones are accepted for nyanpasu launch-compat, a `warn!` would make that visible; otherwise they should error too.

4. **GLOBAL selector omits provider proxies**: it snapshots `proxies` keys at build time, so proxies that arrive via subscription providers later aren't members (mihomo includes them). Also members are alphabetically sorted rather than config-ordered as in mihomo. Cosmetic for dashboards, worth noting.

5. **`/proxies` allocation cost**: `delay_history()` returns a fresh `Vec` per proxy just to read `.last()` — on a dashboard polling `/proxies` every second with hundreds of proxies this compounds the known per-poll build cost (#221). A `last_delay()` accessor on the trait would avoid the Vec. Fine for now given the endpoint already allocates per call.

6. **Minor**: `bind-address` defaults to `"0.0.0.0"` where mihomo reports `"*"`; `redir-port` hardcoded 0 (fine as a stub); `log-level` still hardcoded `"info"` (pre-existing).

### Security

- `--secret` on the command line is visible in the process list — same trade-off mihomo makes, acceptable for compat.
- Storing the API secret in `localStorage` matches standard clash dashboard practice (yacd/metacubexd); input is `type="password"`. OK.

### Verdict

Solid, well-scoped compat work with tests updated to match. Issue 1 is a real (if narrow) correctness bug; the rest are polish. None blocks merge — recommending merge with issue 1 as a follow-up.

PR #322, opened 2026-07-13
fix(test): eliminate port-allocation race in shadowsocks integration tests

madeye:
## Problem

The simple-obfs integration tests (`test_ss_tcp_relay_with_builtin_obfs_http` and friends) flaked in CI with two signatures:

- `read_exact failed: Kind(UnexpectedEof)` — [run 29188056796](https://github.com/madeye/meow-rs/actions/runs/29188056796/job/86637756763)
- `TCP dial failed: Connection refused` — [run 29177464036](https://github.com/madeye/meow-rs/actions/runs/29177464036)

## Root cause

`free_port()` drew server ports from the kernel ephemeral range via `bind(":0")` and released the port immediately. With a SIP003 plugin, that port stays unbound for ~100ms until ssserver spawns `obfs-server` and the plugin finishes initializing. The five tests run concurrently and all hammer the same ephemeral range with `free_port()` and echo-server `bind(":0")` calls, so during that window the kernel can hand the same port to another test. The bare `TcpStream::connect` readiness probe can't distinguish that impostor listener from the real obfs-server, so the test proceeded and then dialed a port its own server never bound.

The proof is in run 29177464036's timeline: the test panicked **7ms** after `running 5 tests` (readiness probe "succeeded"), while its own obfs-server only logged `listening at 127.0.0.1:45583` ~100ms **after** the panic. simple-obfs's `SO_REUSEPORT` (`tcp port reuse enabled`) additionally allows two colliding obfs-server instances to silently double-bind, which produces the `UnexpectedEof` variant when a connection lands on the wrong (e.g. `obfs=tls` vs `obfs=http`) instance.

The plain `test_ss_tcp_relay` never flaked because ssserver binds its port almost instantly — the plugin tests are the ones with the wide race window.

## Fix

- Allocate test server ports from a process-wide atomic counter starting at 21000 — below the Linux ephemeral range (32768–60999), so kernel `bind(":0")` assignments can never collide — and bind-check each candidate to skip ports held by unrelated processes.
- Widen the readiness window from 5s to 10s for slow plugin startup on loaded CI runners.

## Verification

- 21 consecutive local runs of the full 5-test suite with `MIHOMO_REQUIRE_INTEGRATION_BINS=1` and real `ssserver`/`obfs-server`/`obfs-local` binaries: all green.
- Full regression bar: `cargo fmt --check`, three-way clippy (`default` / `--no-default-features` / `--all-features`) with `-D warnings`, `cargo doc` with `-D warnings`, `cargo test --lib` — all pass.

### DIFF
--- a/crates/meow-proxy/tests/shadowsocks_integration.rs
+++ b/crates/meow-proxy/tests/shadowsocks_integration.rs
@@ -140,10 +140,13 @@ async fn start_ssserver_inner(
         args.push(opts.to_string());
     }
 
-    let child = Command::new("ssserver")
+    // stderr is inherited on purpose: when ssserver dies on startup (e.g. a
+    // port bind failure) the test otherwise fails with an opaque reset/EOF
+    // and no clue why.
+    let mut child = Command::new("ssserver")
         .args(&args)
         .stdout(Stdio::null())
-        .stderr(Stdio::null())
+        .stderr(Stdio::inherit())
         .kill_on_drop(true)
         .spawn()
         .expect("failed to start ssserver");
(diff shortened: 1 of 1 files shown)


madeye:
Follow-up commit 2b04614: the first CI round fixed the obfs flakes but surfaced a second race — `test_ss_tcp_relay` got `ConnectionReset` on its first write 3.6ms after the suite started ([run 29215465986](https://github.com/madeye/meow-rs/actions/runs/29215465986/job/86710516983)). `ssserver -U` binds TCP **and UDP** on the same port, and `free_port()` only bind-checked TCP: a UDP-only conflict lets ssserver bring up its TCP listener (probe and dial connect fine) and then abort on the UDP bind, resetting the accepted connections.

Hardening:
- `free_port()` now bind-checks both TCP and UDP, and staggers its base port by PID.
- The readiness loop checks `child.try_wait()` and re-probes 50ms after success, so a server dying right after binding fails with a clear "ssserver exited" panic instead of an opaque reset.
- ssserver's stderr is inherited instead of nulled, so any future startup failure prints its actual error in the CI log.

PR #329, opened 2026-07-14
Adding release windows platform binary by workflow

xmh0511:
Add release windows platform binary by GitHub workflow.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,16 +16,18 @@ permissions:
 jobs:
   build:
     name: Build ${{ matrix.target }}
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
     strategy:
       fail-fast: false
       matrix:
         include:
           - target: x86_64-unknown-linux-musl
+            runner: ubuntu-latest
             # x86_64: informational only, no size gate
             minimal_budget_bytes: 0
             minimal_soft_gate: true
           - target: aarch64-unknown-linux-musl
+            runner: ubuntu-latest
             # aarch64: hard gate — release fails if minimal binary exceeds 8 MiB
             minimal_budget_bytes: 8388608
             minimal_soft_gate: false
@@ -35,6 +37,7 @@ jobs:
               aarch64_generic aarch64_cortex-a53 aarch64_cortex-a72
               aarch64_cortex-a76
           - target: armv7-unknown-linux-musleabihf
+            runner: ubuntu-latest
             # armv7: soft gate only for now (no ADR-0007 budget yet)
             minimal_budget_bytes: 0
             minimal_soft_gate: true
@@ -48,6 +51,10 @@ jobs:
             openwrt_archs: >-
               arm_cortex-a7_neon-vfpv4 arm_cortex-a8_vfpv3
               arm_cortex-a9_vfpv3-d16 arm_cortex-a15_neon-vfpv4
+          - target: x86_64-pc-windows-msvc
+            runner: windows-latest
+            minimal_budget_bytes: 0
+            minimal_soft_gate: true
           # mipsel-unknown-linux-musl: temporarily disabled — `rust-std` is no
           # longer published for this target on stable. Re-enable when stable
           # ships the component again, or when ADR-0007 §1 promotes it to a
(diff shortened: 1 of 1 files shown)


xmh0511:
@madeye Please review this PR and publish a new version.

.github/workflows/release.yml
```diff
@@ -105,23 +120,34 @@ jobs:
...
-          cp README.md LICENSE "${staging}/" 2>/dev/null || true
-          tar -C dist -czf "${name}.tar.gz" "${name}"
-          sha256sum "${name}.tar.gz" > "${name}.tar.gz.sha256"
-          echo "archive=${name}.tar.gz" >> "$GITHUB_OUTPUT"
-          echo "checksum=${name}.tar.gz.sha256" >> "$GITHUB_OUTPUT"
+
+          if [ "${{ runner.os }}" = "Windows" ]; then
+            cp "target/${{ matrix.target }}/release/meow.exe" "${staging}/"
+            cp README.md LICENSE "${staging}/" 2>/dev/null || true
+            pwsh -Command "Compress-Archive -Path '${staging}/*' -DestinationPath '${name}.zip'"
```
madeye: `-Path '${staging}/*'` zips the *contents* of the staging directory, so the zip extracts `meow.exe`, `README.md`, `LICENSE` loose into the current directory. The tarballs all contain a `meow-<ref>-<target>/` top-level folder. Passing the directory itself keeps the layouts consistent and avoids splatting files on extraction:

```suggestion
            pwsh -Command "Compress-Archive -Path 'dist/${name}' -DestinationPath '${name}.zip'"
```

.github/workflows/release.yml
```diff
@@ -105,23 +120,34 @@ jobs:
...
+            pwsh -Command "(Get-FileHash -Algorithm SHA256 '${name}.zip').Hash.ToLower() + '  *${name}.zip'" > "${name}.zip.sha256"
```
madeye: This produces a `.sha256` that `sha256sum -c` can't verify, for two reasons:

1. The GNU checksum line format is either `hash␣␣file` (text mode) or `hash␣*file` (binary mode). This writes `hash␣␣*file` — two spaces *and* an asterisk — so `-c` parses the filename as literally `*meow-….zip` and reports "no such file".
2. `pwsh` on Windows emits CRLF, so the redirected file gets a trailing `\r`, which also trips verification on Linux/macOS.

The step already runs under `shell: bash` (Git Bash), which ships `sha256sum` — using the same line as the Linux branch fixes both and drops the pwsh detour:

```suggestion
            sha256sum "${name}.zip" > "${name}.zip.sha256"
```

madeye (commented):
Thanks for adding Windows to the release matrix! The overall wiring looks solid:

- Archives are passed to `upload-artifact` via `steps.pkg.outputs`, and the new `dist/*.zip` globs in the release job line up with `download-artifact`'s `merge-multiple: true` layout. `fail_on_unmatched_files: true` stays safe since the Windows job is unconditionally in the matrix.
- Build feasibility is already covered: `test.yml`'s `windows` job builds/clippy-checks with default features (including `boring-tls`) on `windows-latest`.
- `runs-on: ${{ matrix.runner }}` + `runner.os == 'Linux'` guards is the right pattern — the minimal-size steps use GNU `stat -c%s` and would break on Windows.

Two packaging bugs need fixing before merge (inline comments with suggestions):

1. **The Windows `.sha256` file is malformed** — `sha256sum -c` can't verify it (wrong marker format + CRLF from pwsh).
2. **Zip layout is inconsistent with the tarballs** — it extracts files loose instead of into a `meow-<ref>-<target>/` folder.

Both fixes make the step simpler, not bigger.

Minor / non-blocking:

- The `cp README.md LICENSE` line is now duplicated in both branches; the staging setup (`cp` of the binary aside) could move above the `if` so only the archive command differs. With the sha256 fix below, the checksum + `$GITHUB_OUTPUT` lines can be unified too (only the extension differs).
- `${{ github.event.inputs.tag }}` now also lands inside a `pwsh -Command` string; a dispatch input containing quotes could break out of the single-quoted context. Pre-existing pattern in this workflow (same input already flows into bash) and `workflow_dispatch` needs write access, so low severity — and dropping the pwsh checksum line removes one of the two new sinks.
- `minimal_soft_gate: true` on the Windows entry is dead config (budget 0 already skips both gated steps) — harmless, consistent with the x86_64 Linux entry.
- No code signing on the Windows binary means SmartScreen warnings for users; worth a line in the release notes eventually, not a blocker here.

madeye (approved):

madeye:
@xmh0511 Released in [v0.17.0](https://github.com/madeye/meow-rs/releases/tag/v0.17.0) — the Windows binary (`meow-v0.17.0-x86_64-pc-windows-msvc.zip`) is on the release page. Thanks for the contribution!

PR #331, opened 2026-07-14
Windows service

xmh0511:
The Windows platform now supports installing, uninstalling, and querying the status of the service.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -273,6 +273,10 @@ jobs:
         shell: bash
         run: cargo test --lib
 
+      - name: App binary unit tests (windows service)
+        shell: bash
+        run: cargo test -p meow-app --bin meow -- --nocapture
+
       - name: Common types tests (windows)
         shell: bash
         run: cargo test --test common_test -- --nocapture
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,6 +76,9 @@ jobs:
       - name: Unit tests
         run: cargo test --lib
 
+      - name: App binary unit tests
+        run: cargo test -p meow-app --bin meow -- --nocapture
+
       - name: Common types tests
         run: cargo test --test common_test -- --nocapture
 
@@ -328,6 +331,7 @@ jobs:
       - name: Core smoke tests
         run: |
           cargo test --lib
+          cargo test -p meow-app --bin meow -- --nocapture
           cargo test --test common_test -- --nocapture
           cargo test --test config_test -- --nocapture
           cargo test --test rules_test -- --nocapture
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -805,6 +805,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "790eea4361631c5e7d22598ecd5723ff611904e3344ce8720784c93e3d83d40b"
 
 [[package]]
+name = "crossbeam-channel"
+version = "0.5.16"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d85363c37faeca707aef026efa9f3b34d077bce547e48f770770625c6013679e"
+dependencies = [
+ "crossbeam-utils",
+]
+
+[[package]]
 name = "crossbeam-deque"
 version = "0.8.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2242,7 +2251,10 @@ dependencies = [
  "tempfile",
  "tokio",
  "tracing",
+ "tracing-appender",
  "tracing-subscriber",
+ "windows-service",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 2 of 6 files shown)


xmh0511:
@madeye Have a review for this PR.

madeye:
## Review — Windows service support

### Overview

Adds full Windows Service Control Manager (SCM) support to `meow-app`: `install` / `uninstall` / `status` replace the previous "not yet supported" stubs, plus a hidden `run-service` subcommand the SCM invokes. The main run path is refactored into `run_application` with pluggable `LogTarget` / `ShutdownSignal` / `on_ready` seams so console and service modes share one code path. Service mode logs to daily-rolling files under `%ProgramData%\meow\logs` (7 retained) via `tracing-appender`; CI gains a Windows step running the binary's unit tests.

### Strengths (verified)

- **SCM state machine handling is unusually careful**: startup checkpoints advance on a dedicated reporter thread, Stop/Shutdown controls feed a oneshot into the shared shutdown wait, and `ERROR_SERVICE_MARKED_FOR_DELETE` / `NOT_ACTIVE` / `CANNOT_ACCEPT_CTRL` are all handled.
- The terminal error is logged and the non-blocking writer guard dropped (flushed) **before** the Stopped status reaches SCM — with `service_startup_error_is_flushed_before_ready` proving that ordering.
- `-d` propagation works in service mode: `install` bakes absolute `-f`/`-d` into the launch arguments (correct, since services get `system32` as cwd), and `run_application_inner` still runs `set_home_dir` + config resolution, so the service resolves resources like a CLI launch.
- Uninstall marks-for-deletion *before* stopping so recovery actions can't respawn the service mid-uninstall.
- Console path refactor is behavior-preserving (unix SIGTERM/ctrl-c select intact); `init()` → `try_init()` is a strict improvement.

### Issues / suggestions

1. **Inconsistent relative `-f` resolution between `install` and direct run** (`crates/meow-app/src/windows_service.rs`, `install`). At runtime, `-d D:\data -f config.yaml` loads `D:\data\config.yaml`; but `install` resolves a relative `-f` against the **cwd** via `absolute_path()`, ignoring `-d`. Same flags, different config registered. Suggest mirroring the runtime rule: if `-d` is set and `-f` is relative, resolve `-f` under `-d`. *This is the one fix I'd request pre-merge.*
2. **`canonicalize()` bakes `\\?\`-prefixed paths** into the SCM command line, install summary, and logs. Works with Rust `std::fs`, but ugly and occasionally trips other tooling. `std::path::absolute` (stable) or `dunce::canonicalize` avoids it; existence is already checked separately.
3. **Persistent-failure restart loop.** With `set_failure_actions_on_non_crash_failures(true)` and a single `Restart(5s)` action, SCM repeats the last action on every failure — a permanently broken config restarts every ~5s forever (the 24h reset never elapses). Consider an escalating action list (5s, 30s, 60s).
4. **Logs are world-readable.** `%ProgramData%\meow\logs` inherits default ACLs, so non-admin users can read proxy logs containing destination domains. Consider tightening the ACL at install time or adding a README caveat.
5. **LocalSystem executes the install-time binary path** — typically a user-profile `target\release` path. Quoting is handled by the `windows-service` crate, but a README note recommending copying the binary to an admin-only location (e.g., `Program Files`) before installing would be appropriate.

### Minor nits

- The startup-error test hardcodes `mixed-port: 17890`; port 0 or a reserved value would be collision-proof on shared CI runners.
- The new `--bin meow` test step runs only in the Windows job — fine for the new gated tests, but the pre-existing `bind_socket_addr` bin tests remain unexercised on Linux/macOS; adding the step to the other jobs is cheap.
- Cosmetic race: a Stop arriving between handler registration and the first status report briefly re-reports `StartPending` after the phase flipped to `Stopping`. Harmless — no controls accepted in that state, and `on_ready` bails correctly.

### Quality, style, tests

- Follows project conventions: `.context()` with actionable messages ("run PowerShell as Administrator"), inline format args (clean against the workspace clippy set), `parking_lot` reused from existing deps, new deps gated under `[target.'cfg(windows)'.dependencies]` so non-Windows builds and binary size (ADR-0007) are untouched.
- Test coverage is genuinely good: unicode-path/launch-arg-order tests, home-resolution tests, SCM status invariants, shutdown-signal test, and the log-flush-ordering integration test. Real SCM interaction is inherently manual; the README PowerShell walkthrough covers it.

**Verdict:** approve with minor changes — fix #1 pre-merge; #2–#5 are fine as follow-ups.

madeye:
Pushed 2d1597a addressing the review findings directly (thanks for enabling maintainer edits):

- **#1** `install` now resolves a relative `-f` under `-d`, matching the runtime rule (new `resolve_service_config` helper + unit test)
- **#2** `std::path::absolute` replaces `canonicalize()` so `\\?\` extended-length paths are no longer baked into the SCM command line
- **#3** recovery actions escalate 5s → 30s → 60s, so a persistently failing service settles at a 60s retry cadence
- **#4/#5** README security notes on the LocalSystem binary location and world-readable `%ProgramData%` logs
- nit: the app-binary unit tests now also run in the Linux/macOS CI jobs

Verified locally: fmt, three-way clippy, docs, `cargo test --lib`, and a full `x86_64-pc-windows-gnu` cross-compile of `meow-app --all-targets` (warning-free). Will merge once CI is green.

madeye (approved):
CI green across all jobs including Windows; review findings addressed in 2d1597a. Approving.

Issue #72, opened 2026-05-13
未来会支持tun么？



thomasdba:
如果支持hy2就更好了

madeye:
TUN inbound is **out of scope** — confirmed not planned. Hy2 outbound is still on the table; tracking it separately. (Will retitle / close as appropriate once hy2 lands or is also dropped.)

madeye:
Adding context on why TUN won't ship: the project already supports **transparent proxy (tproxy)** as an inbound mode, which covers the same "capture everything at the network layer" use case TUN is typically requested for. Maintaining a TUN driver per platform (Linux \`tun\` / macOS \`utun\` / iOS provider extension) on top of tproxy would be redundant.

Hy2 outbound remains under consideration — its scope is unrelated to TUN.

thomasdba:
got it ,thanks

PR #332, opened 2026-07-14
feat(listener): TUN inbound for Windows transparent proxy (#326)

### LINKED ISSUE #326: The suggestion of implementing tproxy on the Windows platform
We can use [tun-rs](https://crates.io/crates/tun-rs) and [route_manager](https://crates.io/crates/route_manager), which are both cross-platform libraries, to implement tproxy for the Windows platform. 

Or, directly references [seeker](https://github.com/gfreezy/seeker), which library also uses `tun-rs`.

madeye:
Implements #326 — a TUN-based transparent-proxy inbound using the crates suggested in the issue ([tun-rs](https://crates.io/crates/tun-rs) + [route_manager](https://crates.io/crates/route_manager)), plus [ipstack](https://crates.io/crates/ipstack) as the userspace TCP/IP stack (the piece tun-rs alone doesn't provide; it's what tun2proxy uses). This is the tproxy equivalent for platforms without a REDIRECT firewall — Windows first and foremost — and unlike tproxy it also relays **UDP**. Works identically on Linux/macOS.

> Note on scope: #72 ruled TUN inbound out of scope while tproxy covered the capture use case. Windows has no tproxy/REDIRECT equivalent, so this lands TUN as the Windows transparent-proxy path (owner-directed).

## Design

```
tun: (config)  →  tun-rs device  →  ipstack userspace stack
                                        ├─ TCP flow → meow_tunnel::tcp::handle_tcp (existing path:
                                        │             fake-IP rewrite, lazy rule match, stats, zero-alloc relay)
                                        └─ UDP flow → per-flow task (mirrors SOCKS5-UDP routing)
                                                       └─ :53 + dns-hijack → DnsServer::handle_query in-process
```

**Loop freedom (v1: fake-IP-scoped capture).** The classic TUN failure mode is the routing loop — a global default route into the device makes meow's own outbound dials re-enter the tun. v1 side-steps the problem class: `auto-route` installs **only the fake-IP-range route**. DNS enters the tun (OS resolver pointed inside the range) and is hijacked to fake IPs; client connections to fake IPs are captured and rules match on the recovered domain; outbound dials — proxy upstreams *and* DIRECT — always target real IPs, which take the physical route and cannot loop. No SO_MARK / interface binding / bypass routes. Trade-off: IP-literal traffic isn't captured; global capture needs outbound loop protection and is left as follow-up. Full rationale + setup guide in `docs/tun.md`.

## Changes

- **meow-listener** — new `tun/` module behind opt-in feature `listener-tun` (in the `full` bundle): device adapter (`AsyncRead/AsyncWrite` over `tun_rs::AsyncDevice`), accept loop, TCP → `handle_tcp`, per-flow UDP relay, DNS hijack, RAII route guard.
- **meow-config** — mihomo-compatible top-level `tun:` section (`enable`, `device`, `mtu`, `inet4-address`, `auto-route`, `dns-hijack`, `udp-timeout`). Upstream-only fields (`stack`, `strict-route`, `auto-detect-interface`, …) warn-and-ignore per the #328 no-silent-compat policy; `mtu < 1280` / `udp-timeout: 0` are hard errors.
- **meow-common** — `ConnType::Tun`.
- **meow-dns** — `Resolver::fake_ip_v4_net()` accessor.
- **meow-app** — spawns the listener from `tun.enable`; builds without the feature log a warning instead.

## Footprint

No ADR-0011 key types touched. The three new deps compile only behind `listener-tun`, so the `minimal` profile and its ADR-0007 size gate are unaffected. `full` grows by the tun stack (opt-in bundle, informational gate only).

## Verification

- `cargo fmt --check`, three-way `clippy -D warnings` (default / no-default / all-features), `RUSTDOCFLAGS="-D warnings" cargo doc`, `cargo test --lib` — all green
- 10 new parser tests (`crates/meow-config/tests/tun_config_test.rs`, T1–T10)
- Full Windows binary cross-builds: `cargo zigbuild --target x86_64-pc-windows-gnu --no-default-features --features full` (msvc CI job will exercise the boring-tls default)
- Unprivileged startup fails the TUN listener with a descriptive error while other inbounds keep serving; `meow -t` validates tun configs and fires the compat warnings

**Still to do before calling the feature done end-to-end:** a privileged traffic smoke test (curl through the tun with fake-ip + dns-hijack) on real Windows and Linux hosts — I couldn't run one here (no root in this environment).

Closes #326


Cargo.toml
```diff
@@ -117,6 +117,13 @@ lru = "0.16"
 ipnet = "2"
 iprange = "0.6"
 
+# TUN inbound (issue #326) — cross-platform L3 device + userspace TCP/IP
+# stack + route table control. All three are optional, pulled in only by
+# meow-listener's `listener-tun` feature.
+tun-rs = { version = "2.8", features = ["async_tokio"] }
+ipstack = "1.0"
```
xmh0511: @madeye  Don't recommend using `ipstack`, which is not widely tested. I suggest we use `smoltcp` instead.

madeye: Done in cd28553 — swapped `ipstack` for [`netstack-smoltcp`](https://crates.io/crates/netstack-smoltcp) (the smoltcp-backed netstack clash-rs uses; raw smoltcp isn't a netstack by itself, so this is the practical reading of the suggestion).

The API shape differs, so the TUN internals were restructured rather than just re-pointed:
- TCP: accept loop now consumes the netstack's `TcpListener` stream.
- UDP: ipstack owned the per-flow NAT; the netstack yields one packet-level socket, so the listener now owns the flow table (per-flow tasks keyed by (src, dst), idle-evicted after `udp-timeout`). `dns-hijack` answering is now stateless per query.
- Device: byte-stream adapter replaced by two packet pumps between `tun-rs` and the stack's `Stream`/`Sink`.

Side effect: smoltcp answers ICMP echo into the device itself, so `ping` to a fake IP now gets a reply (docs updated). Full regression bar (fmt, 3-way clippy, rustdoc, lib + tun config tests) is green.

PR #333, opened 2026-07-14
Improve Mihomo runtime compatibility and validate connection start time

xmh0511:
This pull request introduces several important enhancements and fixes to the API, runtime selection, and test suites, with a focus on improving mihomo compatibility, runtime proxy group selection, and API shape correctness. The most significant changes are grouped below:

### Runtime Proxy Group Selection & API Compatibility

* Added the `ProxySelection` trait and extended the `Proxy` trait to support runtime selection and introspection for outbound groups, aligning with mihomo's API (`ProxySelection`, `Proxy::selection`, `Proxy::test_url`, `Proxy::expected_status`). [[1]](diffhunk://#diff-17f3dd5ad15a5a2ecb4d34b78cf6a7c5f7ed31ad9cc4a66ee71a2721b6fd5553R167-R189) [[2]](diffhunk://#diff-17f3dd5ad15a5a2ecb4d34b78cf6a7c5f7ed31ad9cc4a66ee71a2721b6fd5553R209-R223) [[3]](diffhunk://#diff-b4990cad533d8499c995e63cbc8ab7ee4d47e213419491adbc165c07be2d4c28L16-R18)
* Updated the proxy group parser to wire runtime options and selection store into `url-test` and `fallback` groups, enabling proper API support for group selection and health check fields.
* Added a new `rebuild_from_raw_runtime` function to `meow-config` for runtime config rebuilds that retain provider slots and selection state.

### Logging Improvements

* Added dynamic log level reloading: introduced a `LogReload` mechanism in `log_stream.rs`, and updated the main app to support runtime log level changes via the API. [[1]](diffhunk://#diff-a752ea8efaea2ec97b488bcfe17e42c55468db6274575446a6aef05f4408dfbfR6-R22) [[2]](diffhunk://#diff-5723b2426194bb1c011fe025fa77ac9c2aa773f532b1f674cadc74a46fa924e5R171-R183)

### Proxy Provider Enhancements

* Added `expected_status` to `HealthCheckConfig`, tracked provider update times with an atomic field, and made provider refreshes return results and update timestamps. [[1]](diffhunk://#diff-bdfaec0d4a4f8bab113e57e8049df49908223a84339e149ad82010c0931bc272R7-R16) [[2]](diffhunk://#diff-bdfaec0d4a4f8bab113e57e8049df49908223a84339e149ad82010c0931bc272R29) [[3]](diffhunk://#diff-bdfaec0d4a4f8bab113e57e8049df49908223a84339e149ad82010c0931bc272R110) [[4]](diffhunk://#diff-bdfaec0d4a4f8bab113e57e8049df49908223a84339e149ad82010c0931bc272L228-R261) [[5]](diffhunk://#diff-bdfaec0d4a4f8bab113e57e8049df49908223a84339e149ad82010c0931bc272L255-R273)

### API Test Suite & Behavior Fixes

* Updated log WebSocket tests to match mihomo's frame shapes, removed synthetic lagged frames, and enforced stricter validation of log output. [[1]](diffhunk://#diff-e9b962cecce7c8f3f908e47c175ffd380cd2594e67393d2021b9c9ef803fdce1L179-R179) [[2]](diffhunk://#diff-e9b962cecce7c8f3f908e47c175ffd380cd2594e67393d2021b9c9ef803fdce1L188-R191) [[3]](diffhunk://#diff-e9b962cecce7c8f3f908e47c175ffd380cd2594e67393d2021b9c9ef803fdce1L298-R304)
* Added and improved tests for group selection endpoints, group proxy filtering, and DNS query output to match mihomo's API and ensure correct behavior. [[1]](diffhunk://#diff-6aca45b416abdd2f6554055961da8cd969cd1c5021ff4d21cd1ab998fb1c16c1R1242-R1347) [[2]](diffhunk://#diff-6aca45b416abdd2f6554055961da8cd969cd1c5021ff4d21cd1ab998fb1c16c1L2695-R2814) [[3]](diffhunk://#diff-6aca45b416abdd2f6554055961da8cd969cd1c5021ff4d21cd1ab998fb1c16c1L2613-R2730)
* Fixed traffic and connection API tests to check for correct field names and values, ensuring API responses match expected shapes. [[1]](diffhunk://#diff-6aca45b416abdd2f6554055961da8cd969cd1c5021ff4d21cd1ab998fb1c16c1L401-R412) [[2]](diffhunk://#diff-6aca45b416abdd2f6554055961da8cd969cd1c5021ff4d21cd1ab998fb1c16c1L463-R474)
* Changed authentication tests to reject lowercase "bearer" prefixes, enforcing stricter token validation. [[1]](diffhunk://#diff-6aca45b416abdd2f6554055961da8cd969cd1c5021ff4d21cd1ab998fb1c16c1L1377-R1493) [[2]](diffhunk://#diff-6aca45b416abdd2f6554055961da8cd969cd1c5021ff4d21cd1ab998fb1c16c1L1389-R1505)

### Dependency Updates

* Added `hickory-proto` as a workspace dependency to `meow-api`.

These changes collectively improve mihomo compatibility, enable runtime group selection and log level reloads, and ensure the API and tests match expected behaviors.


xmh0511:
@madeye Please review this PR. This PR makes a significant improvement to using `meow-rs` in nyanpasu by aligning its behavior with the mihomo API.

madeye (changes requested):
Thanks — this is substantial and genuinely useful mihomo external-controller parity work, and the `docs/mihomo-api-compatibility.md` matrix is a great addition. Blocking on the compile failure plus a couple of logging correctness issues; details below.

## Blocker: the PR does not compile

CI lint is red on the head commit: `crates/meow-app/src/main.rs` `init_logging()` was only half-rewritten — the new logging block was added at the top, but the old `match target { … }` arms below it still reference the deleted `tx` binding, the deleted `env_filter()` closure, and imports that moved inside the new block (E0425/E0422/E0599 ×7). All downstream CI jobs were skipped.

Note the old arms are also what carried the Windows-service rolling-file logging; the new block unconditionally installs a plain stdout `fmt::layer()`, so the rewrite loses file logging for the Windows service even once the compile error is fixed.

## Correctness

1. **Log-level reload breaks on mihomo's canonical level names** (`meow-app/src/main.rs` reloader + `routes.rs::update_configs`). The reloader does `EnvFilter::new(level)` with the raw string. `"warning"` and `"silent"` are not valid `EnvFilter` levels, so `parse_lossy` treats them as *target* directives (`warning=trace`), which disables logging for every real target. `PATCH /configs {"log-level":"warning"}` — exactly what mihomo dashboards send — silences the process instead of setting warn. Please map `warning→warn` and `silent→off` before building the filter.

2. **The reloadable `EnvFilter` is registry-global and gates the broadcast layer.** `registry().with(filter_layer).with(fmt).with(log_layer)` — a bare `EnvFilter` layer filters *all* layers. With the default `info` filter, DEBUG events never reach `LogBroadcastLayer`, so `GET /logs?level=debug` stops working. This contradicts the comment kept above the block ("max-level is TRACE, preventing the fmt layer's EnvFilter from silencing DEBUG/TRACE events before LogBroadcastLayer"). The filter must stay per-layer on the fmt layer (`fmt::layer().with_filter(reloadable)`), with the reload handle pointing at that per-layer filter.

3. **Partial-apply in `PATCH /configs`**: mode is applied and persisted to `raw` before log-level validation; an invalid log-level returns 400 after the mode change already took effect. Validate both fields first, then apply.

## Security / auth

- Constant-time comparison preserved in `require_auth_ws` — good. Dropping lowercase `bearer` matches mihomo and tests were updated.
- All REST routes now go through `require_auth_ws`. Since the `?token=` path is gated only on the client-controlled `Upgrade: websocket` *header* (not an actual upgrade), any REST endpoint now accepts the secret in the URL if the client sets that header. Not an auth bypass, but secrets in query strings end up in proxy/access logs, and the doc comment "`?token=` is accepted ONLY on this middleware — REST routes keep header-only auth" is now false. Consider gating on actual upgrade state, or at least fix the comment.
- The orphaned doc comment from the deleted `require_auth` is left dangling above `require_auth_ws`'s own doc — please delete it.

## Performance / ADR gates

- **The relay hot path now does a DashMap `get_mut` (hash + shard write-lock) plus two extra atomics per successful write** (`record_connection_upload/download` from the `poll_copy` progress callback). The zero-per-relay-setup-allocation invariant is preserved (closures capture a `Copy` `Uuid`, no heap), but this is per-write overhead in the tightest loop in the codebase. Per the repo policy, PRs touching `relay.rs`/`tcp.rs` need measured justification, and ADR-0006 sets a ≥0.98× throughput gate — please run the W1–W5 workloads (or at least `meow-bench` throughput) and include before/after numbers in the commit body. A cheaper design if it regresses: accumulate per-connection deltas in plain `u64`s inside the relay frame and flush to the DashMap entry on a coarse cadence (every N bytes or on the 1 s sampler).
- `chrono_now()` RFC 3339 (20 bytes) still fits SmolStr inline and avoids heap — nicely done; `ConnectionInfo` size is unchanged.
- The 1 s traffic sampler in `Tunnel::spawn_background_tasks` is fine (called once), consistent with the NAT sweeper.

## Smaller notes

- `SelectorGroup::force_set(None)` is a silent no-op, contradicting the trait doc ("Set **or clear** a selection without validation"). Unreachable today (`can_unfix()` is false and callers filter on it), but either document the asymmetry or honor `None`.
- `SelectorStore::global()` is set by whichever `SelectorStore::new()` runs first — fine for the app, but a footgun for tests/library users; worth a doc note.
- `GET /group/{name}/delay` clears the user pin *before* query validation, per the comment claiming mihomo parity — please double-check against `hub/route/groups.go`, since clearing state on a malformed request is surprising if the claim is wrong.
- `/dns/query` hardcodes `TTL: 60` for A/AAAA answers from the resolver cache (mihomo reports real TTLs); structured log frames format time as UTC `HH:MM:SS` where mihomo uses local time. Both cosmetic.
- The `fixed` semantics (fallback forgets a dead pin in memory but keeps the persisted value; url-test keeps the pin and routes around it) match upstream and are well covered by the new group unit tests — good.

## Suggested next steps

1. Finish the `init_logging` rewrite: delete the stale `match target` bodies, restore the Windows file-appender path, keep the reloadable `EnvFilter` **per-layer** on the fmt layer, and map `warning`/`silent` in the reloader. Then get CI green end to end.
2. Add relay benchmark numbers (ADR-0006) to the commit body; reconsider per-write DashMap updates if they miss the 0.98× gate.
3. Fix the stale auth doc comments and decide whether REST-with-spoofed-Upgrade should really accept `?token=`.

xmh0511:
@madeye I think we can publish a new version after this PR.

crates/meow-tunnel/src/tcp.rs
```diff
@@ -95,10 +95,18 @@ pub async fn handle_tcp(
...
     match proxy.dial_tcp(&metadata).await {
         Ok(mut remote) => {
-            match copy_bidirectional_buf(&mut conn, &mut remote, &mut buf_up, &mut buf_dn).await {
+            let id = guard.id();
+            match copy_bidirectional_buf_tracked(
+                &mut conn,
+                &mut remote,
+                &mut buf_up,
+                &mut buf_dn,
+                |n| tunnel.stats.record_connection_upload(id, n as i64),
```
madeye: **Main finding:** this is the only relay call site converted to the tracked variant. SOCKS5 (`socks5.rs:263`), HTTP proxy (`http_proxy.rs:207,326`), and TProxy (`tproxy/mod.rs:223`) still call the untracked `copy_bidirectional_buf` and add totals once at connection close.

For the most common deployments (mixed/SOCKS/HTTP inbounds) that means:
- `/connections` per-connection `upload`/`download` stay `0` for the whole lifetime of the connection;
- the new `/traffic` rate stream reads 0 during a long transfer, then spikes with the full byte count in the close second.

Since only TUN flows enter `handle_tcp`, the headline feature is effectively TUN-only. Either convert the three listener call sites too, or document the limitation in `docs/mihomo-api-compatibility.md` (it currently marks `/traffic` as Compatible without mentioning this).

crates/meow-tunnel/src/relay.rs
```diff
@@ -156,6 +159,26 @@ pub async fn copy_bidirectional_buf<A, B>(
...
     A: AsyncRead + AsyncWrite + Unpin + ?Sized,
     B: AsyncRead + AsyncWrite + Unpin + ?Sized,
+{
+    copy_bidirectional_buf_tracked(a, b, buf_a_to_b, buf_b_to_a, |_| {}, |_| {}).await
+}
+
+/// Tracked variant of [`copy_bidirectional_buf`]. The callbacks run after
+/// bytes are successfully written in each direction, allowing live traffic
+/// and connection statistics without waiting for relay completion.
+pub async fn copy_bidirectional_buf_tracked<A, B, FA, FB>(
```
madeye, line 169: The zero-per-relay-setup-allocation invariant holds (monomorphized `FnMut` in the caller's frame — nice), but this touches the ADR-0006-gated hot loop: with `RELAY_BUF_SIZE = 4 KiB`, the tracked path now does a `DashMap::get_mut` (shard write lock) plus four relaxed atomic RMWs **per 4 KiB written** — ~30k map lookups/sec per connection at 1 Gbps.

Per the repo rules, relay-path changes need before/after benchmark numbers (median ≥ 0.98× baseline, W1–W5) in the commit body. If it regresses: a cheaper design is to resolve the connection entry once at setup and hand the closure two `AtomicI64`s owned by the guard, instead of a map lookup per chunk.

crates/meow-api/src/routes.rs
```diff
@@ -485,29 +600,89 @@ struct UpdateConfigRequest {
...
+            level.to_ascii_lowercase().as_str(),
+            "debug" | "info" | "warning" | "warn" | "error" | "silent"
+        ) {
+            return msg_err(StatusCode::BAD_REQUEST, "Body invalid");
         }
     }
-    let _ = body.log_level;
-    StatusCode::NO_CONTENT
+
+    // Both valid — apply atomically.
```
madeye, line 618: Not quite atomic: mode is applied first, so if `reload_log_level` fails the handler returns 500 with the mode already changed. Validation-first covers the common case, but `reload_handle.reload()` can still fail independently. Either apply the log level first (it's the fallible one) or soften the comment.

crates/meow-api/src/routes.rs
```diff
@@ -23,11 +24,36 @@ use std::sync::Arc;
 use std::time::Duration;
 use tokio::sync::{broadcast, Mutex};
 use tower_http::cors::CorsLayer;
-use tracing::{debug, info, warn};
+use tracing::{debug, info};
 
 use crate::log_stream::{parse_log_level, LogMessage};
 use crate::ui;
 
+struct MaybeWebSocket(Option<WebSocketUpgrade>);
```
madeye, line 32: If `Upgrade: websocket` is present but the handshake is malformed (e.g. missing `Sec-WebSocket-Key`), the `.ok()` swallows the rejection and the request falls through to the infinite HTTP-streaming path. A client that *claims* to be websocket should get the 4xx back rather than a hanging chunked response — consider propagating the rejection when the Upgrade header is present.

crates/meow-api/src/routes.rs
```diff
@@ -567,10 +742,118 @@ async fn dns_query(
...
+    response.insert("Status".into(), 0.into());
+    response.insert("Question".into(), serde_json::Value::Array(vec![question]));
+    response.insert("TC".into(), false.into());
+    response.insert("RD".into(), true.into());
+    response.insert("RA".into(), true.into());
+    response.insert("AD".into(), false.into());
+    response.insert("CD".into(), false.into());
+
+    if matches!(record_type, RecordType::A | RecordType::AAAA) {
+        let ips = resolver.resolve_ips(&params.name).await.unwrap_or_default();
```
madeye, line 792: Inconsistent failure behavior between the two paths: `unwrap_or_default()` turns a total resolution failure (or NXDOMAIN) into `Status: 0` with no `Answer`, while the generic-record path below returns 500 `"DNS query failed"`. mihomo surfaces the real RCODE for A/AAAA too. Also `TTL` is hardcoded to 60 rather than the cache's remaining TTL — acceptable, but worth a comment.

crates/meow-proxy/src/group/urltest.rs
```diff
@@ -42,11 +53,59 @@ impl UrlTestGroup {
...
+            store: None,
+            test_url: "http://www.gstatic.com/generate_204".to_string(),
+            expected_status: String::new(),
             fastest: RwLock::new(None),
             health: ProxyHealth::new(),
         }
     }
 
+    #[must_use]
+    pub fn with_runtime_options(
```
madeye, line 66: The persisted selection is restored without checking membership: if a config change removed the pinned member, `fixed` keeps reporting a name that isn't in `all`. Routing is unaffected (falls back to auto), but dashboards will show a stale pin. Consider validating against the member set here (applies to the `FallbackGroup` copy of this method too).

crates/meow-proxy/src/group/selector_store.rs
```diff
@@ -36,10 +38,18 @@ impl SelectorStore {
                 HashMap::new()
             }
         };
-        Arc::new(Self {
+        let store = Arc::new(Self {
             path,
             map: Mutex::new(map),
-        })
+        });
+        let _ = GLOBAL_STORE.set(Arc::clone(&store));
```
madeye, line 45: Constructor side effect on a process-wide global, first-write-wins. In the shared `cargo test` process, whichever test calls `open()` first pins the global for every later `rebuild_from_raw_runtime` call — latent cross-test contamination (selector.rs's own tests call `open` twice). Since the store already flows through `rebuild_from_raw_impl`, plumbing it through `AppState` instead would avoid the global entirely; at minimum, document the first-wins semantics.

madeye (commented):
Thorough mihomo-alignment PR — the `ProxySelection` design is clean, the hot-path allocation discipline was respected (`chrono_now` stays inline in `SmolStr`, tracked-relay callbacks live in the caller's async frame), test coverage is real, and the compatibility matrix doc is genuinely useful. CI is green including the three-way clippy job.

Two things I'd like resolved before merge (details in the inline comments):

1. **Live per-connection/rate stats only work on the TUN path.** SOCKS5, HTTP, and TProxy listeners still run the untracked relay and count bytes at close, so `/connections` byte counters stay 0 and the `/traffic` rate stream is bursty for the most common inbounds.
2. **ADR-0006 benchmark evidence for the relay hot-path change.** The tracked path adds a `DashMap::get_mut` + 4 atomic RMWs per 4 KiB chunk on the exact loop the throughput gate covers; the PR body has no before/after numbers, which the project rules require for relay changes.

Breaking changes that deserve a changelog note (all deliberate mihomo alignment, just worth recording):
- lowercase `bearer` prefix is now rejected;
- `provider_healthcheck` dropped its `url`/`timeout`/`expected` query params in favor of provider config;
- the synthetic `lagged` log frame is gone — slow consumers now silently lose messages;
- `/logs` and `/traffic` over plain HTTP are now infinite chunked streams (a bare `curl` hangs where it used to get an upgrade error).

House-rule nit: `statistics.rs` is a size-gated file (`ConnectionInfo` @ 120 B). The type itself is untouched and `Statistics` only grew by 32 B on a singleton, but per the repo convention the commit body should state that explicitly with `-Zprint-type-sizes` output.

Security notes, both fine but deliberate: `require_auth_ws` now guards every route, so `?token=` auth is honored on any request that carries an `Upgrade: websocket` header (mildly widens the secrets-in-URLs surface beyond the two old WS routes); constant-time comparison is preserved.

crates/meow-app/src/main.rs
```diff
@@ -212,31 +212,49 @@ fn main() -> Result<()> {
...
             let log_layer = LogBroadcastLayer { tx: tx.clone() }.with_filter(LevelFilter::TRACE);
+            let (filter_layer, reload_handle) =
+                tracing_subscriber::reload::Layer::new(env_filter());
             tracing_subscriber::registry()
-                .with(tracing_subscriber::fmt::layer().with_filter(env_filter()))
+                .with(tracing_subscriber::fmt::layer().with_filter(filter_layer))
                 .with(log_layer)
-                .try_init()
-                .map_err(|e| anyhow::anyhow!("failed to initialize logging: {e}"))?;
+                .init();
```
madeye, line 251: `try_init` → `init` converts a recoverable error into a panic on double initialization. Fine for a once-per-process `main`, but worth double-checking the Windows-service restart path can't re-enter `init_logging`.

crates/meow-proxy/src/group/fallback.rs
```diff
@@ -150,6 +210,58 @@ impl Proxy for FallbackGroup {
...
+        let proxy = self
+            .find_member(name)
+            .ok_or_else(|| MeowError::Proxy("proxy not exist".into()))?;
+        self.force_set(Some(name));
+        if !proxy.alive_for_url(&self.test_url) {
+            let _ = crate::health::probe_and_record(
+                &proxy,
+                &self.test_url,
+                (!self.expected_status.is_empty()).then_some(self.expected_status.as_str()),
+                std::time::Duration::from_secs(5),
```
madeye, line 239: This inline probe means `PUT /proxies/{group}` can block up to 5 s on a dead member. Consider spawning the probe instead of awaiting it in the request handler — the selection is already recorded either way.

madeye:
Re-reviewed at `d612090` — CI is green and the two docs commits look good, but the two blocking items from the review are still open:

1. **Live per-connection/rate stats only work on the TUN path.** Only `meow_tunnel::tcp::handle_tcp` uses `copy_bidirectional_buf_tracked`; SOCKS5 (`socks5.rs:263`), HTTP proxy (`http_proxy.rs:207,326`), and TProxy (`tproxy/mod.rs:223`) still run the untracked relay and count bytes once at connection close. On mixed/SOCKS/HTTP inbounds — the most common setups — `/connections` byte counters stay `0` for the connection's lifetime and the `/traffic` rate stream reads `0` during a long transfer, then spikes at close. Please either convert the three listener call sites or document the limitation in `docs/mihomo-api-compatibility.md` (it currently marks `/traffic` as Compatible).

2. **ADR-0006 benchmark evidence for the relay hot-path change.** The tracked path adds a `DashMap::get_mut` + four atomic RMWs per 4 KiB chunk on the throughput-gated relay loop (~30k map lookups/sec per connection at 1 Gbps). Per the repo rules, relay-path changes need before/after numbers (median ≥ 0.98× baseline, W1–W5) in the commit body. If it regresses, resolving the connection entry once at setup and handing the closures two guard-owned `AtomicI64`s avoids the per-chunk lookup.

Everything else from the review is minor and can land as-is or in follow-ups.

madeye:
Blocker 2 is now resolved — I ran the ADR-0006 A/B myself and pushed 835070d with the fix and full evidence in the commit body.

**Measurement (meow-bench `--only throughput`, this PR vs current main `335aa55`, same machine, release profile).** With the per-chunk `DashMap::get_mut` in place, bulk throughput showed a consistent regression: 1 MB×10 at **0.87–0.90×** main and 64 MB×1 at **0.92–0.93×** (9 interleaved runs/side), while the latency-bound 4 KB×10k workload was unaffected (1.00×). So the gate concern was real, not theoretical.

**Fix (835070d).** `ConnectionGuard` now owns an `Arc<ConnCounters>` (two `AtomicI64`s) shared with the statistics table; the relay closures bump the atomics directly. Per-chunk cost drops from hash + shard-write-lock + 4 RMWs to **3 relaxed RMWs, zero map traffic** — the map is touched once at connection setup. After the fix, order-balanced medians over 30 runs (15/side, split across both A/B orderings because ordering alone swings bulk results ~10% on this machine via thermal bias — worth knowing for anyone re-running this):

| workload | main | PR+fix | ratio |
|---|---|---|---|
| 4 KB × 10000 | 2.29 Gbps | 2.29 Gbps | 1.00 |
| 1 MB × 10 | 17.8 Gbps | 16.6 Gbps | ~0.95 (sign flips with run order; within noise) |
| 64 MB × 1 | 21.3 Gbps | 20.9 Gbps | ~0.99 |

Overall median ≈ **0.99, inside the ≥ 0.98 gate**. The residual ~3 relaxed RMWs/chunk is the irreducible cost of live `/traffic` + `/connections` accounting (mihomo pays the same).

Housekeeping folded into the same commit: `ConnectionInfo` shrinks 136 B → 128 B (ADR-0011 delta in the commit body; the two inline `i64`s become one `Arc` thin-ptr, JSON shape unchanged via flattened serialization), the setup-path `Arc::new` keeps `track_connection_alloc_count` under its ≤ 4 allocs/conn gate, the HTTP proxy's pipelined `leftover` bytes now count toward the per-connection counter, and the closure rewrite also subsumes the clippy fix from 64d51a2 (rebased on top of it). Full regression bar (three-way clippy, `--lib` tests, doc build, fmt) is green locally.

Remaining from the review: the Windows `tasklist` CSV parsing in `bench_memory.rs` still truncates at the thousands separator (`"12,345 K"` parses as 12 KB) — see the earlier comment; small fix, non-blocking for this PR's runtime code.

madeye (approved):
Both review blockers resolved and verified (tracked relay on all inbounds; ADR-0006 gate passes at ~0.99x order-balanced after the guard-owned counters fix in 835070d). Windows RSS parsing fixed in c5846e2. CI green.

PR #337, opened 2026-07-15
perf(dns): stop preallocating capacity-sized LRU tables at startup

madeye:
## Summary

Perf-regression run against `baseline-2026-04-18.json` flagged idle RSS +11.8%. A same-machine interleaved A/B (April-commit binary vs HEAD, 3 runs each) confirmed a real, consistent **+1.0 MB idle RSS** delta; `malloc_history` attribution pinned it to `lru::LruCache::new(cap)` eagerly preallocating the full `cap`-slot hash table:

- `DnsCache::new(4096)`: 16 forward + 16 reverse shard tables (~930 KiB virtual, ~420 KiB dirty) charged to **every** process at startup, DNS traffic or not. The reverse table's ×4 capacity multiplier + 16-way sharding (both added since April) turned what used to be one lazy `DashMap` into the dominant startup allocation.
- fake-ip `MemoryStore::new(pool_size)`: same pattern, two ~65k-slot tables for the default /16 range.

## Fix

Construct these LRUs `unbounded()` and enforce the caps manually on insert (evict-after-put; identical visible semantics). For fake-ip the manual path evicts the LRU **pair** from both directions — an improvement over the previous independent per-side eviction, which could desynchronize `by_host`/`by_ip`.

Measured idle RSS (config-bench.yaml): **10,464 KB → 10,016 KB**; `MALLOC_SMALL` dirty 1120 KB → 704 KB. Remaining delta vs April (9,360 KB) is accumulated feature growth, not cache prealloc.

## Baseline refresh

Second commit records `baseline-2026-07-15.json` (median of 3 runs, post-fix) because the April snapshot is unreproducible even by its own binary (it was taken at `--concurrency 32` vs today's default 64; the April binary itself now scores 703/s and 13.2–14.1 MB load RSS against its recorded 762/s and 12.3 MB). Rationale + laptop-variance caveats documented in `methodology.md`.

## Not regressions (investigated, environment/noise)

- **conn-rate −7.9%**: April binary scores the same today (703–705/s); Go mihomo hits the identical 703/s ceiling — client/OS-bound on loopback.
- **throughput / latency flags**: single-run laptop variance exceeds the 5% threshold (13 samples/side: medians within 5%).

## Gates

- ADR-0008 alloc-count reproducers: pass (before and after)
- ADR-0011: `CacheEntry`/`ReverseEntry`/per-slot layouts untouched
- New regression tests: counting-allocator construction guard (`lazy_construction_alloc_test`), reverse-cap eviction, fake-ip paired eviction
- fmt + 3-way clippy `-D warnings` + doc + 916 lib tests: green

### DIFF
--- a/crates/meow-dns/src/cache.rs
+++ b/crates/meow-dns/src/cache.rs
@@ -24,7 +24,6 @@ use smallvec::SmallVec;
 use smol_str::SmolStr;
 use std::borrow::Cow;
 use std::net::IpAddr;
-use std::num::NonZeroUsize;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
@@ -73,6 +72,15 @@ pub struct DnsCache {
     /// Bounded per-shard LRU — entries past capacity are evicted in
     /// least-recently-used order.
     reverse: [Mutex<LruCache<IpAddr, ReverseEntry>>; SHARDS],
+    /// Per-shard capacity caps. The shards are constructed with
+    /// `LruCache::unbounded()` and the cap is enforced manually on insert:
+    /// `LruCache::new(cap)` preallocates the full `cap`-slot hash table per
+    /// shard at construction (16 × 48 KiB ≈ 770 KiB idle RSS at the default
+    /// 4096-entry forward cap), charging every process for tables that only
+    /// fill under sustained DNS load. Lazy tables grow to the same bucket
+    /// count only once the entries actually exist.
+    fwd_shard_cap: usize,
+    rev_shard_cap: usize,
 }
 
 #[derive(Clone, Debug, Eq, PartialEq)]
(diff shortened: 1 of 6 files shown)


madeye (commented):
Self-review via Claude Code (approval blocked — own PR): DNS-cache eviction sites are sound (one put per lock hold, MRU can't self-evict), alloc regression test is well-isolated, baseline refresh rationale is solid.

Follow-up worth a small PR: `MemoryStore::put` ignores overwrite return values, so a re-put of an existing host (possible via the pre-existing check-then-allocate race in `Pool::lookup`) leaves a stale `by_ip` entry — the new paired eviction can then pop a live forward mapping when that stale entry becomes LRU. Fix: pop the old counterpart when `by_host.put`/`by_ip.put` return a previous value. Low severity (worst case fake-IP churn), not blocking.

Issue #284, opened 2026-07-03
Any plans to support Openwrt?


sorrymynameistoolong:

madeye:
Thanks for asking! Partial answer, depending on what "OpenWrt support" means for you:

**What works today**: release builds are fully static musl binaries, so the `aarch64-unknown-linux-musl` and `x86_64-unknown-linux-musl` artifacts from the [releases page](https://github.com/madeye/meow-rs/releases) should run as-is on OpenWrt devices of those architectures (no glibc or ipk needed — just copy the binary and run it with a config). The transparent-proxy gateway setup for Linux/nftables is documented in `docs/tproxy-gateway.md` and the nftables helper script (`scripts/tproxy-gateway-linux.sh`) should adapt to OpenWrt's fw4/nftables stack.

**What doesn't exist yet**:
- `mipsel-unknown-linux-musl` builds (classic MIPS routers) — planned in ADR-0007 with a size budget targeting 16 MiB-flash OpenWrt devices, but currently disabled in the release matrix because stable Rust stopped publishing `rust-std` for that target. It would need a nightly/`build-std` pipeline to re-enable.
- Official `.ipk` packaging, an OpenWrt feed, or a LuCI app.

Could you share what you're looking for specifically — which device/architecture, and whether a plain static binary is enough or you need opkg packaging / LuCI integration? That will help scope this into something actionable.

sorrymynameistoolong:
Thanks for the detailed explanation! I'm using a **Linksys WRT1900ACS** (which uses the **Marvell Armada 385** SoC, **arm_cortex-a9_vfpv3-d16** architecture).

To make it easier to manage and deploy on OpenWrt, I'm specifically looking for **official .ipk packages**, or even better, an **OpenWrt feed and a LuCI app** for graphical configuration. Having a native package would definitely be more convenient than manually managing static binaries.

PR #343, opened 2026-07-16
ci: expand release matrix to 17 targets; support 32-bit platforms without 64-bit atomics

xmh0511:


.github/workflows/release.yml
```diff
@@ -55,16 +55,82 @@ jobs:
...
-          # nightly/build-std path. Tracked separately from ADR-0007.
+          - target: aarch64-pc-windows-msvc
+            runner: windows-latest
+            minimal_budget_bytes: 0
+            minimal_soft_gate: true
+          - target: x86_64-pc-windows-gnu
+            runner: windows-latest
+            minimal_budget_bytes: 0
+            minimal_soft_gate: true
+            cargo_flags: --no-default-features --features full
```
madeye: **Blocking**: `aarch64-pc-windows-gnu` does not exist as a Rust target (check `rustc --print target-list` — only `aarch64-pc-windows-gnullvm` exists, and that one additionally needs an llvm-mingw toolchain the runner doesn't ship). The toolchain step will fail for this job. Please drop this entry, or do the gnullvm variant as separate follow-up work.

.github/workflows/release.yml
```diff
@@ -55,16 +55,82 @@ jobs:
...
             minimal_soft_gate: true
-          # mipsel-unknown-linux-musl: temporarily disabled — `rust-std` is no
-          # longer published for this target on stable. Re-enable when stable
-          # ships the component again, or when ADR-0007 §1 promotes it to a
-          # nightly/build-std path. Tracked separately from ADR-0007.
+          - target: aarch64-pc-windows-msvc
+            runner: windows-latest
+            minimal_budget_bytes: 0
+            minimal_soft_gate: true
+          - target: x86_64-pc-windows-gnu
```
madeye: The `boring` crate (boring-tls, on by default) doesn't support `*-pc-windows-gnu` targets — BoringSSL's cmake build under mingw is unsupported, so this job will very likely fail at build time. Since the msvc targets already cover these CPUs, consider dropping the windows-gnu variants, or at minimum adding `cargo_flags: --no-default-features --features full` like the armv7/MIPS entries.

.github/workflows/release.yml
```diff
@@ -55,16 +55,82 @@ jobs:
...
+            minimal_budget_bytes: 0
+            minimal_soft_gate: true
+          - target: i686-unknown-linux-gnu
+            runner: ubuntu-latest
+            minimal_budget_bytes: 0
+            minimal_soft_gate: true
+          - target: i686-unknown-linux-musl
+            runner: ubuntu-latest
+            minimal_budget_bytes: 0
+            minimal_soft_gate: true
```
madeye, line 85: **Blocking**: this hits the same boring-sys 32-bit-musl breakage documented on the armv7 entry above (bindgen emits time64 `time_t: i64` per musl 1.2 while boring passes i32) — it applies to *all* 32-bit musl targets, including i686. This entry needs `cargo_flags: --no-default-features --features full`.

.github/workflows/release.yml
```diff
@@ -55,16 +55,82 @@ jobs:
...
+          - target: x86_64-apple-darwin
+            runner: macos-latest
+            minimal_budget_bytes: 0
+            minimal_soft_gate: true
+          - target: aarch64-apple-darwin
+            runner: macos-latest
+            minimal_budget_bytes: 0
+            minimal_soft_gate: true
```
madeye, line 137: **Blocking for both apple-darwin entries**: the "Package archive" step runs `sha256sum`, which isn't in PATH on GitHub's macOS runners (Homebrew coreutils only installs g-prefixed binaries). With `set -euo pipefail` the step dies. The packaging step needs a macOS branch using `shasum -a 256` (or `openssl dgst -sha256`).

.github/workflows/release.yml
```diff
@@ -55,16 +55,82 @@ jobs:
...
+          # cannot cross-compile for MIPS).
+          - target: mipsel-unknown-linux-musl
+            runner: ubuntu-latest
+            minimal_budget_bytes: 0
+            minimal_soft_gate: true
+            build_std: true
+            toolchain: nightly
+            cargo_flags: --no-default-features --features full
+          - target: mips-unknown-linux-musl
+            runner: ubuntu-latest
```
madeye: Unpinned `nightly` makes release builds non-reproducible and liable to break on any given day's nightly. Please pin a dated toolchain, e.g. `nightly-2026-07-01` (same for the `mips-unknown-linux-musl` entry below).

.github/workflows/release.yml
```diff
@@ -55,16 +55,82 @@ jobs:
...
+        shell: bash
+        run: |
+          TOOLCHAIN="${{ matrix.toolchain || 'stable' }}"
+          rustup toolchain install "$TOOLCHAIN" --profile minimal --no-self-update
+          rustup default "$TOOLCHAIN"
+          if [ "${{ matrix.build_std }}" = "true" ]; then
+            rustup component add rust-src --toolchain "$TOOLCHAIN"
+          else
+            rustup target add "${{ matrix.target }}" --toolchain "$TOOLCHAIN"
+          fi
```
madeye, line 151: **Blocking**: in GitHub Actions expressions `&&`/`||` return operand values and the **empty string is falsy**, so `matrix.build_std == true && '' || matrix.target` always evaluates to `matrix.target`. Both MIPS jobs will therefore run `rustup target add mips(el)-unknown-linux-musl` on nightly and fail — tier 3 targets have no prebuilt std to add.

Fix by making the fallback the empty string:

```yaml
targets: ${{ !matrix.build_std && matrix.target || '' }}
```

(The `components:` line below is fine — its truthy branch is a non-empty string.)

.github/workflows/release.yml
```diff
@@ -83,10 +149,14 @@ jobs:
       - name: Build release binary
         shell: bash
         run: |
-          if [ "${{ runner.os }}" = "Linux" ]; then
+          BUILD_STD_FLAG=""
+          if [ "${{ matrix.build_std }}" = "true" ]; then
+            BUILD_STD_FLAG="-Zbuild-std=std,panic_abort"
+          fi
+          if [ "${{ runner.os }}" = "Linux" ] && [ "${{ matrix.build_std }}" != "true" ]; then
             cargo zigbuild --release --locked ${{ matrix.cargo_flags }} --target ${{ matrix.target }} --bin meow
```
madeye, line 183: **Blocking**: routing `build_std == true` to plain `cargo build` leaves the MIPS jobs with no cross toolchain — nothing on ubuntu-latest provides a `mips(el)-linux-musl` linker or a C compiler for the `cc`-built deps, so the job fails at link time even once the toolchain step is fixed. `cargo zigbuild` passes `-Z` flags through, so the simpler fix is to keep the zigbuild path for Linux and just append `$BUILD_STD_FLAG`:

```bash
if [ "${{ runner.os }}" = "Linux" ]; then
  cargo zigbuild --release --locked ${{ matrix.cargo_flags }} $BUILD_STD_FLAG --target ${{ matrix.target }} --bin meow
else
  ...
```

(zig 0.13's MIPS support is the least-proven part of this — definitely verify with a `workflow_dispatch` dry run.)

madeye (changes requested):
Thanks for taking this on — broader release coverage is very welcome. One important caveat drives most of this review: **`release.yml` only runs on tag pushes and `workflow_dispatch`, so the green checks on this PR exercise none of the 15 new matrix jobs.** Several of them fail deterministically as written (see inline comments): a nonexistent target (`aarch64-pc-windows-gnu`), a GitHub-expressions falsy-empty-string bug that breaks both MIPS jobs at the toolchain step, MIPS routed to plain `cargo build` with no cross toolchain, `i686-unknown-linux-musl` hitting the documented boring-sys 32-bit-musl breakage, and macOS packaging calling `sha256sum` which isn't in PATH on macOS runners (use `shasum -a 256`).

Beyond the deterministic failures:

- **Unverified boring-tls cross-builds**: `x86_64-pc-windows-gnu` (the `boring` crate doesn't support windows-gnu), `aarch64-pc-windows-msvc`, `i686-pc-windows-msvc`, and both riscv64 targets keep default features, i.e. cross-compiling BoringSSL (cmake + NASM). The repo already uses `--no-default-features --features full` as the escape hatch (armv7, MIPS here) — that's the pragmatic default for the exotic targets until each is proven.
- **Scope vs. title**: the title says macOS + Windows ARM64, but the diff adds 15 targets across 6 arch families. Consider splitting into a "safe" PR (macOS, aarch64-msvc, glibc Linux) and an "experimental" one (MIPS, RISC-V, i686, windows-gnu) — the first can land after one dry run; the second needs real iteration.
- **Unpinned `nightly`** for the MIPS jobs makes release builds non-reproducible; pin a dated nightly.
- **ADR-0007 bookkeeping**: the deleted comment deferred mipsel to "when ADR-0007 §1 promotes it to a nightly/build-std path". This PR takes that path without touching the ADR — please note it in the commit body or amend the ADR.
- Minor: the new gnu targets rely on cargo-zigbuild's default glibc floor; pinning it explicitly in the target string (e.g. `aarch64-unknown-linux-gnu.2.17`) makes the compatibility contract visible.

**Before merging any version of this, please run the workflow via `workflow_dispatch` on this branch (it's designed as a dry run — no upload) and get all matrix jobs green.** That single step would have caught every issue above.

.github/workflows/release.yml
```diff
@@ -55,16 +61,84 @@ jobs:
...
+            minimal_budget_bytes: 0
+            minimal_soft_gate: true
+          - target: x86_64-pc-windows-gnu
+            runner: windows-latest
+            minimal_budget_bytes: 0
+            minimal_soft_gate: true
+            cargo_flags: --no-default-features --features full
+          - target: i686-pc-windows-msvc
+            runner: windows-latest
+            minimal_budget_bytes: 0
```
madeye, line 88: Run 2 shows `-mno-avx512f` doesn't fix this — the i686 failure is byte-identical to run 1 (`always_inline function '_mm512_castsi512_si256' requires target feature 'evex512'`). The root cause is that zig's clang (18+) gates 512-bit vectors behind the `evex512` feature, while blake3 compiles `blake3_avx512.c` with only `-mavx512f -mavx512vl`. The minimal fix is to *add* the missing feature rather than subtract avx512:

```yaml
cflags: -mevex512
```

(same for the i686-musl entry below). If that doesn't take either, the CFLAGS may not be reaching that compilation unit and a blake3-side knob/upgrade is the next lever.

.github/workflows/release.yml
```diff
@@ -55,16 +61,84 @@ jobs:
...
+            minimal_soft_gate: true
```
madeye, line 89: Once the blake3 issue is past, this target should still hit the boring-sys 32-bit-musl breakage documented on the armv7-musleabihf entry (time64 `time_t` mismatch — it applies to all 32-bit musl targets). Run 2 supports the musl-specificity: `armv7-unknown-linux-gnueabihf` built boring-tls successfully via zig. This entry needs:

```yaml
cargo_flags: --no-default-features --features full
```

.github/workflows/release.yml
```diff
@@ -83,10 +157,14 @@ jobs:
       - name: Build release binary
         shell: bash
         run: |
-          if [ "${{ runner.os }}" = "Linux" ]; then
+          BUILD_STD_FLAG=""
+          if [ "${{ matrix.build_std }}" = "true" ]; then
```
madeye: Both dry runs confirm the MIPS jobs die here: `ring`'s build script fails with `ToolNotFound: failed to find tool "mips(el)-linux-musl-gcc"` — plain `cargo build` on ubuntu-latest has no MIPS cross C compiler or linker. `cargo zigbuild` forwards `-Z` flags, so the simplest fix is to stop special-casing build-std out of the zigbuild path:

```bash
if [ "${{ runner.os }}" = "Linux" ]; then
  cargo zigbuild --release --locked ${{ matrix.cargo_flags }} $BUILD_STD_FLAG --target ${{ matrix.target }} --bin meow
else
  ...
```

(zig cc supports mips/mipsel-linux-musl; if zig's MIPS support turns out flaky, the fallback is installing a `mips(el)-linux-musl` cross toolchain and wiring `CC_*`/`CARGO_TARGET_*_LINKER`.)

.github/workflows/release.yml
```diff
@@ -55,16 +61,84 @@ jobs:
...
+            runner: macos-latest
+            minimal_budget_bytes: 0
+            minimal_soft_gate: true
+          - target: aarch64-apple-darwin
+            runner: macos-latest
+            minimal_budget_bytes: 0
+            minimal_soft_gate: true
+          # MIPS: tier 3, no prebuilt rust-std on stable — build std from source
+          # via nightly + -Zbuild-std. boring-tls is disabled (boring-sys
+          # cannot cross-compile for MIPS).
```
madeye: Still worth pinning a dated nightly (e.g. `nightly-2026-07-01`) here and on the mips entry below — an unpinned nightly makes release builds non-reproducible and can break on any given day.

crates/meow-api/src/routes.rs
```diff
@@ -2032,7 +2032,14 @@ async fn read_os_memory_limit_linux() -> u64 {
             rlim_max: 0,
         };
         if libc::getrlimit(libc::RLIMIT_AS, &mut rl) == 0 && rl.rlim_cur != libc::RLIM_INFINITY {
-            return rl.rlim_cur;
+            #[cfg(target_pointer_width = "32")]
```
madeye, line 2065: Verified working — this turned `armv7-unknown-linux-gnueabihf` green in run 2, and it's correct for all four cases (glibc/musl × 32/64-bit; on 32-bit musl `rlim_t` is already `u64` so the cast branch is a no-op there). Optional simplification per the repo's lint policy (CLAUDE.md sanctions allow-with-reason), replacing both cfg blocks:

```rust
#[allow(clippy::unnecessary_cast, reason = "rlim_t is u32 on 32-bit glibc")]
return rl.rlim_cur as u64;
```

Fine as-is too.

madeye (commented):
Follow-up with hard evidence: I ran the release workflow twice via `workflow_dispatch` (dry run, no upload) on this PR's head:

- **Run 1** (`8b12cf2`): https://github.com/madeye/meow-rs/actions/runs/29482867392 — 12/17 green
- **Run 2** (`a3ef0dda`, current head): https://github.com/madeye/meow-rs/actions/runs/29483543038 — **13/17 green**

**Confirmed working** (nice!): both apple-darwin, all three Windows MSVC targets (boring-tls cross-compiles fine, including aarch64 and i686), `x86_64-pc-windows-gnu` (boring disabled), all x86_64/aarch64 Linux gnu+musl, both armv7 targets, and `riscv64gc-unknown-linux-musl`. The `rlim_cur` fix is verified — `armv7-unknown-linux-gnueabihf` went red → green between the runs.

**One retraction from my first review**: the macOS `sha256sum` concern was wrong — both apple-darwin jobs passed packaging, the runners do have it in PATH. Sorry for the noise. (The fmt concern I raised while re-checking locally is also moot — CI's rustfmt 1.97 accepts the cfg blocks.)

**Still failing at the current head (4 jobs)** — details inline:
1. Both i686 Linux targets: blake3's `blake3_avx512.c` fails under zig cc with the clang `evex512` error, and run 2 shows `-mno-avx512f` had no effect (byte-identical error). Suggest `-mevex512` instead.
2. Both MIPS targets: `ring` can't find `mips(el)-linux-musl-gcc` — the build-std path bypasses zigbuild, so there's no cross C compiler/linker. Route it through `cargo zigbuild` (it forwards `-Zbuild-std`).

Remaining from the first review: pin the nightly for MIPS, `--no-default-features --features full` for `i686-unknown-linux-musl` (the boring 32-bit-musl time64 breakage is still queued behind the blake3 failure — armv7-*gnueabihf* building boring fine confirms it's musl-specific), and the ADR-0007 note for the MIPS nightly/build-std path.

Happy to re-dispatch the dry run after the next push — with the two fixes above this looks very close to fully green.

.github/workflows/release.yml
```diff
@@ -55,16 +60,84 @@ jobs:
...
+          - target: mipsel-unknown-linux-musl
```
madeye: Both MIPS jobs now fail on source, not toolchain: `E0432: unresolved import std::sync::atomic::AtomicU64` in `meow-tunnel` — mips32 lacks 64-bit atomics, so std doesn't provide `AtomicU64`/`AtomicI64` there. 64-bit atomics are used in 8+ crates, so this needs a workspace-level decision (`portable-atomic`) before MIPS can ship. Suggest dropping both MIPS entries from this PR and tracking them in a follow-up issue — that also lets the nightly/build-std/cross/wrapper machinery in this workflow shrink dramatically.

.github/workflows/release.yml
```diff
@@ -80,11 +153,52 @@ jobs:
       - run: cargo install --locked cargo-zigbuild
         if: runner.os == 'Linux'
 
+      - name: Install cross-compilation tools
+        if: runner.os == 'Linux'
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cross
```
madeye: Pin the version: `tool: cross@0.2.5` (or whatever was tested). An unpinned tool install in a release workflow means a cross release can silently change what you ship — same reasoning as the dated-nightly pin.

.github/workflows/release.yml
```diff
@@ -80,11 +153,52 @@ jobs:
...
+
+      - name: Install 32-bit dev headers (i686)
+        if: runner.os == 'Linux'
+        run: sudo apt-get update -qq && sudo apt-get install -y --no-install-recommends gcc-multilib libc6-dev-i386
+
+      - name: Create i686 CC and linker wrappers
+        if: runner.os == 'Linux'
+        run: |
+          cat > /tmp/cc-i686.sh <<'SCRIPT'
```
madeye: This rewrite (`--target=*` → `i686-linux-gnu`) is what breaks the musl target now: aws-lc's C is compiled against host glibc ≥2.38 headers, whose C23 mode references `__isoc23_sscanf`/`__isoc23_strtol` — symbols static musl doesn't have, hence the i686-musl link failure. The gnu target is fine. For musl this wrapper would need a real i686-musl toolchain (musl.cc cross toolchain, or zig ≥0.14 — 0.13's LLVM-18 feature translation caused the original blake3 evex512 errors); simplest is to drop `i686-unknown-linux-musl` and keep the proven-green gnu variant.

.github/workflows/release.yml
```diff
@@ -80,11 +153,52 @@ jobs:
...
+            case "$arg" in
+              --target=*) args+=("--target=i686-linux-gnu") ;;
+              *)          args+=("$arg") ;;
+            esac
+          done
+          exec clang "${args[@]}"
+          SCRIPT
+          chmod +x /tmp/cc-i686.sh
+
+          cat > /tmp/ld-i686.sh <<'SCRIPT'
```
madeye: The musl detection here works — the final link is genuinely static musl (rustc's self-contained crt + `-nodefaultlibs`), which is exactly why the glibc-contaminated C objects now fail loudly instead of shipping a mislabeled binary. Good instinct; the remaining gap is on the CC side (see comment above).

madeye (commented):
Consolidated follow-up after dry-running the workflow on each pushed head (13 `workflow_dispatch` runs total today). Latest: **run on `b0bf429` = 14/17 green** — huge progress from 12/17 this morning. https://github.com/madeye/meow-rs/actions/runs/29488997228

**Now proven green** (in addition to the original 12): `i686-unknown-linux-gnu` — the clang-wrapper + gcc-multilib + `--no-default-features --features full` combination is coherent and works. The cross-rs + `-Zbuild-std` toolchain path for MIPS also works *as a toolchain* — both MIPS jobs now compile std and every C dependency successfully.

**The two remaining failures are no longer CI problems:**

### 1. MIPS: `meow-rs` source uses 64-bit atomics that MIPS32 doesn't have

Both MIPS jobs die on `error[E0432]: unresolved import std::sync::atomic::AtomicU64` (and `AtomicI64`) in `meow-tunnel` — mips32 has no native 64-bit atomics (`target_has_atomic = "64"` is false), so std simply doesn't provide these types there. armv7 and i686 pass because they have LDREXD/cmpxchg8b.

This is not fixable in workflow YAML. 64-bit atomics are used across at least 8 production crates (`meow-tunnel/src/statistics.rs`, `udp.rs`, `meow-listener/src/socks5_udp.rs`, `meow-api/src/routes.rs`, `meow-config/src/{proxy_provider,rule_provider}.rs`, `meow-anytls/...`). The standard fix is the [`portable-atomic`](https://crates.io/crates/portable-atomic) crate (drop-in `AtomicU64` that is native where available and lock-based on mips32), but that's a workspace-wide dependency decision for the maintainer, not a release-workflow tweak.

**Recommendation: split the MIPS targets out of this PR into a follow-up issue.** They need a source-level porting decision first.

### 2. `i686-unknown-linux-musl`: C deps compiled against glibc headers, linked against musl

Link fails with `undefined reference to __isoc23_sscanf / __isoc23_strtol` in aws-lc-sys. Root cause: `/tmp/cc-i686.sh` rewrites every `--target=` to `i686-linux-gnu`, so all C objects are compiled against the host's glibc ≥2.38 headers (whose C23 mode redirects `sscanf`/`strtol` to `__isoc23_*` symbols) — while the final link is genuine static musl (good — the new `ld-i686.sh` musl detection works), which has no such symbols. Ironically the *linker* side is now honest and the *compiler* side isn't.

A real i686-musl C toolchain is needed (musl.cc's `i686-linux-musl-cross`, or zig with a newer-than-0.13 release — 0.13's LLVM 18 x86 feature translation is what caused the original blake3 `evex512` failures). **Pragmatic recommendation: drop `i686-unknown-linux-musl` from this PR too and keep `i686-unknown-linux-gnu`**, which is proven green.

### If the two problem targets are split out, this PR is 14/14 green and mergeable after cleanup:

- Pin the MIPS `toolchain: nightly` to a dated nightly if MIPS stays; if MIPS is split out, the nightly/build-std/cross machinery and the wrapper-script step can all be deleted from this PR.
- Pin cross: `tool: cross@<version>` in the taiki-e/install-action step (unpinned tool installs are a supply-chain and reproducibility liability in a release workflow).
- The deleted mipsel comment referenced ADR-0007 §1; whichever way MIPS goes, the ADR should be updated to match.
- Worth adding to the packaging step (cheap insurance this saga demonstrated the value of): `file target/.../meow` + fail if a `*-musl` artifact has an `INTERP`/glibc dependency (`readelf -d`). A green job stopped meaning a correct artifact several iterations ago.

Say the word if you want me to re-run the dry run on the next push.

Cargo.toml
```diff
@@ -110,6 +110,9 @@ maxminddb = { version = "0.27", features = ["mmap"] }
 # Concurrency
 dashmap = "6"
 
+# Portable atomics for 32-bit targets without native AtomicU64 (MIPS)
+portable-atomic = { version = "1", default-features = false }
```
madeye: This `[workspace.dependencies]` entry is currently unused — every member crate declares its own `portable-atomic = "1"` directly instead of `portable-atomic = { workspace = true }`. Also note the member declarations use default features while this entry says `default-features = false`; on mips32 the (default) `fallback` feature is what provides `AtomicU64` at all, so if the crates ever switch to `workspace = true` as-is, MIPS breaks again. Suggest: keep defaults on here, and point all member crates at the workspace entry.

crates/meow-anytls/src/session/session.rs
```diff
@@ -13,7 +13,8 @@ use tokio::sync::{Notify, RwLock, mpsc};
 use tokio::time::{self, Duration, Instant, MissedTickBehavior};
 use tracing::{field, info_span};
 
-static SESSION_COUNTER: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(1);
+static SESSION_COUNTER: AtomicU64 = AtomicU64::new(1);
+use portable_atomic::AtomicU64;
```
madeye: `use` placed after the `static` it feeds (and mid-file). Works, but it'll surprise readers — move it up into the import block at the top of the file.

.github/workflows/release.yml
```diff
@@ -80,11 +151,39 @@ jobs:
...
+          tool: cross
+
+      - name: Install 32-bit dev headers (i686)
+        if: runner.os == 'Linux'
+        run: sudo apt-get update -qq && sudo apt-get install -y --no-install-recommends gcc-multilib libc6-dev-i386
+
+      - name: Create i686 CC and linker wrappers
+        if: runner.os == 'Linux'
+        run: |
+          cat > /tmp/cc-i686.sh <<'SCRIPT'
```
madeye: This wrapper is now the last incoherent piece of the i686 toolchain: it compiles C against the host's glibc 2.39 headers (`--target=i686-linux-gnu` with system clang), while cargo-zigbuild links against zig's glibc-2.17 stubs — glibc ≥2.38 headers emit `__isoc23_*` symbol references that 2.17 can't satisfy, which is the `undefined symbol: __isoc23_sscanf` failing both i686 jobs in this run. Compiler headers and link-time libc have to come from the same toolchain at the same version. Given the run history (blake3 evex512 under zig 0.13 was the original blocker), the realistic options are: bump setup-zig to ≥0.14 and let zigbuild own i686 end-to-end with `-mevex512`/no wrapper, or drop the i686 targets from this PR like MIPS.

madeye (commented):
Dry run on `6989fbe`: https://github.com/madeye/meow-rs/actions/runs/29492997546 — **7/17**, a regression from 14/17, but all four causes are mechanical:

1. **Stale `Cargo.lock` (6 jobs: all Windows + macOS).** The portable-atomic dependency edges were added to the Cargo.tomls but the regenerated lockfile wasn't committed, so every plain-`cargo build --locked` job fails with `cannot update the lock file`. Run `cargo update -p portable-atomic --precise 1.13.1` (or just `cargo check`) locally and commit `Cargo.lock`.
2. **Why Linux didn't catch it (worth knowing):** cargo-zigbuild and cross both run `cargo metadata` *without* `--locked` before building, which silently rewrites the lockfile — so `--locked` is effectively unenforced on the Linux path. The macOS/Windows jobs are currently the only real lockfile gate. A cheap guard: an early `cargo metadata --locked > /dev/null` step before any tool touches the lock.
3. **The atomics sweep missed two spots (both MIPS jobs).** `meow-tunnel/src/statistics.rs:25` still imports `std::sync::atomic::AtomicI64` (the remaining `E0432`), and `meow-api/src/routes.rs:1695` uses `AtomicU64`/`AtomicI64` as `prometheus_client` metric backings — that one is harder, since portable-atomic types don't implement prometheus-client's `Atomic` trait; it needs a cfg'd metric width (or is another argument for splitting MIPS out of this PR).
4. **i686 regressed from green to red on *both* variants** — removing `ld-i686.sh` handed linking back to cargo-zigbuild, which links against zig's glibc-2.17 stubs, while `/tmp/cc-i686.sh` still compiles C against the host's glibc 2.39 headers. glibc ≥2.38's C23 mode emits `__isoc23_sscanf`/`__isoc23_strtol`, which 2.17 stubs don't export → `undefined symbol`. (The previous green i686-gnu worked only because the clang linker wrapper linked the host's real glibc 2.39 — which would also have meant the released binary requires glibc ≥2.38 at runtime, poor portability for a release artifact.) The i686 C toolchain needs to be internally coherent: headers, target, and link libc all from one toolchain. Honestly, at this point I'd drop both i686 targets from this PR alongside MIPS and ship the 12 solid ones.

Smaller wiring nits inline.

crates/meow-api/src/routes.rs
```diff
@@ -1687,12 +1687,13 @@ async fn put_configs(
 // upstream: N/A — meow-rs enhancement; Go mihomo has no native /metrics endpoint.
 
 async fn get_metrics(State(state): State<Arc<AppState>>) -> Response {
+    use portable_atomic::{AtomicI64, AtomicU64};
```
madeye: This is the all-target compile failure: prometheus-client 0.22.3 only implements `Atomic<u64>`/`Atomic<i64>` for `std::sync::atomic::AtomicU64`/`AtomicI64` — `portable_atomic` types don't satisfy the `Registry::register`/`inc_by`/`set` bounds. `cargo check -p meow-api` fails with `E0277`/`E0599` on x86_64 too, not just MIPS. Keep std atomics here and cfg-gate the metric width (or the whole `/metrics` endpoint) on `target_has_atomic = "64"` instead — see the review body for a sketch.

madeye (commented):
Three things on `f77493d` — no dry run dispatched for this head because item 2 is a deterministic all-target compile failure (verified locally), so a run would just be 17 red jobs.

### 1. ⚠️ ~40,000 lines of unrelated files were committed

`f77493d` includes what look like local AI-assistant working files swept up by a broad `git add`:
- `.agents/skills/i-impeccable/**` (~35k lines of design-skill tooling)
- `.mimocode/**` (incl. `.cron-lock` and two plan files)
- `crates/meow-app/PLAN.md`
- `review.md`

Please remove these and force-push (`git rm -r .agents .mimocode crates/meow-app/PLAN.md review.md` + amend/rebase) — and double-check the plan/review files didn't capture anything you'd rather not have in a public repo's history. Adding `.agents/`/`.mimocode/` to your global gitignore avoids a repeat.

### 2. `get_metrics` now fails to compile on **all** targets

prometheus-client 0.22.3 implements its `Atomic` traits only for `std::sync::atomic` types (`counter.rs:118`, gauge equivalent) — there's no portable-atomic support. So:

```
error[E0277]: the trait bound `portable_atomic::AtomicU64: prometheus_client::metrics::counter::Atomic<u64>` is not satisfied
```

(verified with `cargo check -p meow-api` on this head — your own PR CI will hit the same). The metrics code needs to keep std atomics and gate the width instead, e.g.:

```rust
#[cfg(target_has_atomic = "64")]
use std::sync::atomic::{AtomicI64, AtomicU64};
#[cfg(not(target_has_atomic = "64"))]
use std::sync::atomic::{AtomicI32 as AtomicI64Sub, AtomicU32 as AtomicU64Sub};
```
…with `Counter<u32, AtomicU32>`/`Gauge<i32, AtomicI32>` on the non-64 side (32-bit counters wrap at 4 GiB of traffic; Prometheus `rate()` tolerates resets, and it's mips32-only). Or, simpler: leave `get_metrics` on std atomics unconditionally and `#[cfg]` the whole `/metrics` endpoint off on `not(target_has_atomic = "64")` until MIPS is a real target.

The `statistics.rs` and test fixes in this commit are correct — with the metrics issue resolved those should clear the last MIPS `E0432`s.

### 3. Run on `fe835fa` (previous head): 13/17 ✅

https://github.com/madeye/meow-rs/actions/runs/29499232482 — the `Cargo.lock` fix restored all six Windows/macOS jobs, matching prediction. Remaining reds were the two MIPS (now addressed modulo item 2) and two i686 jobs. The `-std=gnu17` fix for i686 in this head looks right (the runner's clang defaults to C23, which is where `__isoc23_*` came from) — I'll dispatch the next dry run as soon as item 2 is fixed, and if i686 goes green the remaining question is only whether the i686-musl artifact should ship with glibc-header-compiled C objects (the `file`/`readelf` check from the earlier review).

.github/workflows/release.yml
```diff
@@ -55,16 +55,57 @@ jobs:
...
+            minimal_soft_gate: true
+          - target: x86_64-apple-darwin
+            runner: macos-latest
+            minimal_budget_bytes: 0
+            minimal_soft_gate: true
+          - target: aarch64-apple-darwin
+            runner: macos-latest
+            minimal_budget_bytes: 0
+            minimal_soft_gate: true
+          - target: riscv64gc-unknown-linux-musl
```
madeye: Non-blocking: the old matrix carried a comment explaining mipsel's absence; after this PR there's no trace of why MIPS and i686 aren't here. Suggest restoring a short note, e.g.:

```yaml
# mips(el)-unknown-linux-musl: blocked on 64-bit atomics in source (mips32 has
# no AtomicU64) — see #<follow-up issue>. i686: needs a coherent 32-bit C
# toolchain (glibc C23 symbol drift / zig 0.13 evex512); same issue.
```

madeye (approved):
**Dry run on the final head `87aea93`: 13/13 build jobs green, publish correctly skipped.** https://github.com/madeye/meow-rs/actions/runs/29501625211

That clears everything blocking from my original review, so this supersedes the earlier request-changes. What this PR now delivers, all dry-run-validated:

- **9 new release targets**: x86_64 + aarch64 macOS, aarch64 + i686 Windows MSVC, x86_64 windows-gnu (boring-tls off), aarch64/armv7/x86_64 Linux glibc, riscv64gc musl
- The `rlim_cur` 32-bit fix in meow-api (verified: it's what turned armv7-gnueabihf green)
- Artifact spot-check: the riscv64 binary is a genuine `static-pie` RISC-V ELF (double-float, stripped) — exactly what the target promises

Two non-blocking follow-ups:

1. **File an issue capturing today's MIPS/i686 findings** so the 16 dry-run iterations aren't lost. Key facts for it: mips32 needs a workspace answer for `AtomicU64`/`AtomicI64` (8+ crates; portable-atomic works for plain uses but **not** for `prometheus_client` metric backings, which only accept std atomics — that needs a cfg'd width or gating `/metrics`); the working MIPS *toolchain* recipe is cross-rs + nightly + `-Zbuild-std=std,panic_abort`; i686 needs an internally coherent C toolchain (the host-clang hybrid died on C23 `__isoc23_*` symbol drift — `-std=gnu17` was the right patch — and zig 0.13 has the LLVM-18 evex512 bug on x86; zig ≥0.14 is the likely clean path).
2. Inline: a one-line comment in the matrix noting why MIPS/i686 are absent, so the next person doesn't re-walk today's path.

Also worth knowing (informational, pre-existing): `--locked` is only genuinely enforced on the macOS/Windows jobs — cargo-zigbuild and cross run `cargo metadata` unlocked first, which silently rewrites the lockfile on the Linux runners.

Nice persistence getting this to green — the final shape (13 solid targets, zero exotic toolchain machinery) is the right one.

madeye:
Dry run on `4ecfa2e` (i686 targets re-added via cross): **15/15 green** — https://github.com/madeye/meow-rs/actions/runs/29548406879

Artifact verification (the part a green check can't prove):
```
i686-unknown-linux-gnu/meow:  ELF 32-bit LSB pie executable, Intel 80386, dynamically linked,
                              interpreter /lib/ld-linux.so.2, for GNU/Linux 2.6.32, stripped
i686-unknown-linux-musl/meow: ELF 32-bit LSB executable, Intel 80386, statically linked, stripped
```
Both are exactly what their target names promise — the musl one is truly static, and the gnu one links against cross's old-glibc baseline (runs on anything ≥ 2.6.32-era glibc), which is what a release binary should look like. The Docker/cross route was the right call for i686: one coherent gcc toolchain, no clang/zig header-stub mismatches, no evex512.

My approval stands and now covers the full 15-target matrix. One last nit repeated from earlier: pin the cross version in the taiki-e step (`tool: cross@0.2.5`) so a future cross release can't silently change these images.

madeye (changes requested):
⛔ **The merge commit `87d1a5d` clobbered the validated workflow — this supersedes my approval, which applied to `4ecfa2e`.**

What the merged `release.yml` lost (it's down to 126 lines from ~270):

- **The matrix now contains ONLY the two MIPS targets.** All 15 validated targets are gone — x86_64/aarch64/armv7 Linux (musl+gnu), all four Windows, both macOS, both i686, riscv64. As-is, a tag would release two MIPS binaries and nothing else.
- **The ADR-0007 size gates are gone** — the minimal-binary build + hard 8 MiB check for `aarch64-unknown-linux-musl`.
- **OpenWrt `.ipk` packaging is gone** (`openwrt_archs`, the build-ipk steps, and check whether the luci ipk step in the publish job survived).
- **The Windows zip packaging branch is gone** — the packaging step now unconditionally copies `meow` (not `meow.exe`) and tars.
- **The setup-zig / cargo-zigbuild install steps are gone** while the build script still calls `cargo zigbuild` in its Linux branch — any non-cross Linux target added back will fail immediately.

Separately: **`Cargo.lock` got a wholesale dependency upgrade** (~15 crates bumped, including `aws-lc-sys` 0.41→0.42). Unrelated dep upgrades don't belong in this PR — they move the ADR-0006/0007 perf/size baselines invisibly. Please restore the lock to `4ecfa2e`'s state plus only the edges your source changes need.

It looks like the side branch's old `release.yml` won the merge wholesale. The fix is mechanical: `git checkout 4ecfa2e -- .github/workflows/release.yml Cargo.lock`, then re-apply *only* the two MIPS matrix entries + the `$BUILD_STD` line in the cross branch, and re-resolve the lock minimally.

**On the substance (the atomic polyfill): the approach looks right.** `meow-common/src/atomic.rs` with `cfg(target_has_atomic = "64")` type aliases is the correct gate (and thank you for switching off `target_pointer_width` — i686 Windows has 64-bit atomics and would have been wrongly narrowed). Three asks:

1. `statistics.rs` is on the ADR-0011 watch list (`ConnectionInfo` 120 B target) — the commit body needs before/after `-Zprint-type-sizes` byte counts per the repo's invariants. On 64-bit targets the aliases resolve identically so the answer should be "no change", but it needs to be stated.
2. On mips32 the traffic counters are now u32 — they wrap at 4 GiB. Fine as a v1 trade-off, but document it (a line in the matrix comment and in the polyfill module doc).
3. The hot-path files touched (`tcp.rs`, `http_proxy.rs`, `socks5*.rs`, `tproxy`) should be cast-only changes at counter boundaries — if any of them changed allocation or added branches beyond `as Uint/Int` casts, ADR-0008 wants a note.

I've dispatched a dry run of this head as-is — it only has the 2 MIPS jobs, which conveniently answers whether polyfill + cross + build-std actually produces MIPS binaries before you restructure the merge.

madeye:
Heads-up on `cefb3a2`: the restored `release.yml` is **UTF-16 LE with a BOM and CRLF line endings** (git shows it as binary — first bytes `FF FE`, every character NUL-padded). GitHub Actions only parses UTF-8 workflow files, so on this head the Release workflow won't load at all — it can't even be dispatched. It looks like the file was copied through a Windows-encoded editor/clipboard.

The cleanest repair for both this and the dependency churn from the merge:

```bash
git checkout 4ecfa2e -- .github/workflows/release.yml Cargo.lock
# re-apply, in a UTF-8 editor: the two MIPS matrix entries (+ $BUILD_STD in the cross branch)
cargo check   # re-resolves only the lock edges your source changes actually need
```

That lands you on: the 15 validated targets + MIPS entries + your atomic polyfill, with none of the merge fallout. The dry run of `87d1a5d`'s two MIPS jobs is running now and will tell us whether polyfill+cross works before you push the repaired version.

madeye:
Good news to pair with the repair work: the dry run of `87d1a5d`'s two MIPS jobs is **green** — https://github.com/madeye/meow-rs/actions/runs/29556772023 — so the atomic polyfill + cross\@0.2.5 + `-Zbuild-std` combination genuinely works. That was the last open technical question.

Artifact inspection of `mipsel-unknown-linux-musl/meow`:
```
ELF 32-bit LSB pie executable, MIPS, MIPS32 rel2, dynamically linked,
interpreter /lib/ld-musl-mipsel-sf.so.1, stripped
```
Two observations:
- **Soft-float ✓** (`-sf` interpreter) — matches the OpenWrt mips_24kc-class targets this is presumably for.
- **It's dynamically linked against musl**, unlike the static x86_64/aarch64/i686 musl artifacts — Rust's mips-musl targets don't default to `crt-static`. On OpenWrt this is fine (musl is the system libc). If you want it static like the others, add `RUSTFLAGS="-C target-feature=+crt-static"` for the MIPS entries; either way, note the choice in the matrix comment.

So the remaining work is exactly the repair items from the earlier review: restore `release.yml` (UTF-8! — `823a3f0` is still UTF-16/CRLF and the workflow can't parse) and `Cargo.lock` from `4ecfa2e`, re-apply the two MIPS entries, plus the ADR-0011 byte-count note for `statistics.rs` and the u32-wrap doc line. Then this is a 17-target release, fully validated.

madeye:
`a9502e7`: encoding fixed ✓ (the workflow parses again — dry run dispatched: https://github.com/madeye/meow-rs/actions/runs/29557101331). Two leftovers before this is mergeable:

1. **The 15 non-MIPS targets are commented out** in the restored matrix — looks like local debug state for iterating on MIPS. They need uncommenting (and the deleted inline comments — the armv7 boring-sys explanation, the aarch64 ADR-0007 gate note — restored with them). Note the packaging/size-gate/OpenWrt steps this file lost in the merge are still absent too; comparing against `4ecfa2e`'s `release.yml` (270 lines vs this one's ~210) will show what's missing.
2. **`Cargo.lock` still carries the wholesale dependency upgrade** from the merge (~15 unrelated bumps incl. aws-lc-sys 0.41→0.42). Please restore it from `4ecfa2e` and re-resolve minimally.

The dispatched run covers the two active (MIPS) jobs on the post-clippy-fix source, which completes validation of the polyfill code as it now stands.

madeye:
`a298dd1` restores nearly everything — setup-zig, the ADR-0007 minimal-binary size gates, OpenWrt/luci ipk packaging, and the Windows zip branch are all back, and the workflow parses. Dry run on the 16 active targets: https://github.com/madeye/meow-rs/actions/runs/29557333968

Two last items:
1. **`x86_64-unknown-linux-musl` is still commented out** (line 24) — that's the flagship Linux target; presumably the last leftover of the debug commenting.
2. **`Cargo.lock` still has the ~15 unrelated dependency bumps** from the merge (aws-lc-sys 0.41→0.42 etc.). Restore from `4ecfa2e` + minimal re-resolve, per the earlier review.

Once those two are in, the next dry run should be the full 17/17 and I'd expect this to be mergeable.

madeye:
Dry run on `7f76488`: **15/17** — the fully restored matrix is green everywhere except MIPS: https://github.com/madeye/meow-rs/actions/runs/29557491791

The MIPS failure is a one-commit regression from `a298dd1`: silencing clippy's `useless_conversion` in `connections_json` replaced the working `.into()` with a plain assignment —

```rust
let up_i64: i64 = up;      // E0308 on mips32: up is Int = i32 there
```

The conversion was only "useless" on 64-bit hosts; on mips32 it's load-bearing. The fix is the pattern already used three lines down in `traffic_json` (keep `.into()`, keep the allow):

```rust
#[allow(clippy::useless_conversion)]
let up_i64: i64 = up.into();
#[allow(clippy::useless_conversion)]
let down_i64: i64 = down.into();
```

(routes.rs:481 and :483.)

With that plus the `Cargo.lock` restore (still showing the ~15 unrelated bumps), the next run should be 17/17. Nits to fold in while you're there: the explanatory matrix comments lost in the restore (armv7 boring-sys time64 note, aarch64 ADR-0007 gate note, the #284 OpenWrt-archs rationale) and the ADR-0011 byte-count statement for the `statistics.rs` change in the commit body.

madeye (approved):
**Dry run on `04c9d36`: 17/17 green.** https://github.com/madeye/meow-rs/actions/runs/29558702693 — this supersedes my request-changes; every item from it is resolved:

- ✅ Full 17-target matrix restored (UTF-8, active, no commented leftovers)
- ✅ ADR-0007 size gates, OpenWrt/luci ipk packaging, Windows zip, setup-zig — all back and exercised in the run
- ✅ `Cargo.lock` restored to the pre-merge state + minimal re-resolve (single-edge delta)
- ✅ The mips32 `E0308` fixed with width-safe casts
- ✅ MIPS artifact re-verified: soft-float MIPS32r2 musl ELF (`ld-musl-mipsel-sf.so.1`)

What this PR delivers, all dry-run-validated across 23 workflow_dispatch runs: **11 new release targets** (macOS ×2, Windows aarch64/i686-msvc/x86_64-gnu, Linux glibc ×3, i686 ×2 via cross, riscv64 musl, MIPS ×2 via cross + build-std + the `meow-common` atomic polyfill), plus the `rlim_cur` 32-bit fix and the `cfg(target_has_atomic)` portability work.

Non-blocking, fine as follow-ups:
1. Restore the explanatory matrix comments that got lost in the merge repair (armv7 boring-sys time64 note, aarch64 ADR-0007 hard-gate note, the #284 OpenWrt-archs rationale) — this week demonstrated exactly why they're worth keeping.
2. ADR-0011 asks for before/after `-Zprint-type-sizes` byte counts in the commit body for the `statistics.rs` change (expected answer on 64-bit: identical, since the aliases resolve to the same types — but it should be stated). ADR-0007 §1's mipsel note should also be updated to reflect the cross/build-std path now taken.
3. MIPS binaries are dynamically-linked musl (mips targets don't default `crt-static`) — fine on OpenWrt; add `-C target-feature=+crt-static` later if standalone static binaries are wanted.
4. On mips32 the traffic counters are u32 and wrap at 4 GiB — worth a line in the polyfill module docs.

Good work grinding through this — the final shape is solid.

xmh0511:
@madeye Please merge this PR and publish a new version.

crates/meow-listener/src/sniffer.rs
```diff
@@ -99,7 +99,7 @@ impl SnifferRuntime {
         if self.skip.search(host).is_some() {
             return;
         }
-        debug!("sniffer: {} → {}", metadata, host);
+        debug!("sniffer: {} ??{}", metadata, host);
```
madeye: **Corrupted string literal** (was `"sniffer: {} → {}"`): this is runtime log output, not a comment — the encoding mangling changed program behavior. Also: this file has no functional change in this PR; after fixing, `git checkout main -- crates/meow-listener/src/sniffer.rs` should drop it from the diff entirely.

crates/meow-listener/src/tun/route.rs
```diff
@@ -32,7 +32,7 @@ impl RouteGuard {
                 }
                 Err(e) => warn!(
                     "tun auto-route: failed to add {net} via if_index {if_index}: {e} \
-                     (continuing — the device subnet may already cover it)"
+                     (continuing ??the device subnet may already cover it)"
```
madeye: **Corrupted user-facing `warn!` text** (was `continuing — the device subnet…`). Same class of fix: this file is corruption-only, restore it from main. (The `warn!` in `tun/mod.rs` — "DNS is not in fake-ip mode ?? no routes installed" — is the third corrupted literal.)

.github/workflows/release.yml
```diff
@@ -209,4 +284,3 @@ jobs:
             dist/**/*.ipk
             dist/**/*.ipk.sha256
           generate_release_notes: true
```
madeye, line 307: **`fail_on_unmatched_files: true` went missing** from this step somewhere in the merge repair. Without it, a release with missing artifacts publishes silently instead of failing. Please restore it below `generate_release_notes`.

crates/meow-tunnel/src/udp.rs
```diff
@@ -53,7 +56,9 @@ impl UdpSession {
     pub fn idle_for(&self) -> Duration {
         let now = monotonic_ms();
         let last = self.last_activity_ms.load(Ordering::Relaxed);
-        Duration::from_millis(now.saturating_sub(last))
+        #[allow(clippy::useless_conversion)]
+        let last_u64: u64 = last.into();
+        Duration::from_millis(now.saturating_sub(last_u64))
     }
 }
```
madeye, line 65: **mips32 wrap bug**: `last` was stored as `monotonic_ms() as Uint` — truncated to `u32` on mips32 (wraps every ~49.7 days) — but `now` here is the full `u64`. After 49.7 days of uptime, `now.saturating_sub(last_u64)` is astronomically large for every session → all UDP sessions get evicted on every sweep, forever. Long-running routers are exactly the MIPS use case. Fix by comparing in the wrapped domain:

```rust
let now = monotonic_ms() as meow_common::atomic::Uint;
let last = self.last_activity_ms.load(Ordering::Relaxed);
Duration::from_millis(u64::from(now.wrapping_sub(last)))
```

(`wrapping_sub` in the `Uint` domain is correct for any real idle interval; on 64-bit targets it's identical to today's code.) The same pattern needs the same fix in `socks5_udp.rs`'s NAT `retain` sweep.

madeye (changes requested):
I re-read the **full PR diff** end-to-end (rather than the per-push deltas), and it surfaced three problems that CI green can't catch — two of them regressions that slipped past my earlier approval. Sorry for the flip-flop; this one is based on the complete picture.

### Blocking

**1. Encoding corruption across 12 files (68 lines).** The UTF-16 round-trip mangled every multi-byte character (`—`, `→`, `…`, `≥`) into `??` throughout `meow-listener`: `http_proxy.rs`, `mixed.rs`, `sniffer.rs`, `socks5.rs`, `socks5_udp.rs`, `tproxy/{firewall,mod,orig_dest}.rs`, `tun/{device,mod,route,udp}.rs`. Several of these files (`mixed.rs`, `sniffer.rs`, `tproxy/firewall.rs`, `tproxy/orig_dest.rs`, `tun/device.rs`, `tun/route.rs`) have **no functional change** — they're in this PR purely as comment vandalism. Worse, **three runtime string literals are corrupted** (inline comments below): a `debug!` in sniffer.rs and two user-facing `warn!` messages in tun. Repair: for the corruption-only files, `git checkout main -- <file>`; for the functionally-touched ones, fix the mangled lines by hand (or redo those commits from a UTF-8 editor). `git diff main | grep '??'` should come back empty.

**2. `fail_on_unmatched_files: true` was dropped from the publish step.** That flag is what makes a release fail loudly when an expected artifact is missing instead of silently publishing an incomplete release. It needs to come back.

**3. mips32: the millisecond clock wraps after 49.7 days and breaks UDP session eviction.** `monotonic_ms()` is a `u64` of ms since process start; storing it as `Uint` (= `u32` on mips32) truncates it, while `now` in `idle_for()` (`meow-tunnel/src/udp.rs`) and the NAT sweep in `socks5_udp.rs` stays full-width `u64`. For the first 49.7 days the truncated and full values agree; after that `now - last` is enormous and **every UDP session looks infinitely idle → continuous eviction and re-dialing**. Long uptimes are exactly the deployment profile of the MIPS/OpenWrt boxes this PR targets. Fix: do the comparison in the wrapped domain — store `monotonic_ms() as Uint` as now, and compute `(monotonic_ms() as Uint).wrapping_sub(last)` (correct for any interval < 49.7 days between touches) — or store seconds instead of ms.

### Non-blocking

- The new `#[allow(clippy::…)]` attributes are all bare — the repo's lint policy (CLAUDE.md) requires `#[allow(lint, reason = "…")]`, "no silent allows". One sweep to add reasons.
- Standing from earlier reviews: restore the deleted matrix comments (armv7 boring-sys, aarch64 ADR-0007 gate, #284 OpenWrt archs); ADR-0011 byte-count statement for `statistics.rs`/`udp.rs` in the commit body (expected: unchanged on 64-bit); update ADR-0007 §1's mipsel note to the cross/build-std reality.
- `updated_at` unix-seconds in `u32` on mips32 overflows in 2106 — fine, but worth a one-line comment where it's stored.

### What's good (unchanged verdict from the dry runs)

The polyfill design (`cfg(target_has_atomic = "64")`, not pointer width) is right; the `get_metrics` 501 gate is clean; the `statistics.rs` conversion is type-alias-only (no layout change on 64-bit); the workflow matrix/toolchain routing is validated 17/17 with artifacts inspected. Once the three blockers above are fixed this is genuinely done — I'll re-run the dry run on the fix push.

madeye:
`c2e6a7b`: two of three blockers verified fixed — the encoding corruption is fully gone (zero `??` in the diff; the corruption-only files no longer appear at all) and `fail_on_unmatched_files` is back. ✅

**But the mips32 clock fix is misapplied**, in both `udp.rs` and `socks5_udp.rs`. The change was `saturating_sub` → `wrapping_sub`, keeping `now` full-width:

```rust
let now = monotonic_ms();              // u64, full width
let last_u64: u64 = last.into();       // u32 truncated at store time, widened
now.wrapping_sub(last_u64)             // identical to the bug: no wrap ever happens in u64
```

After 49.7 days, `last` was stored mod 2³² but `now` keeps its high bits, so the difference is still ~2³² ms — `wrapping_sub` in the u64 domain only differs from `saturating_sub` when `last > now`, which isn't the failure mode. The subtraction has to happen in the **truncated (Uint) domain**, where both sides wrap together:

```rust
let now = monotonic_ms() as meow_common::atomic::Uint;   // truncate BEFORE subtracting
let last = self.last_activity_ms.load(Ordering::Relaxed);
Duration::from_millis(u64::from(now.wrapping_sub(last))) // exact for any idle < 49.7 days
```

and in the `socks5_udp.rs` sweep: `u64::from((monotonic_ms() as Uint).wrapping_sub(last)) < idle_ms`. On 64-bit targets `Uint = u64` so this is bit-identical to today's behavior.

Dry run on this head (validates the cleanup + everything else; it can't exercise the 49-day path): https://github.com/madeye/meow-rs/actions/runs/29564487100.

madeye:
Dry run on `c2e6a7b`: **17/17 green** — https://github.com/madeye/meow-rs/actions/runs/29564487100. The encoding cleanup and workflow repairs are fully validated. The only remaining item is the mips32 clock truncation fix from the comment above (runtime-behavioral, so CI can't exercise it) — once that lands I'll flip back to approve.

madeye:
`736b6d9`: the clock fix is now **semantically correct** in both places — truncate `now` to `Uint`, wrap-subtract in the truncated domain, widen the result. 

One mechanical problem: I ran clippy on this head locally and `u64::from(now.wrapping_sub(last))` fires `useless_conversion` twice on 64-bit hosts (where `Uint = u64`), so the `lint` job's `-D warnings` will reject it. The allows need to come back — and per the repo's lint policy they should carry a reason this time:

```rust
#[allow(clippy::useless_conversion, reason = "identity on 64-bit; u32→u64 widening on mips32")]
```

at both sites (`udp.rs` `idle_for` and the `socks5_udp.rs` sweep). That's the last item — dry run on this head dispatching now for the cross-target validation.

madeye:
I've pushed `f359db2` directly to this branch (maintainer edit) to close out the remaining review items in one go:

- **Fixed the E0658 compile error** in `846f936`: an attribute on the left operand of a binary expression isn't a statement position (`#[allow] u64::from(..) < idle_ms` fails with *attributes on expressions are experimental*) — restructured to an attributed `let` binding. Your `udp.rs` placement was fine (a lone tail call parses as a statement); `socks5_udp.rs` wasn't.
- Added `reason = "…"` to the six other bare `#[allow(clippy::…)]` sites from the polyfill work (repo policy: no silent allows).
- Documented the 32-bit fallback trade-offs in `meow-common/src/atomic.rs` (4 GiB counter wrap; 49.7-day ms-clock wrap + the wrapping_sub-in-truncated-domain rule).
- Restored the release.yml matrix comments lost in the merge repair (aarch64 ADR-0007 hard gate, #284 OpenWrt fan-out, armv7 boring-sys/vfpv3) and added MIPS/i686 notes recording this PR's findings.
- The commit body carries the ADR-0011 type-size statement (aliases resolve identically on 64-bit; `ConnectionInfo`/`UdpSession` layouts unchanged).

Verified locally before pushing: `cargo fmt --check` ✓, `cargo clippy --all-targets -- -D warnings` ✓, `cargo test --lib` 920 passed ✓. Dry run: https://github.com/madeye/meow-rs/actions/runs/29566334167

If PR CI and the dry run both come back green, this is merge-ready from my side.

madeye (approved):
Final state verified: dry run #26 on `f359db2` fully green (https://github.com/madeye/meow-rs/actions/runs/29566334167), all PR CI checks pass (lint/test/features/msrv/macos/windows/openwrt/tproxy/audit), and artifacts were binary-inspected earlier (riscv64 static-pie, i686-musl truly static, mipsel soft-float musl). All items from my earlier reviews are resolved. Merging. Thanks for the persistence — 26 dry runs later, this is a solid 17-target release matrix.

Issue #248, opened 2026-06-26
macOS transparent proxy is loopback-only and can't recover original destination


madeye:
## Background

#247 fixed two bugs that prevented meow's macOS (pf) transparent proxy from working at all (pf rule ordering → listener never started; anchor not nested under `com.apple/*` → rules never evaluated). After those fixes the host's own loopback-destined connections are now **intercepted** (verified in a macOS 26.4 VM via `scripts/verify-tproxy-setup.sh` + `scripts/verify-tproxy-test.sh`: 3/4 checks pass).

Two deeper limitations remain — meow's macOS transparent proxy is still not usable for real traffic.

## 1. Original destination not recovered (loopback handshake fails)

With `MATCH,REJECT` and the rig up, a client connecting to the test IP is diverted away from the echo server (interception works), but:

- `nc` to the rdr'd address reports `Operation timed out` — the redirected-to-`127.0.0.1` connection never completes its TCP handshake.
- meow's listener logs **nothing** for the connection — `get_original_dst` (`DIOCNATLOOK`) / `accept()` never yields a usable stream, so the original destination is never recovered.
- `verify-tproxy-test.sh` check `meow_logged_original_dst` fails.

pf state confirms the `rdr` fires (states exist for `… <- 10.88.0.1:9999 <- … 127.0.0.1:7893`), but the connection doesn't establish cleanly at meow. Suspected causes to investigate:
- martian-style src→loopback delivery (source is a non-loopback `lo0` alias, dst rewritten to `127.0.0.1`);
- the tproxy listener binding `127.0.0.1` vs. needing a routable/`0.0.0.0` bind;
- `DIOCNATLOOK` struct/ioctl ABI on recent macOS (the `DIOCNATLOOK` constant encodes a struct size; macOS 26.x may differ).

## 2. `rdr on lo0` only covers loopback

`build_pf_ruleset` emits `rdr pass on lo0 …`. Real outbound traffic (a browser to the internet) leaves via the physical interface (`en0`), **not** `lo0`, so it is never intercepted. As written, macOS tproxy can only ever touch loopback-destined traffic — which isn't useful for a general transparent proxy.

A proper fix likely needs:
- `rdr` on the real (and/or all) interface(s), not just `lo0`;
- correct reverse-path / source handling so the handshake completes and `DIOCNATLOOK` resolves;
[...]

PR #348, opened 2026-07-19
refactor(tproxy): centralize pf natlook query construction

floze-the-genius:
## Summary

- query macOS `DIOCNATLOOK` with `PF_OUT`, where pf stores the pre-translation destination for an `rdr` state
- centralize construction of the natlook query
- add a regression test for the lookup direction

This addresses the original-destination recovery part of #248. It intentionally does not widen the managed pf rule beyond `lo0` or redesign loop avoidance.

## Tests

- `cargo test -p meow-listener`
- `cargo clippy -p meow-listener --all-targets -- -D warnings`

The privileged `scripts/verify-tproxy-setup.sh` rig was not run because noninteractive sudo is unavailable on the test host.

### DIFF
--- a/crates/meow-listener/src/tproxy/orig_dest.rs
+++ b/crates/meow-listener/src/tproxy/orig_dest.rs
@@ -116,6 +116,15 @@ mod macos {
     // must be exactly 84 bytes or the ioctl is rejected / corrupts memory.
     const _: () = assert!(mem::size_of::<PfiocNatlook>() == 84);
 
+    fn natlook_query() -> PfiocNatlook {
+        PfiocNatlook {
+            af: PF_ADDR_IPV4,
+            proto: libc::IPPROTO_TCP as u8,
+            direction: PF_OUT,
+            ..Default::default()
+        }
+    }
+
     pub fn get_original_dst(stream: &TcpStream, listen_addr: SocketAddr) -> io::Result<SocketAddr> {
         let peer = stream.peer_addr().map_err(io::Error::other)?;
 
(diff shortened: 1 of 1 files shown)


madeye:
Thanks for this — you found the `PF_IN` → `PF_OUT` root cause first, and your diagnosis was exactly right. Apologies for the collision: the same direction fix just landed on `main` via #353 (verified against Squid's `Intercept.cc` and OpenBSD ftp-proxy precedent before merging), which unfortunately conflicts with this branch.

Your `redirected_lookup_uses_outbound_state` pinning test is a genuinely useful addition that #353 doesn't have. If you'd like the credit on `main` (you've earned it), feel free to rebase this down to the `natlook_query()` helper + test on top of current `main` — happy to merge that. Otherwise we can close this as superseded.

Note the fix is still pending runtime verification in a macOS VM (`scripts/verify-tproxy-setup.sh` + `verify-tproxy-test.sh`, acceptance = `meow_logged_original_dst` passing) — tracked in #248, which stays open for §2 (`rdr` only on `lo0`) as well.

madeye:
Rebased this onto main per the earlier comment (maintainer edit): the PF_IN→PF_OUT direction fix itself landed via #353, so this PR now carries the parts that weren't duplicated — the `natlook_query()` helper and the unit test pinning the PF_OUT lookup direction. Your authorship is preserved on the commit. Will merge once CI is green — thanks again for the fix and the diagnosis!

madeye:
Merged — thanks @floze-the-genius for the contribution! You correctly diagnosed the PF_IN/PF_OUT direction bug that broke original-destination recovery on macOS (#248 §1), and the `natlook_query()` helper and regression test from this PR are now on main with your authorship. Looking forward to more contributions!

PR #355, opened 2026-07-23
fix(listener): exempt ephemeral-port destinations from macOS pf rdr

### LINKED ISSUE #354: tproxy(macOS): broad rdr rule re-redirects meow's own SYN-ACK replies on lo0, wedging every handshake
Found during the macOS-VM runtime verification of #248 §1 (see https://github.com/madeye/meow-rs/issues/248#issuecomment-5053155931).

## Symptom

With the managed pf anchor loaded (`crates/meow-listener/src/tproxy/firewall.rs`), no intercepted TCP connection ever completes its handshake, and meow logs nothing — even though the pf anchor and listener are up. `pfctl -ss` shows the redirected connection stuck in `SYN_SENT` plus a second, reversed rdr state:

```
ALL tcp 127.0.0.1:7893 <- 10.88.0.1:9999 <- 10.88.0.1:49607   SYN_SENT:ESTABLISHED   (client's connection)
ALL tcp 127.0.0.1:7893 <- 10.88.0.1:49607 <- 10.88.0.1:9999   TIME_WAIT:TIME_WAIT    (meow's reply, re-redirected)
```

## Cause

The managed anchor installs:

```
rdr pass on lo0 inet proto tcp all -> 127.0.0.1 port 7893
```

On lo0 every packet traverses pf twice (out, then in). Meow's SYN-ACK reply is un-translated on the outbound pass, then re-enters lo0 **inbound**, where `all` matches it again as a fresh connection and rewrites its destination back into the listener port. The client never sees the SYN-ACK; the handshake never completes. The `pass out quick` uid/127.0.0.0-8 exceptions in the anchor are filter rules and don't exempt the reply from rdr translation.

## Evidence

Replacing the rule in-place with a destination-scoped variant (`rdr pass on lo0 inet proto tcp from any to 10.88.0.1 port 9999 -> ...`) immediately makes the full path work — `verify-tproxy-test.sh` goes from 3/4 (`meow_logged_original_dst` failing) to 4/4 on macOS 26.4 (arm64, tart VM), with meow correctly logging the recovered original destination.

## Direction

The rdr rule needs to be constrained so reply traffic can't re-match. Options to evaluate:

- exclude packets sourced from the proxy itself (e.g. `from ! 127.0.0.1` — needs testing for whether the un-translated reply still carries a 127.0.0.1 source at inbound evaluation time, the observed reversed state suggests it does not);
- scope the rdr to the destination sets meow actually wants to intercept (interacts with #248 §2, which already tracks the lo0-only scope problem);
- longer-term, the TUN inbound (`docs/tun.md`) sidesteps pf entirely on macOS and remains the strategic path.

Until fixed, macOS pf tproxy cannot work out of the box regardless of the natlook fix (#353/#348) — the two bugs stacked.

madeye:
## Summary

Fixes #354: the managed pf anchor's broad `rdr pass on lo0 proto tcp from any to any` also matched the tproxy listener's own SYN-ACK replies as they re-entered lo0 inbound (loopback double-pass), redirecting them back into the listener. Every intercepted handshake wedged in `SYN_SENT`, which is why macOS tproxy failed out of the box even after the natlook fix (#353/#348).

## Approach

pf translation rules cannot match TCP flags, so the discriminator is the destination port: replies always target the client's **ephemeral** source port, while fresh SYNs target service ports. The ruleset now leads with

```
no rdr on lo0 proto tcp from any to any port <first>:65535
```

where `<first>` is read from `net.inet.ip.portrange.first` at setup (fallback 49152) — translation rules are first-match, so the exemption precedes the `rdr`.

**Trade-off**: destinations listening on ephemeral-range ports bypass the proxy (fail open) instead of wedging (fail closed, the status quo). Documented in `docs/tproxy-gateway.md`.

## Verification

- Full regression bar: fmt, three-way clippy `-D warnings`, `RUSTDOCFLAGS="-D warnings" cargo doc`, `cargo test --lib` — all green.
- New unit tests: exemption emitted, precedes the rdr, range parameterized (not hardcoded).
- Ruleset parse-validated with `pfctl -n`.
- **e2e in a macOS 26.4 tart VM with the stock `verify-tproxy-*` rig**: previously 3/4 (`meow_logged_original_dst` failing, handshake stuck in SYN_SENT, reversed rdr state in `pfctl -ss`); with this fix **4/4 out of the box**, meow logs `10.88.0.1:49157 --> 10.88.0.1:9999 match MATCH() using REJECT`.

Together with #353/#348 this makes macOS local tproxy functional end-to-end for the first time. #248 §2 (lo0-only scope) remains open.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -83,7 +83,7 @@ The workspace has 13 crates (see also [ADR-0009](docs/adr/0009-cleanup-scope.md)
 
 ### Transparent-proxy gateway
 
-The built-in TProxy listener firewall (`meow-listener/src/tproxy/firewall.rs`) is `output`-chain/REDIRECT-based and only covers the **host's own** traffic — it is *not* a forwarding LAN gateway. To proxy *other* devices' traffic you add prerouting rules + a DNS hijack yourself. Helper scripts automate this: `scripts/tproxy-gateway-linux.sh` (nftables) and `scripts/tproxy-gateway-macos.sh` (pf, experimental). Full setup, DNS-mode (fake-ip vs redir-host) trade-offs, and systemd wiring are in [docs/tproxy-gateway.md](docs/tproxy-gateway.md). Note: the top-level `tproxy-port` hard-binds `127.0.0.1`; a gateway must declare the listener via `listeners:` with a non-loopback `listen`.
+The built-in TProxy listener firewall (`meow-listener/src/tproxy/firewall.rs`) is `output`-chain/REDIRECT-based and only covers the **host's own** traffic — it is *not* a forwarding LAN gateway. To proxy *other* devices' traffic you add prerouting rules + a DNS hijack yourself. Helper scripts automate this: `scripts/tproxy-gateway-linux.sh` (nftables) and `scripts/tproxy-gateway-macos.sh` (pf, experimental). Full setup, DNS-mode (fake-ip vs redir-host) trade-offs, and systemd wiring are in [docs/tproxy-gateway.md](docs/tproxy-gateway.md). macOS *local* tproxy (managed pf anchor + manual `route-to lo0` for real outbound traffic, IPv4 TCP only) is documented in [docs/tproxy-macos.md](docs/tproxy-macos.md). Note: the top-level `tproxy-port` hard-binds `127.0.0.1`; a gateway must declare the listener via `listeners:` with a non-loopback `listen`.
 
 ### TUN inbound (Windows transparent proxy)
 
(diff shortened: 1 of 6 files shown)


madeye:
Added docs in 09a2bc7: new [`docs/tproxy-macos.md`](https://github.com/madeye/meow-rs/blob/fix/354-pf-rdr-reply-reredirect/docs/tproxy-macos.md) usage guide for macOS tproxy — managed-anchor behavior, quick start, and a **VM-verified `route-to lo0` recipe** for intercepting real outbound traffic (curl to 1.1.1.1:80 relays end-to-end through meow; the root-UID escape rule is mandatory or proxied dials loop). Also refreshes the stale "handshake does not complete" note in `tproxy-local-macos.sh` and cross-links from the gateway doc, scripts README, and CLAUDE.md.

PR #360, opened 2026-07-25
fix(tun): support runtime TUN start/stop via config reload

xmh0511:
Previously TUN was only spawned once at startup and the PUT /configs handler ignored the tun: section entirely. nyanpasu's TUN toggle sends a config reload with tun.enable changed, but meow-rs never acted on it.

Changes:
- Add tun_handle to Tunnel for lifecycle management
- Reconcile TUN state in put_configs and commit_raw_candidate
- Store TUN handle at startup instead of fire-and-forget spawn
- Make parse_tun_config public for API crate reuse
- Forward listener-tun feature from meow-app to meow-api

### The following found issues are irrelevant to this PR; however, it's about the correctness of previously implementing `tun` mode(currently doesn't work):

The nyanpasu needs to have its privileges elevated; otherwise, there is an error
>TUN listener error: failed to create TUN device: 拒绝访问。 (os error 5) (requires root/CAP_NET_ADMIN on Linux/macOS; elevation + wintun.dll on Windows)

There are some warnings
> tun auto-route: failed to add 198.18.0.1/16 via if_index 40: 参数错误。 (os error 87) (continuing — the device subnet may already cover it)
> tun.stack: field is not supported in meow-rs and will be ignored; remove it to suppress this warning
> tun.auto-detect-interface: field is not supported in meow-rs and will be ignored; remove it to suppress this warning
> tun auto-route: failed to add 198.18.0.1/16 via if_index 39: 参数错误。 (os error 87) (continuing — the device subnet may already cover it)

There seems to loopback in the implementations
> UDP 169.254.235.176:137 --> 169.254.255.255:137 match MATCH() using

There are so many logs in the log panel when starting tun
````
10:03:20
info
UDP 172.19.0.1:62199 --> 172.19.0.3:137 match IP-CIDR(172.16.0.0/12) using DIRECT
10:03:20
info
UDP 172.19.0.1:62200 --> 172.19.0.3:137 match IP-CIDR(172.16.0.0/12) using DIRECT
10:03:20
info
UDP 172.19.0.1:62201 --> 172.19.0.3:137 match IP-CIDR(172.16.0.0/12) using DIRECT
10:03:20
info
UDP 172.19.0.1:62202 --> 172.19.0.3:137 match IP-CIDR(172.16.0.0/12) using DIRECT
10:03:20
info
UDP 172.19.0.1:62203 --> 172.19.0.3:137 match IP-CIDR(172.16.0.0/12) using DIRECT
10:03:20
info
UDP 172.19.0.1:62204 --> 172.19.0.3:137 match IP-CIDR(172.16.0.0/12) using DIRECT
10:03:20
info
UDP 172.19.0.1:62205 --> 172.19.0.3:137 match IP-CIDR(172.16.0.0/12) using DIRECT
10:03:20
info
UDP 172.19.0.1:62206 --> 172.19.0.3:137 match IP-CIDR(172.16.0.0/12) using DIRECT
10:03:20
info
UDP 172.19.0.1:62207 --> 172.19.0.3:137 match IP-CIDR(172.16.0.0/12) using DIRECT
10:03:20
info
UDP 172.19.0.1:62208 --> 172.19.0.3:137 match IP-CIDR(172.16.0.0/12) using DIRECT
10:03:20
info
UDP 172.19.0.1:62209 --> 172.19.0.3:137 match IP-CIDR(172.16.0.0/12) using DIRECT
10:03:20
info



UDP 172.19.0.1:65535 --> 172.19.0.3:137 match IP-CIDR(172.16.0.0/12) using DIRECT
10:05:52
info
UDP 172.19.0.1:65535 --> 172.19.0.3:137 match IP-CIDR(172.16.0.0/12) using DIRECT
10:05:52
info
UDP 172.19.0.1:65535 --> 172.19.0.3:137 match IP-CIDR(172.16.0.0/12) using DIRECT
````

### CODE REVIEW

**Comment by xmh0511** [comment]
Drive-by fixing the abuse of the route_manage, eliminate the following issue
> tun auto-route: failed to add 198.18.0.1/16 via if_index 40: 参数错误。 (os error 87) (continuing — the device subnet may already cover it)

However, the tproxy function still doesn't work.

File: crates/meow-tunnel/src/tunnel.rs [hunk 1]
```diff
@@ -421,6 +425,33 @@ impl Tunnel {
...
+        if let Some(prev) = slot.take() {
+            prev.abort();
+            info!("abandoned previous TUN listener");
+        }
+        *slot = Some(handle);
+        info!("TUN listener handle stored");
+    }
+
+    /// Abort the running TUN listener, if any.
+    pub fn stop_tun(&self) {
```
madeye: **Critical:** `abort()` here does not stop TUN. `TunListener::run()` spawns four detached tasks (`tokio::spawn(runner)`, both pumps in `device::spawn_pumps`, and `udp::run_udp`) that survive the abort:

- the pumps hold `Arc<AsyncDevice>` clones, so **the TUN device is never destroyed** (on Windows the wintun adapter persists);
- `run_udp` owns its own netstack UDP socket and a `Tunnel` clone, so **UDP flows and dns-hijack keep being proxied** — and since the device's own on-link subnet route still covers the fake-IP range, traffic keeps entering the "stopped" tun;
- the next enable creates a **second device** → name collision on Linux/Windows with an explicit `device` name.

Only the `RouteGuard` is cleaned up (it's owned by the aborted future). This needs a cooperative shutdown: pass a `CancellationToken`/watch channel into `run()`, `select!` on it, and have `run()` keep and abort its children's `JoinHandle`s on exit. `stop_tun` should signal and await teardown so a disable→enable flip can't race the old device. The same applies to the `prev.abort()` in `set_tun_handle` above.

crates/meow-tunnel/src/tunnel.rs
```diff
@@ -421,6 +425,33 @@ impl Tunnel {
...
+        let mut slot = self.inner.tun_handle.write();
+        if let Some(handle) = slot.take() {
+            handle.abort();
+            info!("TUN listener stopped");
+        }
+    }
+
+    /// Returns `true` when a TUN listener handle is currently held (the
+    /// listener may still be winding down after abort).
+    pub fn has_tun(&self) -> bool {
```
madeye, line 462: `has_tun` is never called anywhere in this PR. Either drop it, or use it — exposing TUN state to the API (e.g. a field on `GET /configs`) would be genuinely useful for nyanpasu's toggle. Note it would currently also report `true` when the spawned task already died (see the comment in `routes.rs` about failed starts).

crates/meow-api/src/routes.rs
```diff
@@ -1692,6 +1763,12 @@ async fn put_configs(
             state.tunnel.set_mode(mode);
         }
     }
+
+    // Reconcile TUN: start/stop listener when tun.enable changes.
+    {
+        let old_raw = state.raw_config.read();
+        reconcile_tun(&state.tunnel, &old_raw, &raw_config);
```
madeye: TOCTOU: the old config is read in one critical section, `reconcile_tun` runs, then the new config is written under a separate lock acquisition. Two concurrent PUTs can both observe `old_enable == false` and both spawn a listener (the second aborts the first — which leaks, per the `stop_tun` comment). Same pattern in `commit_raw_candidate`. Suggest doing the swap atomically: take `write()`, read the old enable flag, `mem::replace` the config, drop the guard, then act on the transition.

crates/meow-api/src/routes.rs
```diff
@@ -1574,6 +1582,69 @@ async fn get_group_delay(
...
+            udp_timeout: tun_cfg.udp_timeout,
+        },
+        "tun".to_string(),
+    );
+    let handle = tokio::spawn(async move {
+        if let Err(e) = listener.run().await {
+            tracing::error!("TUN listener error: {e}");
+        }
+    });
+    Some(handle)
```
madeye, line 1650: If `listener.run()` fails immediately — the exact unprivileged case from the PR description (`os error 5`) — the error is only logged, the dead `JoinHandle` is still stored, and the client gets 204. nyanpasu will render TUN as enabled while nothing is running. Consider a readiness signal (e.g. a oneshot sent from `run()` after device creation succeeds) so `put_configs` can surface the failure, or at minimum clear the stored handle when the task exits.

crates/meow-api/src/routes.rs
```diff
@@ -1574,6 +1582,69 @@ async fn get_group_delay(
...
+#[cfg(not(feature = "listener-tun"))]
+fn spawn_tun_from_raw(_tunnel: &Tunnel, raw: &RawConfig) -> Option<tokio::task::JoinHandle<()>> {
+    if raw.tun.as_ref().is_some_and(|t| t.enable) {
+        tracing::warn!("tun.enable is set but this build lacks the 'listener-tun' feature");
+    }
+    None
+}
+
+/// Reconcile TUN state: compare old vs new `tun.enable` and start/stop
+/// the TUN listener accordingly. Must be called after `raw_config` is updated.
```
madeye: This doc comment says "Must be called after `raw_config` is updated", but both call sites (correctly) call it *before* the write — it needs the pre-update value for `old_raw`. Please fix the comment to say it must be called before committing the new config.

Also worth a note here (or a follow-up issue): only `enable` is compared, so changing `device`/`mtu`/`dns-hijack` while enabled is silently ignored. mihomo restarts TUN on any `tun:` change, so client UIs may expect that.

crates/meow-api/src/routes.rs
```diff
@@ -1574,6 +1582,69 @@ async fn get_group_delay(
...
+        Ok(c) => c,
+        Err(e) => {
+            tracing::error!("tun config parse error: {e}");
+            return None;
+        }
+    };
+    if !tun_cfg.enable {
+        return None;
+    }
+    let listener = TunListener::new(
```
madeye, line 1614: This duplicates the `TunConfig` → `TunListenerConfig` mapping + spawn already in `meow-app/src/main.rs`. Extracting one shared helper (e.g. `impl From<&TunConfig> for TunListenerConfig` in `meow-listener`, or a small spawn fn) would keep the two paths from drifting.

crates/meow-listener/src/tun/route.rs
```diff
@@ -24,7 +24,7 @@ impl RouteGuard {
         let mut manager = RouteManager::new()?;
         let mut installed = Vec::with_capacity(nets.len());
         for net in nets {
-            let route = Route::new(net.addr(), net.prefix_len()).with_if_index(if_index);
+            let route = Route::new(net.network(), net.prefix_len()).with_if_index(if_index);
```
madeye, line 27: Good catch — `Route::new` with a host address (`198.18.0.1`) plus prefix 16 is exactly what produced the Windows `os error 87`; `net.network()` is correct, and delete stays consistent since the stored `Route` is the corrected one. This fix is unrelated to the lifecycle change though — it'd be cleaner as its own commit, or at least called out in the commit body.

madeye (commented):
Thanks for tackling this — the reconcile plumbing, feature forwarding, and the `route.rs` fix are all sound. However, the stop mechanism as implemented doesn't actually tear TUN down, which breaks the exact toggle use case this PR targets. Details inline; summary:

**Critical — `abort()` leaks the listener.** `TunListener::run()` spawns four detached tasks it does not own: the netstack runner, both device pumps, and `udp::run_udp`. Aborting the `run()` task only drops its own future, so after a "successful" stop the TUN device stays open (the pumps hold `Arc<AsyncDevice>` clones), UDP flows keep being proxied by the surviving `run_udp` task, and re-enabling creates a *second* device (name collision on Linux/Windows with an explicit `device` name). Only the `RouteGuard` gets cleaned up. The listener needs a cooperative shutdown: pass a `CancellationToken` (or watch channel) into `run()`, `select!` on it, and have `run()` own/abort its children on exit; `stop_tun` should signal and await teardown before a re-spawn is allowed.

Also worth addressing:
- TOCTOU between the `raw_config` read and write in both call sites (inline).
- A failed TUN start (e.g. the unprivileged case from the PR description) still returns 204 with a dead handle stored — nyanpasu will show the toggle as on.
- Only `enable` is compared; changing `device`/`mtu`/`dns-hijack` while enabled is silently ignored (mihomo restarts TUN on any `tun:` change) — fine for v1 but deserves a comment or follow-up issue.
- No tests: `reconcile_tun`'s four transitions (absent→enabled, enabled→disabled, unchanged, parse-error) are pure enough to unit-test.
- CI runs three-way clippy; the `--all-features` leg is the only one compiling the new `routes.rs` TUN code — please make sure it passes locally.

madeye (changes requested):
Requesting changes per the detailed review above (https://github.com/madeye/meow-rs/pull/360#pullrequestreview-4785896722). Blocking items:

1. **`stop_tun` must actually tear TUN down.** `abort()` on the `run()` task leaks the device pumps, the netstack runner, and the `run_udp` task — the device stays open, UDP keeps being proxied, and a disable→enable flip collides on the device name. Needs a cooperative shutdown (cancellation token + `run()` owning/aborting its children), with teardown completing before a re-spawn.
2. **Fix the `raw_config` read/write TOCTOU** in `put_configs` and `commit_raw_candidate` so concurrent reloads can't double-spawn.
3. **Surface TUN start failure to the API caller** (or at least clear the stored handle when the task exits) — currently an unprivileged start logs an error but returns 204, and the toggle appears on.

The rest (doc-comment fix, dead `has_tun`, spawn-code dedup, unit tests for the reconcile transitions, splitting out the `route.rs` fix) are non-blocking but would be good to land in the same pass.

crates/meow-listener/src/tun/mod.rs
```diff
@@ -53,6 +53,41 @@ use tracing::{info, warn};
...
+
+    /// Track a `JoinHandle<T>` by recording its `AbortHandle`. Type-erased
+    /// via `AbortHandle` so heterogeneous task return types (e.g. `()`,
+    /// `io::Result<()>`) coexist in the same collection.
+    fn push<T>(&mut self, h: &tokio::task::JoinHandle<T>) {
+        self.aborts.push(h.abort_handle());
+    }
+
+    /// Spawn a future and automatically track its handle.
+    fn spawn(&mut self, f: impl std::future::Future<Output = ()> + Send + 'static) {
```
madeye, line 82: **Unbounded growth:** the accept loop calls `tasks.spawn(...)` for every TCP flow, and `aborts` is never reaped. An `AbortHandle` keeps the task's allocation alive even after the task completes, so a long-running TUN session accumulates memory proportional to the total number of flows ever accepted.

Simple fix — reap completed handles on insertion:

```rust
fn spawn(&mut self, f: impl std::future::Future<Output = ()> + Send + 'static) {
    self.aborts.retain(|a| !a.is_finished());
    let h = tokio::spawn(f);
    self.aborts.push(h.abort_handle());
}
```

(or keep the four infra tasks in the `Vec` and put per-flow tasks in a `tokio::task::JoinSet`, which reaps on `poll`).

crates/meow-app/src/main.rs
```diff
@@ -985,23 +987,39 @@ async fn run(
...
+            let handle = tokio::spawn(async move {
                 if let Err(e) = listener.run().await {
                     error!("TUN listener error: {}", e);
                 }
             });
+
+            // Await device readiness before treating TUN as "running".
+            // If device creation fails (permission denied, etc.), don't
+            // store a dead JoinHandle — the next reconcile will retry.
+            match tokio::time::timeout(std::time::Duration::from_secs(5), ready_rx).await {
```
madeye: The last commit bumped the readiness timeout to 30 s in `routes.rs` (commit message: wintun + netstack init can be slow), but this startup path still uses 5 s — and the error text below still says "timed out after 5 s". Whatever value is right for Windows should be a shared constant used by both paths, e.g. `pub const TUN_STARTUP_TIMEOUT: Duration` next to `tun_config_to_listener_config` in `meow-api`, or on `TunListener` itself.

crates/meow-tunnel/src/tunnel.rs
```diff
@@ -421,6 +425,39 @@ impl Tunnel {
...
+        // parking_lot RwLock write guard is dropped here — safe to .await
+        if let Some(handle) = handle {
+            handle.abort();
+            let _ = handle.await;
+            info!("TUN listener stopped");
+        }
+    }
+
+    /// Returns `true` when a TUN listener handle is currently held.
+    pub fn has_tun(&self) -> bool {
```
madeye, line 462: `has_tun` returns `true` for a dead handle: if the listener crashes at runtime (e.g. a pump error after successful startup), the handle stays stored and `GET /configs` reports `tun-enable: true` until the next reconcile. Suggest:

```rust
self.inner.tun_handle.read().as_ref().is_some_and(|h| !h.is_finished())
```

`stop_tun` still works either way since it takes the slot unconditionally.

crates/meow-tunnel/src/tunnel.rs
```diff
@@ -421,6 +425,39 @@ impl Tunnel {
...
+    /// released before the new one starts.
+    pub async fn set_tun_handle(&self, handle: tokio::task::JoinHandle<()>) {
+        let prev = self.inner.tun_handle.write().replace(handle);
+        // parking_lot RwLock write guard is dropped here — safe to .await
+        if let Some(prev) = prev {
+            prev.abort();
+            // Await teardown: TaskGroup::drop aborts child tasks, and
+            // their Arc<AsyncDevice> refs are released. This prevents
+            // device-name collision on a disable→enable flip.
+            let _ = prev.await;
```
madeye, line 441: Nit on the comment: `prev.await` only guarantees the *parent* future is dropped — `TaskGroup::drop` has *requested* abort of the children at that point, but the children are torn down when the runtime processes those aborts, so the `Arc<AsyncDevice>` refs may be released shortly *after* this returns. The window is tiny in practice, but the comment claims more than the code guarantees; worth softening it (or, if you want the hard guarantee, have `run()` own the child `JoinHandle`s and await them on cancellation via a shutdown token).

crates/meow-api/src/routes.rs
```diff
@@ -1692,7 +1808,42 @@ async fn put_configs(
...
+        };
+        // parking_lot guard dropped — safe to .await below
+
+        if old_enable != new_enable {
+            if new_enable {
+                // Clone RawConfig inside the read-lock scope so the
+                // parking_lot guard (!Send) is dropped before the
+                // first .await — axum 0.8 requires handler futures to
+                // be Send.
+                let raw_snapshot = state.raw_config.read().clone();
```
madeye: Two nits on this block:

1. The re-read here is roundabout — `raw_config` was just swapped to `raw_config`/`candidate` two lines up, so cloning the candidate *before* moving it into the lock is simpler and immune to writers that don't hold `config_mutation_lock` (the PATCH handler writes `raw_config` directly).
2. This ~25-line reconcile sequence is duplicated verbatim in `commit_raw_candidate` — worth re-extracting into an async helper (e.g. `async fn reconcile_tun(state, old_enable, new_enable)`), which would also give the transition logic a single seam to unit-test.

madeye (commented):
Re-reviewed after the latest commits — thanks for the thorough rework. All three blocking items from the previous review are substantively addressed:

- **Teardown** — `TaskGroup` aborts the runner, pumps, `run_udp`, and flow tasks when `run()` is dropped, and `stop_tun`/`set_tun_handle` abort **and await** the parent. Right shape. ✅
- **TOCTOU** — `config_mutation_lock` serialises both mutation paths, and the old-enable read + config swap are atomic under the write guard. I verified `put_configs` and `commit_raw_candidate` are disjoint (no nested-lock deadlock) and the PATCH handler that writes `raw_config` without the lock only touches mode/log-level. ✅
- **Failure surfacing** — readiness oneshot, no dead handle stored on failure, and `tun-enable` on `GET /configs`. ✅

CI is green across the matrix. Two new issues in the rework should be fixed before merge (inline): the unbounded `AbortHandle` accumulation in `TaskGroup`, and the 5 s vs 30 s startup-timeout split between `main.rs` and `routes.rs`. The rest of the inline notes are non-blocking.

Still open from last round (non-blocking): unit tests for the enable/disable transitions — the lock made it into the test fixtures but no new tests cover the reconcile itself. Also, a pre-existing pattern worth a follow-up issue rather than this PR: the subscription endpoints clone the config, `await` a network fetch, then commit — a concurrent TUN toggle in that window gets silently reverted (lost update), which now carries a device start/stop side effect.

madeye (approved):
Pushed the review follow-ups directly to this branch as d666cdd (maintainer edits):

- `TaskGroup::push` now reaps finished `AbortHandle`s, fixing the unbounded per-flow growth.
- `TUN_STARTUP_TIMEOUT` (30 s) is a shared `meow-api` constant used by both the startup and reload paths — `main.rs` was still on 5 s.
- `Tunnel::has_tun` checks `is_finished()`, so `GET /configs` no longer reports a crashed listener as enabled.
- Extracted the duplicated swap+reconcile block into `swap_config_and_reconcile_tun` (clones only on an off→on transition, no post-swap re-read).
- Softened the `set_tun_handle` teardown comment to match the actual abort guarantee.
- Added lifecycle unit tests (store/stop idempotence, replace-aborts-and-awaits, dead-task reporting).

Verified locally: `cargo fmt --check`, all three clippy legs with `-D warnings`, `cargo doc` with `-D warnings`, and `cargo test --lib` (936 passed). With those on top, the runtime TUN lifecycle looks correct — approving. The subscription-commit lost-update window (config cloned before the network fetch, committed after) remains a pre-existing pattern worth a separate issue.

Thanks @xmh0511 for the thorough rework — the TaskGroup + readiness-signal + mutation-lock structure came out clean.

PR #368, opened 2026-07-28
fix(dns): validate upstream responses

### LINKED ISSUE #365: DNS client accepts mismatched responses and unrelated answers, enabling cache poisoning
## Summary

`DnsClient::query()` generates a random DNS transaction ID but accepts the first decoded response without checking that it matches the outstanding query. `lookup_ip()` then collects every A/AAAA record in the answer section, regardless of the record owner.

A connected UDP socket only filters the peer address and port. It does not authenticate a packet whose source tuple has been spoofed. Consequently, an off-path attacker who can spoof the configured upstream address can race a forged response without guessing the 16-bit transaction ID.

## Affected code

- [`crates/meow-dns/src/client.rs:276-337`](https://github.com/madeye/meow-rs/blob/51e8f861ac242740e1f9b0886ad0872ef0cff84f/crates/meow-dns/src/client.rs#L276-L337)
- [`crates/meow-dns/src/client.rs:424-439`](https://github.com/madeye/meow-rs/blob/51e8f861ac242740e1f9b0886ad0872ef0cff84f/crates/meow-dns/src/client.rs#L424-L439)
- [`crates/meow-dns/src/resolver.rs:343-425`](https://github.com/madeye/meow-rs/blob/51e8f861ac242740e1f9b0886ad0872ef0cff84f/crates/meow-dns/src/resolver.rs#L343-L425)
- [`crates/meow-dns/src/resolver.rs:971-1039`](https://github.com/madeye/meow-rs/blob/51e8f861ac242740e1f9b0886ad0872ef0cff84f/crates/meow-dns/src/resolver.rs#L971-L1039)

When DNS is not explicitly enabled, the default resolver is also plain UDP to `8.8.8.8:53`:

- [`crates/meow-config/src/dns_parser.rs:26-37`](https://github.com/madeye/meow-rs/blob/51e8f861ac242740e1f9b0886ad0872ef0cff84f/crates/meow-config/src/dns_parser.rs#L26-L37)

## Reproduction

A temporary UDP upstream was used to return a response with all of the following:

- a transaction ID different from the request;
- a different question name;
- an A record whose owner was unrelated to the requested name.

The response was accepted:

```text
addrs returned for victim.example: [6.6.6.6]
```

The resolver subsequently stores the address under the requested host via `cache.put_with_source()`. With multiple upstreams, `query_pool()` queries in parallel and the first non-empty result wins, increasing the forged response's chance of winning the race.

## Expected behavior

A response must be associated with the outstanding query before any flags or records are trusted:

- response ID matches the request ID;
- QR indicates a response and opcode is appropriate;
- the question count and question name/class/type match the request;
- for UDP, mismatched datagrams are discarded while receiving continues on the same socket under the original overall timeout;
- validation happens before honoring `TC=1` and retrying over TCP;
- TCP, DoT, DoH, and proxied TCP exchanges reject mismatched responses as protocol errors.

For A/AAAA lookup, only address records whose owner is reachable from the requested name through a valid CNAME chain should be returned. Unrelated answer records must not be cached. The effective TTL should not outlive any accepted CNAME link.

The generic forwarding path should receive equivalent response-association validation before answer records are re-emitted.

## Regression tests

Please cover at least:

1. wrong ID followed by a valid UDP response;
2. wrong question name, type, and class;
3. wrong-ID `TC=1` response does not trigger TCP fallback;
4. unrelated A/AAAA owner is ignored;
5. single-hop, multi-hop, and out-of-order CNAME chains remain accepted;
6. unrelated records alongside a valid CNAME chain are ignored;
7. mismatched framed responses over TCP are rejected.

## Attribution

gengjiawen:
## Summary

- validate transaction ID, QR, opcode, and the complete question before accepting an upstream response
- ignore unrelated UDP datagrams under the original query timeout and validate a response before honoring TC fallback
- accept A/AAAA answers only for the requested owner or names reachable through its CNAME chain, with CNAME TTLs limiting cache lifetime
- traverse CNAME answers through a bounded, indexed chain instead of an unbounded fixed-point scan
- apply response association checks to TCP, proxied TCP, DoT, and DoH responses too

## Tests

- wrong UDP ID followed by a valid response
- wrong-ID truncated UDP response does not trigger TCP fallback
- mismatched metadata, question name/type/class, and framed TCP response
- unrelated A records and an out-of-order multi-hop CNAME chain
- long and cyclic CNAME input remains bounded
- cargo test -p meow-dns
- cargo clippy -p meow-dns --all-targets -- -D warnings
- cargo fmt --all -- --check

## Attribution

Fixes #365

### DIFF
--- a/crates/meow-dns/src/client.rs
+++ b/crates/meow-dns/src/client.rs
@@ -8,6 +8,7 @@
 use hickory_proto::op::{Message, MessageType, OpCode, Query, ResponseCode};
 use hickory_proto::rr::{Name, RData, Record, RecordType};
 use hickory_proto::serialize::binary::{BinDecodable, BinEncodable};
+use std::collections::{hash_map::Entry, HashMap, HashSet};
 use std::future::Future;
 use std::io;
 use std::net::{IpAddr, SocketAddr};
@@ -470,6 +471,18 @@ fn ip_from_record(rec: &Record) -> Option<IpAddr> {
     }
 }
 
+fn canonical_name(name: &Name) -> Name {
+    let mut canonical = name.to_lowercase();
+    canonical.set_fqdn(true);
+    canonical
+}
+
+struct CnameLink {
+    target: Name,
+    ttl: u32,
+    ambiguous: bool,
+}
+
 fn relevant_ip_answers(message: &Message) -> (Vec<IpAddr>, Option<u32>) {
     let Some(question) = message.queries.first() else {
         return (Vec::new(), None);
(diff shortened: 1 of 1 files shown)


madeye (approved):
Reviewed the full diff. This is a solid, well-scoped hardening of the DNS client and I'm approving it.

**What I checked**

- Response association (ID / QR / opcode / exact question match) is now enforced on every transport — UDP, TCP, proxied TCP, DoT, DoH — because validation lives in `exchange()` rather than one call site. No path returns an unvalidated `Message`.
- The UDP receive loop correctly *ignores* non-matching datagrams and keeps waiting under the original timeout instead of failing the query — so an off-path spoofer can no longer DoS a lookup with one garbage packet, and a spoofed TC=1 can no longer force a TCP downgrade (TC is only honored on a validated response).
- `relevant_ip_answers` is order-independent, walks the CNAME chain from the QNAME with loop detection, rejects ambiguous (multi-target) CNAME owners rather than picking a branch, and caps the cached TTL by the CNAME link TTLs. Reverse-ordered hostile chains stay linear via the indexed map.
- Scope is right: A/AAAA caching flows through `query_ips` (filtered), while `query_pool_generic` (TXT/MX/SRV/HTTPS pass-through in the resolver) still benefits from the new question-association check in `exchange()`.
- Tests exercise the live socket paths (wrong-ID-then-valid UDP, wrong-ID TC not triggering TCP fallback, mismatched framed TCP), not just the pure functions.

**One non-blocking nit**

`relevant_ip_answers` accepts A/AAAA records owned by *intermediate* chain nodes, i.e. a name that also has a CNAME link (RFC 1034 §3.6.2 forbids other data at a CNAME owner). Since every reachable name is asserted by the same server response, this isn't a poisoning vector beyond what the upstream can already do — but skipping address records at owners present in `cname_links` would be slightly stricter. Fine as a follow-up or never.

CI is green across all platforms including MSRV and the tproxy/openwrt jobs. Merging.

PR #369, opened 2026-07-28
fix(listener): parse HTTP headers as strict bytes

### LINKED ISSUE #366: HTTP proxy mixes strict framing with LF-tolerant lossy header rewriting
## Summary

The HTTP proxy locates the request-header boundary using only `\r\n\r\n`, but then parses and rewrites that block through `String::from_utf8_lossy(...).lines()`.

These two parsing rules disagree: `str::lines()` treats a bare LF as a line boundary, and the rewrite canonicalizes every resulting line to CRLF. Invalid UTF-8 header bytes are replaced with U+FFFD before forwarding.

- [`crates/meow-listener/src/http_proxy.rs:50-87`](https://github.com/madeye/meow-rs/blob/51e8f861ac242740e1f9b0886ad0872ef0cff84f/crates/meow-listener/src/http_proxy.rs#L50-L87)
- [`crates/meow-listener/src/http_proxy.rs:123-136`](https://github.com/madeye/meow-rs/blob/51e8f861ac242740e1f9b0886ad0872ef0cff84f/crates/meow-listener/src/http_proxy.rs#L123-L136)
- [`crates/meow-listener/src/http_proxy.rs:302-332`](https://github.com/madeye/meow-rs/blob/51e8f861ac242740e1f9b0886ad0872ef0cff84f/crates/meow-listener/src/http_proxy.rs#L302-L332)

## Reproduction

A raw client sent the following header bytes through the proxy:

```text
X-A: 1\nContent-Length: 5\r\n
```

The origin received:

```text
X-A: 1\r\nContent-Length: 5\r\n
```

The proxy therefore turned an ambiguous input line into a clean upstream `Content-Length` header.

The same reproduction showed two additional mutations:

- an intermediate `\n\n` makes the rewrite loop stop, silently dropping subsequent fields;
- an invalid UTF-8 byte such as `caf\xe9` is forwarded as `caf\xef\xbf\xbd`.

The latter also invalidates the comment that the rewritten request cannot grow beyond the original request.

## Impact

RFC 9112 permits a recipient to recognize a lone LF as a line terminator, so accepting LF is not by itself a protocol violation. The defect is that framing, parsing, and forwarding use different rules, after which the intermediary canonicalizes or drops bytes.

This is a concrete request-corruption bug and creates a parser differential that can contribute to request smuggling when another recipient interprets the original or rewritten message differently. The current reproduction demonstrates the differential and corruption; it should not be read as a complete cross-boundary request-smuggling exploit on its own.

Bare CR is also currently forwarded without being rejected or normalized as required for safe intermediary handling.

## Expected behavior

Use one consistent HTTP/1 parsing policy on raw bytes:

- parse the request line and header block as `&[u8]`;
- either consistently support LF framing or, preferably for an intermediary, reject any bare LF or bare CR with `400 Bad Request`;
- split accepted fields only on CRLF;
- preserve original field-value bytes, including valid `obs-text`, when forwarding;
- remove proxy-specific and hop-by-hop fields without converting the entire block through a lossy UTF-8 string;
- validate conflicting or malformed `Content-Length` / `Transfer-Encoding` framing before forwarding.

Malformed syntax should be rejected before authentication or route dispatch.

## Regression tests

Please cover:

1. bare LF in a field line is rejected and is not normalized upstream;
2. bare CR is rejected;
3. `\n\n` cannot truncate or silently discard later bytes;
4. non-UTF-8 field-value bytes are either preserved or the entire request is rejected, never lossy-converted;
5. conflicting Content-Length values and TE+CL combinations are rejected;
6. valid absolute-form requests still have only their request target and proxy-specific fields rewritten.

## Related issue

#367 tracks per-request routing on persistent HTTP proxy connections. A shared strict HTTP/1 framing layer may address both issues.

## Attribution

gengjiawen:
## Summary

- reject bare CR/LF and malformed HTTP/1 request heads with 400 before authentication or route dispatch
- parse field names and values from raw bytes, preserving legal obs-text when forwarding
- rewrite only the absolute-form request target and remove proxy-specific headers without lossy UTF-8 conversion
- reject conflicting Content-Length values, malformed transfer codings, and Transfer-Encoding plus Content-Length ambiguity
- accept valid transfer-coding parameters, including quoted commas and quoted-pairs, without weakening framing checks

## Tests

- bare LF and bare CR are rejected by the live listener
- an intermediate LF/LF sequence cannot truncate later headers
- non-UTF-8 field bytes survive the full proxy-to-origin path unchanged
- body line feeds arriving with the headers remain body bytes
- conflicting Content-Length and TE+CL combinations are rejected
- valid and malformed parameterized transfer codings
- cargo test -p meow-listener
- cargo clippy -p meow-listener --all-targets -- -D warnings
- cargo fmt --all -- --check

## Attribution

Fixes #366

### DIFF
--- a/crates/meow-listener/src/http_proxy.rs
+++ b/crates/meow-listener/src/http_proxy.rs
@@ -483,7 +483,7 @@ fn parse_request_head(head: &[u8]) -> Result<RequestHead<'_>, &'static str> {
 fn validate_message_framing(headers: &[HeaderField<'_>]) -> Result<(), &'static str> {
     let mut content_length = None;
     let mut has_content_length = false;
-    let mut transfer_codings: Vec<&[u8]> = Vec::new();
+    let mut transfer_codings = Vec::new();
 
     for header in headers {
         if header.name.eq_ignore_ascii_case(b"content-length") {
@@ -505,13 +505,7 @@ fn validate_message_framing(headers: &[HeaderField<'_>]) -> Result<(), &'static 
                 content_length = Some(parsed);
             }
         } else if header.name.eq_ignore_ascii_case(b"transfer-encoding") {
-            for coding in header.value.split(|byte| *byte == b',') {
-                let coding = trim_ows(coding);
-                if coding.is_empty() || !coding.iter().copied().all(is_tchar) {
-                    return Err("invalid Transfer-Encoding");
-                }
-                transfer_codings.push(coding);
-            }
+            parse_transfer_codings(header.value, &mut transfer_codings)?;
         }
     }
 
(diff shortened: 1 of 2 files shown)


madeye (approved):
Reviewed the full diff line by line. Approving.

**What I checked**

- The incremental bare-CR/LF check is correct: it validates only header bytes (body bytes past `header_end` are excluded, covered by the `a\nb` body test), and a final CR at the end of a read is correctly left undecided until the next read completes or refutes the CRLF.
- `parse_request_head` grammar is strict per RFC 9112: single-SP request line with exactly three parts, tchar method, visible-ASCII target, version restricted to HTTP/1.0|1.1; header names tchar-only, values `HTAB / 0x20-0xFF minus DEL` so obs-text survives. Obs-fold lines fail the name check and get a clean 400, which the RFC permits for an intermediary.
- Forwarding uses the raw header slices, so the old `from_utf8_lossy` corruption (U+FFFD rewriting changing byte lengths) and `.lines()` bare-LF normalization — the actual smuggling vector from #366 — are both gone.
- Framing validation: list-form and repeated Content-Length must agree digit-for-digit with checked u64 arithmetic; TE+CL rejected; chunked must appear exactly once, last, and parameterless; the transfer-coding parser handles quoted-string parameters (incl. quoted commas and quoted-pairs) without weakening the list split.
- 400 responses are only sent for `InvalidData`-class failures, with `Connection: close` + `Content-Length: 0`, before auth or route dispatch.
- Tests exercise the live listener through `handle_http`, including the end-to-end obs-text preservation to a real origin.

**Non-blocking nits**

- `rewrite_plain_request`'s capacity hint omits the header `raw` byte lengths, so the Vec typically reallocates once or twice; summing `header.raw.len()` would make the hint exact. (This is per-request setup, not relay path, so no invariant concern.)
- `has_bare_line_ending` rescans the buffer from index 0 on every socket read — worst case O(MAX_HEADERS²) ≈ tens of ms for an attacker dribbling 1-byte reads, bounded by the handshake timeout. Fine as-is; could resume from the previous scan position if it ever shows up in profiles.

CI is green on all jobs. Merging this first so #370 rebases down to its own changes.

PR #370, opened 2026-07-28
fix(listener): close plain HTTP after one request

### LINKED ISSUE #367: HTTP proxy routes only the first request on persistent client connections
## Summary

The plain HTTP proxy parses, authenticates, routes, and rewrites exactly one request. After dialing the selected destination, it hands the entire client connection to `copy_bidirectional_buf_tracked()` as an opaque byte stream.

Any later HTTP/1.1 request on the same persistent client connection therefore bypasses request parsing and route selection. It is sent on the TCP connection opened for the first request, and its absolute-form request target is not rewritten.

- [`crates/meow-listener/src/http_proxy.rs:36-137`](https://github.com/madeye/meow-rs/blob/51e8f861ac242740e1f9b0886ad0872ef0cff84f/crates/meow-listener/src/http_proxy.rs#L36-L137)
- [`crates/meow-listener/src/http_proxy.rs:242-300`](https://github.com/madeye/meow-rs/blob/51e8f861ac242740e1f9b0886ad0872ef0cff84f/crates/meow-listener/src/http_proxy.rs#L242-L300)
- [`crates/meow-listener/src/http_proxy.rs:302-365`](https://github.com/madeye/meow-rs/blob/51e8f861ac242740e1f9b0886ad0872ef0cff84f/crates/meow-listener/src/http_proxy.rs#L302-L365)

## Reproduction

Two local origin servers were started:

- SERVER_A on `127.0.0.1:18001`;
- SERVER_B on `127.0.0.1:18002`.

A single keep-alive connection to meow sent a first request for SERVER_A and then a second request for SERVER_B. The second request was observed by SERVER_A:

```text
SERVED-BY=SERVER_A PATH=http://127.0.0.1:18002/two HOST=127.0.0.1:18002
```

This confirms both that the second request used the first destination connection and that its absolute URI was forwarded without rewriting.

## Impact

HTTP/1.1 uses persistent connections by default and permits multiple requests, including pipelined requests, on one connection. Forward-proxy clients can reuse their connection to the proxy for requests targeting different origins.

For every request after the first:

- routing rules are not evaluated;
- the destination and proxy adapter are not selected again;
- absolute-form is not converted to origin-form;
- sniffing and request-level processing are skipped;
- all connection statistics remain attributed to the first destination and rule.

This directly breaks rule-based routing during normal plain-HTTP proxy use.

## Expected behavior

Each plain HTTP request must have an explicit message boundary and independently perform:

1. request parsing and proxy authentication;
2. absolute-form target parsing and origin-form rewriting;
3. metadata preprocessing, sniffing, and rule matching;
4. destination/proxy selection;
5. per-request or accurately scoped connection accounting.

A full request/response loop is the preferred fix. Outbound connections may only be reused when their destination and routing decision are compatible.

A one-request-per-client-connection fallback is acceptable only if it is implemented completely. Merely injecting `Connection: close` into the first upstream request is insufficient because already-buffered or pipelined bytes would still be copied to the first destination. Such a fallback must frame and forward exactly the first request body, communicate close semantics to the client, and prevent later request bytes from entering the first outbound connection.

## Regression tests

Please cover:

1. two sequential keep-alive requests to different origins;
2. two pipelined requests to different origins;
3. each request receives its own rule and proxy selection;
4. each absolute URI is rewritten before reaching its origin;
5. POST bodies split across reads do not consume bytes from the next request;
6. chunked request and response bodies preserve message boundaries;
7. connection statistics are not all assigned to the first target.

Related: #366 tracks the HTTP header framing and byte-preservation defect. A shared strict HTTP/1 framing layer may address both issues.

## Attribution

gengjiawen:
## Summary

- limit each non-upgraded plain HTTP client connection to one framed request while leaving CONNECT tunnels unchanged
- force `Connection: close` upstream and rewrite the final response with close semantics for the client
- strip `Connection`, `Keep-Alive`, proxy fields, and Connection-nominated hop-by-hop fields
- forward exactly the declared Content-Length or chunked body, including trailers, while withholding and draining pipelined request bytes
- preserve informational responses before rewriting the final response
- preserve valid bodyless HTTP/1.1 Upgrade requests, including `ws://`, and switch to an opaque tunnel only after a validated `101`
- observe client EOF after the request boundary so silent origins are reaped by the relay linger

## Behavior and tradeoffs

This branch is rebased onto `main`, which already contains the strict byte-oriented parser from #369.

The one-request safety fallback opens a fresh outbound connection for every ordinary plain HTTP request and can repeat an encrypted adapter handshake. A future per-request routing loop can recover connection reuse while still rerouting every request independently.

For Upgrade handshakes, the proxy preserves and normalizes `Connection: upgrade` and `Upgrade`, removes other hop-by-hop fields, and withholds client protocol bytes until the origin returns a valid `101 Switching Protocols`. It then forwards those held bytes and resumes ordinary bidirectional tunnel and half-close semantics. If the origin declines the Upgrade with a normal final response, the proxy does not forward the held protocol bytes and completes that response with `Connection: close`.

After an ordinary framed request body, the proxy drains and discards at most 64 KiB of client data. Client EOF completes the upload relay direction and arms the existing half-close linger; an upstream shutdown gate prevents that EOF from sending early TLS close semantics before the final response. Once the final response is complete, the proxy drains currently queued client bytes and closes as soon as the socket becomes idle, preventing unread pipeline data from resetting and truncating a large response.

## Footprint

Measured with `RUSTC_BOOTSTRAP=1 cargo rustc -p meow-listener --lib -- -Zprint-type-sizes`:

| revision | `handle_http_inner` | `handle_http` | `SingleRequestClient` |
|---|---:|---:|---:|
| `origin/main` (`7c43cbb`) | 19,184 B | 19,376 B | n/a |
| final (`223915b`) | 19,176 B | 19,368 B | 9,344 B |

The response and discard buffers live only in the boxed plain-HTTP state machine. This adds one plain-HTTP-only setup allocation, leaves the final connection future 8 B smaller than current `main`, and keeps CONNECT free of the 9,344 B state. `Box<SingleRequestClient>` is 8 B; the relay helper and both caller-supplied 4,096 B relay buffers remain allocation-free.

## Tests

- pipelined requests to different origins cannot enter the first origin connection
- split Content-Length bodies stop exactly before a pipelined request
- chunked bodies and trailers stop exactly before a pipelined request
- final responses replace hop-by-hop headers after informational responses
- client EOF arms half-close linger without half-closing the upstream adapter early
- queued pipeline data does not reset or truncate a 1 MiB response
- `Expect: 100-continue` forwards a body sent after the interim response
- HTTP/1.0 requests and response heads split across upstream writes are supported
- response Connection options reject framing-field nominations and accept empty list elements
- a real `ws://` Upgrade holds early protocol bytes until a validated `101`, then relays them losslessly
- a declined Upgrade closes without forwarding early protocol bytes
- `cargo test --lib`
- `cargo test -p meow-listener` (48 unit tests plus 12 single-request integration tests and the remaining listener integration suites)

## Attribution

Fixes #367

### DIFF
--- a/crates/meow-listener/src/http_proxy.rs
+++ b/crates/meow-listener/src/http_proxy.rs
@@ -308,13 +308,7 @@ async fn handle_http_inner(
         inner.pre_handle_metadata(&mut metadata);
         let Some((proxy, rule_name, rule_payload)) = inner.resolve_proxy_lazy(&mut metadata).await
         else {
-            stream
-                .write_all(
-                    b"HTTP/1.1 502 Bad Gateway\r\n\
-                      Connection: close\r\n\
-                      Content-Length: 0\r\n\r\n",
-                )
-                .await?;
+            write_bad_gateway(stream).await?;
             return Err("no matching rule".into());
         };
 
@@ -388,18 +382,25 @@ async fn handle_http_inner(
                     Ok((up, down)) => {
                         debug!("HTTP single-request relay closed: up={up} down={down}");
                     }
-                    Err(e) => debug!("HTTP single-request relay error: {}", e),
+                    Err(e) => {
+                        debug!("HTTP single-request relay error: {}", e);
+                        // A response head the proxy rejects (oversized,
+                        // malformed, invalid status line) aborts the relay
+                        // before anything reached the client. Turn that into
+                        // an explicit 502 rather than a silent connection
+                        // close; once response bytes have flowed, appending
+                        // an error would corrupt the stream instead.
+                        let sent_response_bytes = client.sent_response_bytes;
+                        drop(client);
+                        if !sent_response_bytes {
+                            let _ = write_bad_gateway(stream).await;
+                        }
+                    }
                 }
             }
             Err(e) => {
                 warn!("{}:{} HTTP dial error: {}", host, port, e);
-                stream
-                    .write_all(
-                        b"HTTP/1.1 502 Bad Gateway\r\n\
-                          Connection: close\r\n\
-                          Content-Length: 0\r\n\r\n",
-                    )
-                    .await?;
+                write_bad_gateway(stream).await?;
             }
         }
         // _guard drops here, removing the entry from Statistics.
@@ -784,7 +785,12 @@ impl<T: AsyncWrite + Unpin> AsyncWrite for ShutdownGate<'_, T> {
         if this.pass_shutdown.load(Ordering::Acquire) {
             Pin::new(&mut this.inner).poll_shutdown(cx)
         } else {
-            Poll::Ready(Ok(()))
+            // Withholding the protocol-level close must not also withhold the
+            // flush that `poll_shutdown` implies. The relay goes straight from
+            // reader EOF to `poll_shutdown` without flushing, so a buffering
+            // transport (TLS, WebSocket) may still hold tail request-body
+            // bytes that would otherwise be silently discarded.
+            Pin::new(&mut this.inner).poll_flush(cx)
         }
     }
 }
@@ -804,6 +810,7 @@ struct SingleRequestClient<'stream, 'buffered, 'gate> {
     response_len: usize,
     response_pos: usize,
     response_state: ResponseWriteState,
+    sent_response_bytes: bool,
     upgrade_requested: bool,
     upgrade_declined: bool,
     pass_upstream_shutdown: &'gate AtomicBool,
@@ -832,6 +839,7 @@ impl<'stream, 'buffered, 'gate> SingleRequestClient<'stream, 'buffered, 'gate> {
             response_len: 0,
             response_pos: 0,
             response_state: ResponseWriteState::Head,
+            sent_response_bytes: false,
             upgrade_requested,
             upgrade_declined: false,
             pass_upstream_shutdown,
@@ -849,7 +857,10 @@ impl<'stream, 'buffered, 'gate> SingleRequestClient<'stream, 'buffered, 'gate> {
                         "write zero bytes to HTTP proxy client",
                     )));
                 }
-                Poll::Ready(Ok(written)) => self.response_pos += written,
+                Poll::Ready(Ok(written)) => {
+                    self.response_pos += written;
+                    self.sent_response_bytes = true;
+                }
                 Poll::Ready(Err(error)) => return Poll::Ready(Err(error)),
                 Poll::Pending => return Poll::Pending,
             }
@@ -892,6 +903,14 @@ impl<'stream, 'buffered, 'gate> SingleRequestClient<'stream, 'buffered, 'gate> {
         let status = parse_response_status(&self.response_buf[..self.response_len])?;
         self.response_pos = 0;
         if (100..200).contains(&status) && status != 101 {
+            // Interim heads previously passed through verbatim, leaking
+            // origin hop-by-hop fields (Connection, Keep-Alive, nominated
+            // headers) that contradict the close semantics of the final head.
+            self.response_len = rewrite_response_head_in_place(
+                &mut self.response_buf,
+                self.response_len,
+                ResponseConnection::Interim,
+            )?;
             self.response_state = ResponseWriteState::InterimPending;
             return Ok(());
         }
@@ -1090,16 +1109,10 @@ impl AsyncRead for SingleRequestClient<'_, '_, '_> {
                     .remaining()
                     .min(this.buffered.len() - this.buffered_pos);
                 let source = &this.buffered[this.buffered_pos..this.buffered_pos + available];
-                let mut consumed = 0;
-                for &byte in source {
-                    if let Err(error) = this.body.consume(byte) {
-                        return Poll::Ready(Err(error));
-                    }
-                    consumed += 1;
-                    if this.body.is_done() {
-                        break;
-                    }
-                }
+                let consumed = match this.body.consume_slice(source) {
+                    Ok(consumed) => consumed,
+                    Err(error) => return Poll::Ready(Err(error)),
+                };
                 destination.put_slice(&source[..consumed]);
                 this.buffered_pos += consumed;
                 continue;
@@ -1116,16 +1129,10 @@ impl AsyncRead for SingleRequestClient<'_, '_, '_> {
                 Poll::Ready(Ok(())) => {
                     let initialized = socket_read.initialized().len();
                     let filled = socket_read.filled().len();
-                    let mut consumed = 0;
-                    for &byte in socket_read.filled() {
-                        if let Err(error) = this.body.consume(byte) {
-                            return Poll::Ready(Err(error));
-                        }
-                        consumed += 1;
-                        if this.body.is_done() {
-                            break;
-                        }
-                    }
+                    let consumed = match this.body.consume_slice(socket_read.filled()) {
+                        Ok(consumed) => consumed,
+                        Err(error) => return Poll::Ready(Err(error)),
+                    };
                     let discarded = filled - consumed;
                     // `take` does not propagate the child ReadBuf's initialized
                     // length back to its parent. Record it before advancing so
@@ -1294,6 +1301,29 @@ impl RequestBodyDecoder {
                 "bytes arrived after the request body boundary",
             )),
         }
+    }
+
+    /// Consume body bytes from `bytes`, stopping at the framing boundary.
+    /// Returns how many bytes belong to the body. Content-Length bodies
+    /// advance in bulk; only chunked framing needs the per-byte state machine.
+    fn consume_slice(&mut self, bytes: &[u8]) -> io::Result<usize> {
+        if let Self::ContentLength(remaining) = self {
+            let consumed = u64::min(*remaining, bytes.len() as u64) as usize;
+            *remaining -= consumed as u64;
+            if *remaining == 0 {
+                *self = Self::Done;
+            }
+            return Ok(consumed);
+        }
+        let mut consumed = 0;
+        for &byte in bytes {
+            self.consume(byte)?;
+            consumed += 1;
+            if self.is_done() {
+                break;
+            }
+        }
+        Ok(consumed)
     }
 }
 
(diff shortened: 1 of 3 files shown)


madeye:
First-pass review of the #370-only delta (diffed against the #369 head, now merged). Overall this looks correct and carefully built; a second adversarial verification pass is running before I merge. Findings so far:

**Verified**

- The `Poll::Pending` returned by `poll_read` when the request body is done relies on both directions sharing one task inside `copy_bidirectional_buf_tracked`, with `poll_shutdown` doing `wake_by_ref()` when the response finishes — that contract holds with the current relay, and the early-413 test pins it.
- `destination.take()` + `destination.advance(consumed)` without `assume_init` is panic-free only because the relay builds `ReadBuf::new` over an initialized stack buffer (crates/meow-tunnel/src/relay.rs:129). Worth a one-line comment at the `advance` site noting that assumption, in case the relay ever moves to `ReadBuf::uninit`.
- Zero-alloc invariant holds: the 8,211-byte response scratch and the drop-lines bitmap live in the async frame / stack; the only per-request heap work is the request rewrite, which is setup, not relay.
- Response head handling: bare-CR/LF rejection, 1xx pass-through (101 deliberately treated as final), in-place hop-by-hop stripping incl. Connection-nominated fields, forced `Connection: close`, and `poll_shutdown` failing on upstream EOF before a complete head — all check out.
- Request side: chunked state machine bounds (size line, extensions, trailers) and the pipelined-byte withholding are right; bytes after the framing boundary are left in the kernel buffer, never forwarded.

**Non-blocking follow-ups**

1. `RequestBodyDecoder::consume` is called per byte even in the bulk states (`ContentLength(n)`, `Data(n)`); a fast path that consumes `min(remaining, len)` at once would take the per-byte match off the plain-HTTP upload path. CONNECT (the common case) is unaffected, so this is a perf nicety, not a gate — but ADR-0006's W-workloads include plain-HTTP POST, so worth doing before the next baseline run.
2. `MAX_RESPONSE_HEADERS = 8192` can break legitimate plain-HTTP origins with cookie/CSP-heavy responses (>8 KB heads were previously forwarded untouched). Consider 16–32 KB — still frame-allocated — or a config knob.
3. `response_connection_nominates` rescans the whole head per header line (O(head²) worst case ~16M byte-ops for a hostile origin). Bounded and fine; noting for the record.

madeye (changes requested):
The adversarial verification pass is done. The framing core, hop-by-hop handling, 1xx path, pipelined-byte withholding, and test coverage all check out (details in my earlier comment), but it surfaced **one blocking defect** plus two decisions that need to be recorded, so I'm requesting changes rather than merging.

## Blocking

**After the request body boundary, the client socket is never read again** (`SingleRequestClient::poll_read` returns `Poll::Pending` once `body.is_done()`), and I've verified the consequence against the relay: `copy_bidirectional_buf_tracked` arms `RELAY_HALF_CLOSE_LINGER` only when one direction reaches `Cycle::Done` (crates/meow-tunnel/src/relay.rs:325-337). Since the upload direction can now never observe the client's FIN, it never completes on its own — the only exit is the upstream ending its response.

- *Leak:* a client sends a complete bodyless `GET http://attacker-chosen-origin/` and closes; the origin accepts and never replies. Pre-PR the relay reaped this 30s after the FIN; post-PR the task, both sockets, and the connection's future are pinned indefinitely. N such connections = unbounded resource growth, and the attacker picks the origin, so "a server that never replies" is attacker-supplied. The same hang affects legitimate traffic whenever an origin ignores `Connection: close`.
- *RST truncation:* bytes past the framing boundary that arrive in the same read are consumed from the kernel queue but not forwarded, and later pipelined bytes stay unread. Dropping the `TcpStream` with unread receive data sends RST, which also discards the proxy's in-flight send buffer — a pooling client (Go `http.Transport`, curl multi) reading a large response slowly can lose the response tail. The existing pipelining test can't see this because its response is 2 bytes.

Suggested shape of the fix (closes both at once): once the body completes, keep polling the client and discard (capped) whatever arrives, recording EOF when it comes; to preserve the deliberate "no early close_notify to encrypted upstreams" property, wrap `remote` so its `poll_shutdown` is a no-op. Then the upload direction reaches `Done`, the linger arms and re-arms on response progress (so a slow origin is never truncated), and the receive queue is empty at close.

## Decisions to record (either fix or state the tradeoff in the PR body)

1. **Future-size growth hits CONNECT too.** The 8,211-byte response scratch lives in `handle_http_inner`'s frame, and the generator is sized to its largest branch — so every HTTPS CONNECT connection (the dominant workload) pays ~8.2 KB more in its per-connection task allocation (~+41 MB at the 5k-connection scale ADR-0011 targets). `Box`ing `SingleRequestClient` would confine the cost to plain-HTTP requests, which already heap-allocate during setup. Either way, CLAUDE.md requires a before/after byte delta (`-Zprint-type-sizes`) in the commit body for footprint-relevant changes — please add one.
2. **ws:// upgrades over the plain-HTTP path are now silently broken.** Stripping `Connection`/nominated fields neuters `Connection: Upgrade` + `Upgrade: websocket`, and a 101 is routed to the final-response path (stripped + `Connection: close`). RFC-correct for a proxy that doesn't do upgrades, and browsers use CONNECT for WebSocket — but it's a capability removal that appears nowhere in the PR body or docs. Please document it (or handle upgrade requests like CONNECT).

## Non-blocking follow-ups

- `destination.advance(consumed)` after `destination.take(...)` without `assume_init` is panic-free only because the relay builds `ReadBuf::new` over an initialized buffer; tokio's own adapters do `assume_init` first. One-line hardening.
- The response side has no equivalent of `validate_connection_options`: an upstream `Connection: Content-Length` gets its framing header dropped, leaving a close-delimited but framing-less response. Contained (connection always closes) but asymmetric; ~3 lines.
- Empty `Connection` list elements (`close,`) are rejected, contrary to RFC 9110 §5.6.1.2 — newly load-bearing on the response side, where one trailing comma from an origin aborts the exchange.
- Response-head scanning is quadratic (`find_crlf_crlf` from offset 0 per `poll_write`; `response_connection_nominates` rescans the head per header line) where the request path deliberately uses an overlap window. Bounded at 8 KB, but easy to fix.
- Test gaps worth adding with the fix: `Expect: 100-continue` where the body is sent *after* the interim response; an HTTP/1.0 client; a response head split across multiple upstream writes.

Happy to re-review promptly once the blocking item is addressed — the rest of the PR is in good shape.

gengjiawen:
Addressed the requested changes in `1e75f1f`:

- continue capped post-boundary reads so client EOF reaches relay `Done` and arms the half-close linger
- wrap the upstream with no-op `poll_shutdown` so EOF does not send early TLS close semantics
- add a 1 MiB slow-response/pipelining regression that catches RST truncation
- box the 9,328 B plain-HTTP adapter only in that branch; measured `handle_http_inner`/`handle_http` at 19,160/19,352 B, identical to current `main`
- explicitly document that plain HTTP Upgrade / `ws://` is unsupported by this one-request fallback and that 101 is closed
- apply the non-blocking hardening and add the suggested `100-continue`, HTTP/1.0, and split-response-head tests

All repository regression gates listed in the updated PR body pass locally. @madeye, please re-review when convenient.

gengjiawen:
Rebased and addressed the remaining Upgrade decision in `223915b`:

- preserve valid bodyless HTTP/1.1 Upgrade requests, including real `ws://` absolute targets
- hold coalesced client protocol bytes until a validated upstream `101`, then switch to an opaque bidirectional tunnel with normal half-close semantics
- when the origin declines Upgrade, return the ordinary final response with close semantics without forwarding held protocol bytes
- keep the 9,344 B plain-HTTP state behind an 8 B `Box`; measured `handle_http_inner` / `handle_http` at 19,176 / 19,368 B versus current main at 19,184 / 19,376 B

The PR body and commit body now record the behavior, tradeoff, and `-Zprint-type-sizes` before/after data. All listed repository gates pass locally, including the new real-TCP success and decline Upgrade regressions. @madeye, please re-review when convenient.

madeye:
Pushed `50b5d95` fixing a resource leak the verification pass found on the new Upgrade path.

**The leak:** while an Upgrade request waited for the origin's 101, `SingleRequestClient::poll_read` returned `Pending` without ever polling the client socket. A client that abandoned the handshake (FIN) was therefore never observed, so the upload relay direction could not reach `Done` and the half-close linger never armed — with a silent origin the connection (task, both sockets, boxed state machine) was pinned indefinitely. Same class as the earlier post-boundary defect, relocated to the upgrade-withhold branch; the origin is attacker-chosen, so it was an unbounded-growth primitive.

**The fix:** during the decision window, early client protocol bytes are consumed into a bounded 64 KiB hold (`MAX_HELD_UPGRADE_BYTES`) rather than left in the kernel receive queue, keeping the client's FIN observable so the linger arms. Held bytes are replayed into the tunnel only after a validated 101; a declined upgrade still drops them unforwarded; exceeding the cap closes the exchange.

Regression test `abandoned_upgrade_is_reaped_by_the_half_close_linger` reproduces the pin without the fix (verified failing) and passes with it. Full regression bar (fmt, three-way clippy, doc, `cargo test --lib`, `cargo test -p meow-listener`) is green; footprint delta is in the commit body (`SingleRequestClient` +32 B, connection futures unchanged).

madeye:
Second verification pass complete (multi-agent review, adversarially verified). Ten findings survived; disposition:

**Fixed in `50b5d95`:**
- Upgrade-withhold branch never polled the client → abandoned handshakes pinned connections indefinitely on silent origins (unbounded-growth primitive). Fixed with a bounded 64 KiB hold that keeps the client FIN observable; regression test verified failing without the fix.

**Fixed in `66f05c9`:**
- `ShutdownGate::poll_shutdown` no-op swallowed the flush the relay relies on (`relay.rs` goes reader-EOF → `poll_shutdown` with `need_flush` still set), silently discarding tail request-body bytes buffered in TLS/WS-based outbound transports. Gated shutdown now flushes without closing.
- Rejected response heads (oversized/malformed/bad status line) closed the client connection with zero bytes; now an explicit `502 Bad Gateway` when nothing was sent yet.
- Interim (1xx) heads were forwarded verbatim, leaking origin hop-by-hop fields (`Connection: keep-alive`, `Keep-Alive`, nominated headers); now stripped like final heads, with nothing appended.
- Quadratic per-line `response_connection_nominates` rescan replaced with a one-pass nominated-options collection.
- Content-Length bodies now advance the decoder in bulk instead of per byte (~1e9 iterations for a 1 GiB POST).
- The triplicated response-side CRLF header walk is consolidated into one `for_each_response_header` helper.

**Recorded as deliberate tradeoffs (not changed):**
- Exceeding the 64 KiB post-boundary discard cap aborts the exchange, which can truncate an in-flight response for a client pipelining >64 KiB mid-response. Parking the upload direction instead would make the client FIN unobservable again and reintroduce the silent-origin pin — the abort is the safer failure mode for a misbehaving client.
- Strict response-head parsing (single-space status line, no obs-fold) intentionally rejects nonconforming origins; with the 502 fix these now fail loudly instead of silently.
- A nonconforming origin replying 1xx to an HTTP/1.0 client is still forwarded; compliant origins never do this.

**Deferred (non-behavioral cleanup):** `upgrade_declined` provably mirrors `upgrade_requested` and could be removed; `poll_flush`/`poll_shutdown` share a verbatim drain block; the chunked decoder duplicates `meow-config`'s `decode_chunked` (different constraints, likely fine as is).

Verified refuted (no action): anytls/h2 stream leak via swallowed shutdown (Drop impls emit fin/RST_STREAM), response-side framing smuggling (forced close semantics leave no reusable connection to poison), `RESPONSE_BUFFER_SIZE` arithmetic, `drop_lines` bounds, `ReadBuf::take`/`assume_init` soundness, chunk-size overflow, request-side CL/TE smuggling.

Full regression bar green locally at `66f05c9`; 15 integration + 22 unit tests in the listener crate.

madeye (approved):

madeye (approved):
All review findings addressed: the original blocking defect, the upgrade-abandonment leak, and the second-pass findings (gated-shutdown flush loss, silent-close→502, interim hop-by-hop leak, quadratic rescan, per-byte CL decode). Full regression bar and CI green at 66f05c9.

Issue #363, opened 2026-07-28
The built-in dns in tun mode doesn't work on Windows platform


xmh0511:
As the drive-by fixing in https://github.com/madeye/meow-rs/pull/360, the route_manager has successfully set the route `198.18.0.0/16` to the created `tun`
````
 45...........................tun0 Tunnel  

0.0.0.0          0.0.0.0      192.168.1.1    192.168.1.114     35                                                                 
127.0.0.0        255.0.0.0            在链路上         127.0.0.1    331                                                             
127.0.0.1  255.255.255.255            在链路上         127.0.0.1    331                                                       
127.255.255.255  255.255.255.255            在链路上         127.0.0.1    331                                                           
172.19.0.0  255.255.255.252            在链路上        172.19.0.1    261                                                            
172.19.0.1  255.255.255.255            在链路上        172.19.0.1    261                                                            
172.19.0.3  255.255.255.255            在链路上        172.19.0.1    261                                                          
172.24.208.0    255.255.240.0            在链路上      172.24.208.1   5256                                                          
172.24.208.1  255.255.255.255            在链路上      172.24.208.1   5256                                                        
172.24.223.255  255.255.255.255            在链路上      172.24.208.1   5256                                                           
192.168.1.0    255.255.255.0            在链路上     192.168.1.114    291                                                         
192.168.1.114  255.255.255.255            在链路上     192.168.1.114    291                                                         
192.168.1.255  255.255.255.255            在链路上     192.168.1.114    291                                                         
192.168.147.0    255.255.255.0            在链路上     192.168.147.1    291                                                         
192.168.147.1  255.255.255.255            在链路上     192.168.147.1    291                                                       
192.168.147.255  255.255.255.255            在链路上     192.168.147.1    291                                                         
192.168.217.0    255.255.255.0            在链路上     192.168.217.1    291                                                         
192.168.217.1  255.255.255.255            在链路上     192.168.217.1    291                                                       
192.168.217.255  255.255.255.255            在链路上     192.168.217.1    291                                                            
198.18.0.0      255.255.0.0            在链路上        172.19.0.1    261                                                        
198.18.255.255  255.255.255.255            在链路上        172.19.0.1    261     
````

The expected behavior is that the request to a domain will be assigned a fake IP within the `198.18.0.0/16`, and the traffic will be routed to the created `tun`. However, this doesn't work. 

By manually looking at the DNS resolution
````
[...]

PR #374, opened 2026-07-29
Fix/tun dns os resolver on the Windows platform

xmh0511:
## Summary

Migrate the Windows TUN-mode DNS implementation from "system DNS → TUN fake-IP gateway (198.18.0.1)" to "local UDP DNS server + system DNS → loopback", adopting the same architecture used by seeker.

## Why: Three obstacles that made the old approach unworkable on Windows

The original design pointed the system DNS resolver at the TUN device's fake-IP gateway (198.18.0.1). DNS queries would be routed through the TUN userspace IP stack, intercepted by `dns-hijack`, and answered with fake IPs. This works on macOS and Linux, but Windows threw up three blockers that could not be resolved:

### 1. IPv6 DNS servers are unremovable

The Windows DNS Client **prefers IPv6 DNS servers** when any are present. Virtualization platforms (WSL, Docker Desktop) persistently inject `fec0:0:0:ffff::*` as IPv6 DNS servers on their virtual adapters. Attempting to clear them via `netsh` or `Remove-DnsClientServerAddress` is futile — the addresses are re-injected within seconds by the virtualization platform. As a result, most DNS queries bypass the TUN entirely and go directly to these IPv6 resolvers.

### 2. Setting IPv6 DNS to ::1 causes ECONNRESET

We first tried blocking all IPv6 DNS traffic via Windows Firewall (outbound UDP 53 for `fe80::*`, `fec0::*`, and all IPv6 addresses). The user reported no observable effect.

We then tried setting the IPv6 DNS to `::1`, expecting the Windows DNS Client to send queries to loopback. Instead, it returned `ECONNRESET` — the Windows DNS Client implementation does **not** use standard UDP sockets for DNS queries. It sends them through an internal NDIS path, and directing it to `::1` causes an immediate connection reset because no process is listening on a UDP socket at `[::1]:53`.

### 3. Firewall rules had no effect

Outbound Windows Firewall rules blocking IPv6 UDP 53 traffic were added and confirmed active. The DNS Client's queries continued to bypass them — the client appears to use a lower-level path that does not traverse the Windows Filtering Platform in the same way regular socket traffic does.

## What: The new approach

Abandon the "DNS through TUN userspace IP stack" strategy and adopt seeker's approach: run a real UDP socket DNS server on loopback, and point system DNS at it.

```
                 ┌──────────────────────────┐
                 │   Windows DNS Client      │
                 │  (system resolver)        │
                 └────────────┬──────────────┘
                              │ DNS query
                              ▼
    ┌─────────────────────────────────────────┐
    │  local_dns (real UDP sockets)            │
    │  127.0.0.1:53 / [::1]:53                │
    │  → DnsServer::handle_query()             │
    │  → returns fake IP                       │
    └─────────────────────────────────────────┘
```

Key files:

| File | Change |
|------|--------|
| `crates/meow-listener/src/tun/local_dns.rs` (NEW) | Binds real UDP sockets on `127.0.0.1:53` and `[::1]:53`, answers queries via `DnsServer::handle_query` — the same pipeline used by TUN `dns-hijack` |
| `crates/meow-listener/src/tun/dns.rs` | `DnsGuard::setup()` now sets system DNS to `127.0.0.1` / `::1` instead of the fake-IP gateway; `drop()` adds a safety net that resets all adapters to DHCP before restoring from backup; per-adapter logging at info level |
| `crates/meow-listener/src/tun/mod.rs` | Spawns `local_dns::run()` on startup when `dns-hijack` is active; replaces bare `oneshot::Sender<()>` with `TunReady` enum; adds per-stage startup timing logs |

### TunReady enum

Replaced `oneshot::Sender<()>` with `oneshot::Sender<TunReady>`, where `TunReady` has two variants: `Ready` and `Failed(String)`. A `ReadyNotifier` RAII wrapper ensures that any `?` early-return in `run()` immediately sends `Failed(msg)` to the caller, eliminating the 30-second startup timeout that was previously needed to detect setup failures.

### Startup timing logs

Each stage now logs its duration at info level for easy profiling:

```
info TUN listener 'tun' starting...
info creating TUN device...
info TUN device 'meow-tun' created in 85ms
info auto-route installed in 320ms
info dns-guard setup took 450ms (active: true)
info netstack built in 0ms
info TUN listener 'tun' started on device 'meow-tun' (198.18.0.1/16, mtu 1500, ... total startup 920ms)
```

### DNS restore reliability

`DnsGuard::drop()` now first calls `reset_all_dns()` to reset all adapters (IPv4 + IPv6) to DHCP, then restores from the backup. This belt-and-suspenders approach ensures that `127.0.0.1` / `::1` never survive a TUN shutdown, even if the backup data is incomplete.

## Testing

- [x] `cargo check --workspace` passes
- [x] `cargo clippy --workspace` zero warnings
- [x] All unit tests pass
- [x] TUN mode on Windows: `Resolve-DnsName` returns fake IPs
- [x] TUN mode off: DNS correctly restored, `Resolve-DnsName` works normally


xmh0511:
@madeye Please review this PR; it makes tproxy work on Windows.

xmh0511:
This PR fixes https://github.com/madeye/meow-rs/issues/363

xmh0511:
@madeye Can we make progress on this PR?

madeye:
I pushed 1f2f06b to this branch fixing the four review blockers:

1. **macOS TUN startup regression** — the retry loop forced a device name (default `meow-tun`), but tun-rs on macOS rejects any name not starting with `utun`, so every start with an unset `tun.device` failed after 5 attempts. macOS now passes the configured name through unchanged (platform picks `utunN` when unset); the `meow-tun` default + suffix rotation stays on Windows/Linux where stale-adapter recovery matters.
2. **DNS backup poisoning after an unclean shutdown** — if a run was SIGKILLed/crashed, the next run's `DnsGuard` backed up the leftover `127.0.0.1`/`::1`/fake-gateway values and a later clean exit "restored" them, leaving machine-wide DNS dead. The backup now filters meow's own values on all three platforms; Linux additionally writes a sidecar (`/etc/resolv.conf.meow-backup`) before the first mutation and recovers the true original from it when it finds its own marker in `/etc/resolv.conf`.
3. **Swallowed port-53 bind failure (Windows)** — `127.0.0.1:53`/`[::1]:53` are now bound *before* DnsGuard repoints system DNS at loopback, and a bind failure (ICS, Docker, another resolver holding :53) fails startup via `TunReady::Failed` instead of silently leaving all system DNS aimed at a dead address while reporting Ready.
4. **Clippy gate** — `io_other_error` and `cloned_ref_to_slice_refs` in the macOS backend broke `cargo clippy -- -D warnings` (CI-enforced).

Verified locally: three-way clippy (default / no-default / all-features) clean, `cargo doc -D warnings` clean, `cargo test --lib` all green, and meow-listener cross-compiles for `x86_64-unknown-linux-gnu` and `x86_64-pc-windows-gnu` with `listener-tun`. The Windows paths compile but I couldn't runtime-test them here — a smoke test on a real Windows box (especially the port-53-occupied case and a kill -9 → restart → clean-exit cycle) would be appreciated.

Non-blocking review findings still open (can be follow-ups): the 30s→600s `TUN_STARTUP_TIMEOUT` is awaited while holding `config_mutation_lock` (one hung start blocks all config mutations for 10 min); leftover `info!` debug logging + eager hex dumps on the shared DNS hot path (logs every queried domain at default verbosity); `DnsGuard::drop` runs blocking PowerShell on a tokio worker; Windows restore pins DHCP-assigned DNS as static; `local_dns` bypasses the hardened `DnsServer::run` loop (unbounded spawn-per-packet); and the hand-rolled response builders can emit QDCOUNT/body mismatches on multi-question or malformed queries.

crates/meow-api/src/lib.rs
```diff
@@ -22,7 +22,7 @@ use tracing::{info, warn};
 /// config-reload path (`routes.rs::spawn_tun_from_raw`) so the two don't
 /// drift. Windows wintun adapter creation plus smoltcp netstack init can
 /// take tens of seconds on slow machines; 5 s proved too aggressive.
-pub const TUN_STARTUP_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(30);
+pub const TUN_STARTUP_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(600);
```
madeye: **Timeout raised 30s → 600s while a lock is held.** `swap_config_and_reconcile_tun` (routes.rs) holds `config_mutation_lock` across the whole `spawn_tun_from_raw().await`, so a genuinely hung TUN start (stuck wintun creation or PowerShell call — `spawn_blocking` is not cancellable) now blocks the triggering `PUT /configs` **and every other config-mutation endpoint** for up to 10 minutes; app startup similarly waits inline before "meow-rs is running". The new `TunReady::Failed` fast-path only covers failures that *return* — the timeout exists precisely for hangs, which is what the old 30 s bound was sized for. Suggest keeping the timeout at 30–60 s, or awaiting startup outside the lock. The doc comment above still says "5 s proved too aggressive", justifying the old 30 s value — it should be updated to explain why 600 s is needed if it stays.

crates/meow-dns/src/server.rs
```diff
@@ -89,23 +104,36 @@ impl DnsServer {
...
-        let (domain, qtype, _offset) = Self::parse_question(&data[12..])?;
-        debug!("DNS query: {} type={}", domain, qtype);
+        let (domain, qtype, _offset) = Self::parse_question(&data[12..]).map_err(|e| {
+            info!(
+                "DNS query parse_question failed: {e} | bytes: {}",
+                hex_prefix(&data[12..], 64)
+            );
+            e
+        })?;
+        info!(
```
madeye: **Leftover debug logging at `info!` on the shared DNS hot path.** `handle_query` serves the main UDP DNS server (`:1053`), the TUN dns-hijack path, and the new loopback server — with the default `info` filter this emits one multi-field line per DNS query on the whole box (log flood), and records every queried domain at default verbosity (a privacy regression vs. the previous `debug!`). The parse-failure hex dump at `info!` a few lines up has the same issue. These should go back to `debug!`.

crates/meow-listener/src/tun/udp.rs
```diff
@@ -73,12 +73,22 @@ pub(super) async fn run_udp(
         if dns_hijack && dst.port() == 53 {
             let resolver = Arc::clone(tunnel.resolver());
             let reply_tx = reply_tx.clone();
+            let qhex = hex_prefix(&data, 48);
```
madeye: **Eager hex-dump allocations on the dns-hijack hot path.** `let qhex = hex_prefix(&data, 48)` runs before the `info!` macro, so ~48 `format!` allocations + a `Vec<String>` + `join` are paid for **every** hijacked DNS packet (and again for the reply at line 84) even when info logging is filtered out — this codebase has zero-alloc hot-path invariants (ADR-0008) and an allocation-focused lint set (ADR-0010 Addendum A). Move the computation inside a `tracing::enabled!`/`debug!` guard or drop it; `hex_prefix` is also duplicated byte-for-byte in `meow-dns/src/server.rs`.

crates/meow-listener/src/tun/dns.rs
```diff
@@ -0,0 +1,568 @@
...
+        #[cfg(not(any(target_os = "windows", target_os = "macos", target_os = "linux")))]
+        {
+            let _ = dns_addr;
+            debug!("tun dns-guard: no DNS backend for this platform; skipping");
+            None
+        }
+    }
+}
+
+impl Drop for DnsGuard {
```
madeye, line 110: **`DnsGuard::drop` runs blocking process spawns on a tokio worker thread.** When `stop_tun`/config-reload aborts the TUN task, this `Drop` executes on the runtime worker doing the abort: `reset_all_dns` (2 PowerShell spawns) + `restore` (one spawn per backed-up adapter) + `Clear-DnsClientCache`, each commonly 100 ms–seconds on Windows — a multi-second stall of unrelated async tasks on every TUN teardown. The setup path wraps these same calls in `spawn_blocking` for exactly this reason. Consider moving restore into an explicit async teardown (spawn_blocking) with the `Drop` impl as a last-resort fallback.

crates/meow-listener/src/tun/dns.rs
```diff
@@ -0,0 +1,568 @@
...
+        restore_section(v6_section, &mut had_error);
+
+        if had_error {
+            Err(std::io::Error::other("one or more DNS restores failed"))
+        } else {
+            Ok(())
+        }
+    }
+
+    fn restore_section(section: &str, had_error: &mut bool) {
```
madeye: **Restore converts DHCP-assigned DNS to static.** `Get-DnsClientServerAddress` reports *effective* servers with no DHCP/static distinction, and `restore_section` re-applies them via `Set-DnsClientServerAddress -ServerAddresses`, so an adapter that got its DNS from DHCP ends up statically pinned to a snapshot of those values (including virtualization-injected `fec0::` entries). A laptop that later joins a different network keeps the old network's DNS servers. Fixing this properly means backing up only *statically configured* servers (e.g. the `NameServer` registry value under `Tcpip{,6}\Parameters\Interfaces\<guid>`) and letting everything else fall to the `-ResetServerAddresses` safety net.

crates/meow-listener/src/tun/local_dns.rs
```diff
@@ -0,0 +1,103 @@
...
+    loop {
+        match socket.recv_from(&mut buf).await {
+            Ok((len, peer)) => {
+                if len == 0 {
+                    continue;
+                }
+                let query = buf[..len].to_vec();
+                let sock = Arc::clone(&socket);
+                let res = Arc::clone(&resolver);
+                tokio::spawn(async move {
```
madeye: **Unbounded `tokio::spawn` per packet, bypassing the hardened serve loop.** `meow-dns`'s `DnsServer::run` was just hardened (bounded worker pool with `try_send` backpressure, panic-guarded workers — commit 1ab923c) and already implements exactly this socket→`handle_query`→`send_to` loop for an arbitrary bind address. This re-implementation spawns one task per received datagram with no bound — a UDP flood on loopback:53 (any local process can send there) grows tasks without limit — and future hardening of the meow-dns loop won't apply here. Reusing `DnsServer` bound to `127.0.0.1:53`/`[::1]:53` (or extracting a shared serve fn) would keep one hardened implementation.

crates/meow-dns/src/server.rs
```diff
@@ -254,32 +292,77 @@ impl DnsServer {
         Ok((domain, qtype, pos))
     }
 
+    /// Copy the full question section (all `qdcount` entries) from `query`
+    /// into `buf`. Returns the number of bytes copied.
+    fn copy_question_section(buf: &mut Vec<u8>, query: &[u8], qdcount: u16) -> usize {
```
madeye: **Header counts can contradict the response body.** The builders echo the query's raw QDCOUNT while `copy_question_section` may copy fewer questions: truncated packets break out of the loop early, and the label walk treats a compression-pointer byte (`>= 0xC0`) as a raw length (`pos += 1 + query[pos]` skips ~193 bytes). Multi-question queries (wire-legal) also get `ANCOUNT=1` answering only the first question while the header claims all of them. No panic — bounds are checked — but this is a network-facing parser emitting wire-invalid responses, in a repo that just shipped DNS-hardening commits (#368, #369). The hickory `Message` API already used in `handle_generic_forward` would get this right; alternatively reject `qdcount != 1` up front (RFC-compliant for most resolvers).

crates/meow-listener/src/tun/mod.rs
```diff
@@ -112,15 +115,78 @@ pub struct TunListenerConfig {
...
+            tracing::warn!("ReadyNotifier::ready called but tx was already None");
+        }
+    }
+}
+
+impl Drop for ReadyNotifier {
+    fn drop(&mut self) {
+        if let Some(tx) = self.tx.take() {
+            let _ = tx.send(TunReady::Failed(
+                "listener setup failed before reaching readiness".into(),
```
madeye, line 169: Minor: `ReadyNotifier`'s `Drop` sends a fixed generic string, so the *actual* setup error (`run()`'s `Err`) never reaches the readiness channel — callers log "listener setup failed before reaching readiness" instead of the real cause, which contradicts the `TunReady::Failed` doc ("carries the underlying error message"). Consider a `fail(msg)` method called from an explicit error path, or logging the real error at the `run()` call sites that own the JoinHandle.

crates/meow-api/src/routes.rs
```diff
@@ -1686,11 +1699,28 @@ async fn swap_config_and_reconcile_tun(state: &AppState, candidate: RawConfig) {
...
-        if let Some(handle) = spawn_tun_from_raw(&state.tunnel, &snapshot).await {
-            state.tunnel.set_tun_handle(handle).await;
-            info!("TUN listener started via config reload");
+        // off → on
+        match spawn_tun_from_raw(&state.tunnel, &snapshot).await {
+            Ok(Some(handle)) => {
+                state.tunnel.set_tun_handle(handle).await;
+                info!("TUN listener started via config reload");
+            }
+            Ok(None) => {
```
madeye, line 1729: Minor: on a build **without** `listener-tun`, `spawn_tun_from_raw` returns `Ok(None)` even when `tun.enable=true`, so this arm treats it as success and the stored config keeps `tun.enable=true` with nothing running — the exact state the new rollback was added to prevent. The non-feature stub could return `Err("this build lacks the 'listener-tun' feature")` when `enable` is set, so the rollback path also covers feature-less builds.

madeye (commented):
Review of this branch (multi-agent review + manual verification). The four merge blockers found earlier — the macOS forced device name breaking TUN startup, DNS backup poisoning after an unclean shutdown, the swallowed port-53 bind failure, and the two clippy `-D warnings` errors — are already fixed in 1f2f06b, which I pushed to this branch. The inline comments below are the remaining non-blocking findings; they can be follow-ups but are worth addressing before or shortly after merge.

crates/meow-api/src/lib.rs
xmh0511: @madeye  In my test, I found that setting DNS by command takes much longer on my Windows platform. Fairly, we may reduce this time to `300` seconds?

madeye:
**Runtime verification of 1f2f06b on real Windows** (ARM64 Windows 11 VM, elevated, binary cross-built with `minimal,listener-tun`):

| Test | Result |
|---|---|
| Config test (`-t`) + TUN startup: wintun device on attempt 1, auto-route, dns-guard active, local-dns on `127.0.0.1:53` + `[::1]:53`, readiness signalled | ✅ |
| Function: `Resolve-DnsName example.com` → fake IP `198.18.0.5`; HTTP through the TUN → 200 | ✅ |
| Graceful teardown (config reload `tun.enable=false`): DNS restored, no loopback leftovers | ✅ |
| **Poisoning fix, real-world case**: the VM's Ethernet adapter *already* had a leftover `127.0.0.1` from earlier testing of the unfixed build — startup logged `adapter 'Ethernet' still points at 127.0.0.1 (leftover from an unclean shutdown?) — excluding it from the backup` and teardown reset it to DHCP (`192.168.1.1,192.168.0.1`), repairing the machine | ✅ |
| **Crash cycle**: `taskkill /F` with TUN active → DNS left at `127.0.0.1,::1` → restart detected and excluded all 3 poisoned entries (v4 + v6 + loopback pseudo-interface) → clean exit → DNS back to DHCP + original IPv6, resolution works | ✅ |
| **Port 53 occupied**: with another process holding `127.0.0.1:53`, enabling TUN fails immediately with `loopback DNS server startup failed: bind 127.0.0.1:53 … (os error 10048)`, system DNS is never touched, and the API rolls `tun.enable` back to `false` | ✅ |

One pre-existing issue surfaced during testing (not introduced by this PR, worth its own issue): right after TUN start, Windows NetBIOS name-service broadcasts to the TUN subnet's broadcast address (`172.19.0.3:137` for the default `172.19.0.1/30`) are captured and relayed DIRECT, and the relayed packet re-enters the TUN — a UDP amplification loop that logged ~16k flows in a minute and transiently exhausted ephemeral ports (`WSAENOBUFS 10055`, DNS lookups fail until NetBIOS backs off). The UDP path should drop packets destined to the TUN's own subnet/broadcast/multicast before relaying.

xmh0511:
@madeye except for the set value of the timeout issue, please fix the other issues you found in your review, thanks.

madeye (approved):

crates/meow-api/src/lib.rs
madeye: yep, sounds good.

madeye:
@xmh0511 Done — all remaining review findings are fixed in 0d86e91, and per your request the timeout value is left at 600s (9f07f92 reverts my 120s change; I only kept a doc note that the config-reload path holds `config_mutation_lock` across the wait).

What 0d86e91 covers:

- **Routing loop (pre-existing, found during runtime testing)**: traffic to the TUN's own subnet/broadcast/multicast re-entered the device — Windows NetBIOS broadcasts to `172.19.0.3:137` spawned ~16k flows/min at startup and transiently killed DNS via ephemeral-port exhaustion (WSAENOBUFS). Both UDP and TCP paths now drop such destinations (`is_looping_dst`).
- **Hot-path logging**: per-query `info!` lines back to `debug!`, hex dumps moved inside the log macros (zero cost when filtered), `hex_prefix` deduplicated into meow-dns.
- **DHCP→static restore**: the Windows backup now reads only statically configured registry `NameServer` values (`Tcpip{,6}`), so DHCP adapters are never pinned to a snapshot — they fall to the reset-to-DHCP pass.
- **Drop blocking**: teardown batched into two PowerShell invocations (reset-all + one combined restore script) instead of 3+N sequential spawns; kept synchronous in `Drop` deliberately (`stop_tun` awaits the aborted task, which serialises reload disable→enable against the restore — documented inline).
- **local_dns**: now runs through the extracted `DnsServer::serve` (bounded worker pool + backpressure) instead of unbounded spawn-per-packet.
- **DNS wire format**: `qdcount != 1` → header-only FORMERR; builders copy the single validated question by its parsed length (no more compression-pointer misparse or QDCOUNT/body mismatch).
- **Error propagation**: `TunReady::Failed` now carries the real setup error (`ReadyNotifier::fail`), and non-`listener-tun` builds return `Err` on `tun.enable=true` so the config rollback covers them too.
- **Windows CI**: the windows job's test leg used default features, which exclude `listener-tun` — none of the tun code ever ran on CI. Added a `cargo test -p meow-listener --features listener-tun --lib` leg to the windows and macos jobs, with new unit tests for the loop guard, device-name selection (macOS regression), backup poisoning filter, batched restore-script generation, and a read-only smoke test of the real PowerShell/registry backup path on the Windows runner.

Re-verified on the ARM64 Windows 11 VM: startup log went from ~16k lines/min to 29 total (zero loop flows), DNS + HTTP work immediately after start, a static `8.8.8.8,1.1.1.1` config round-trips through backup/restore via the registry path, and the crash → restart → clean-exit cycle still recovers to DHCP.

xmh0511:
@madeye Please resolve the branch confliction issue, thanks.

crates/meow-api/src/lib.rs
madeye: Keeping 600s per the author's preference (slow first-time wintun driver installs) — see 9f07f92, which documents the lock-holding trade-off in the constant's doc comment. Leaving this thread open as a tracked follow-up: moving the await outside `config_mutation_lock` (or splitting the lock) would remove the worst-case 10-minute API stall without touching the timeout.

madeye: Done in 648466d — `TUN_STARTUP_TIMEOUT` is now 300s as proposed. That halves the worst-case `config_mutation_lock` hold for a hung startup while keeping the headroom your Windows testing showed the PowerShell DNS commands need. Resolving this thread; moving the await outside the lock entirely can be a separate follow-up if the stall ever bites in practice.

madeye (approved):
All review findings addressed (see resolved threads), CI green across all 8 jobs including the new Windows/macOS listener-tun test legs, and the Windows behavior runtime-verified on an ARM64 Windows 11 VM (startup, fake-IP DNS + HTTP through the TUN, DNS restore on teardown, crash-recovery, port-53-occupied failure path).

PR #376, opened 2026-07-31
Feat/xhttp support

navysummer:
# feat: add XHTTP (splithttp) protocol support

Implement XHTTP outbound proxy protocol — an HTTP/2-based transport from Xray-core, also supported by mihomo and sing-box. Uses `stream-one` mode (single bidirectional HTTP/2 POST stream) for tunnelling proxy traffic.

### New files
- **`crates/meow-transport/src/xhttp.rs`** — `XhttpLayer: Transport` implementing stream-one mode over h2 with configurable path, host, and custom headers
- **`crates/meow-proxy/src/xhttp_adapter.rs`** — `XhttpAdapter: ProxyAdapter` with TCP dial, `ProxyHealth` integration, and 6 unit tests
- **`crates/meow-proxy/tests/xhttp_integration.rs`** — 8 integration tests using embedded h2 mock server
- **`docs/specs/proxy-xhttp.md`** — spec doc with YAML schema, wire format, error handling, and test plan

### Modified files
- **`meow-common/src/adapter_type.rs`** — add `Xhttp` variant to `AdapterType` and `ConnType`
- **`meow-transport/src/error.rs`** — add `Xhttp(String)` error variant
- **`meow-transport/src/lib.rs`** — expose `xhttp` module
- **`meow-proxy/src/lib.rs`** — export `XhttpAdapter`, update cfg gate
- **`meow-config/src/proxy_parser.rs`** — add `parse_xhttp()` with 4 config parsing tests
- **Cargo.toml** files in `meow-transport`, `meow-proxy`, `meow-config`, `meow-app` — add `xhttp` feature flag, included in `full` profile

## YAML config example

```yaml
proxies:
  - name: my-xhttp
    type: xhttp
    server: example.com
    port: 443
    path: /xhttp
    host: my.example.com
    headers:
      X-Custom: value

### CODE REVIEW

File: crates/meow-config/src/proxy_parser.rs [hunk 1]
```diff
@@ -1877,6 +1884,55 @@ fn parse_relay_group(
...
+///
+/// Required fields: `server`, `port`. Optional: `path`, `host`, `headers`, `udp`.
+///
+/// XHTTP is an HTTP/2-based transport protocol that tunnels proxy traffic over
+/// HTTP/2 streams. It supports three modes (stream-one, stream-up, packet-up)
+/// but only stream-one is implemented in this initial version.
+///
+/// upstream: mihomo transport/xhttp
+#[cfg(feature = "xhttp")]
+fn parse_xhttp(
```
madeye, line 1897: **[security — confirmed]** `parse_xhttp` silently drops `tls`/`sni`/`skip-cert-verify`/`mode`, and `XhttpAdapter` has no TLS composition path — it always dials cleartext HTTP/2.

A standard mihomo/xray xhttp config (`tls: true`, `sni: …`, port 443) parses with no error or warning; `xhttp_adapter.rs` dials raw TCP into `XhttpLayer`, which hardcodes an `http://` URI and does a cleartext h2c handshake. Proxy traffic goes out unencrypted or fails opaquely against a TLS-only server — exactly ADR-0002's worked Class-A example (user expects TLS, gets plaintext), and it's not in the PR's own divergence table. Unlike `trojan.rs`/`vless_adapter.rs`, which compose a `TlsLayer` from config, there is no TLS-capable code path at all. At minimum these keys should hard-error until TLS is wired up.

crates/meow-transport/src/xhttp.rs
```diff
@@ -0,0 +1,258 @@
...
+
+impl XhttpLayer {
+    pub fn new(config: XhttpConfig) -> Self {
+        Self { config }
+    }
+}
+
+#[async_trait]
+impl Transport for XhttpLayer {
+    async fn connect(&self, inner: Box<dyn Stream>) -> Result<Box<dyn Stream>> {
```
madeye, line 51: **[correctness — confirmed]** The target address is never encoded anywhere in the wire format. `Transport::connect` never receives `Metadata`; `dial_tcp` only logs `metadata.remote_address()` via `debug!` and discards it, so every destination produces an identical POST to the same fixed server/path.

Real xhttp servers (xray-core) expect an inner protocol (VLESS/Trojan) to carry the destination — this implementation cannot tell any real server where to forward, so it only interoperates with a bespoke test server despite shipping under a name that implies mihomo/xray wire compatibility. This needs an architectural decision (xhttp as a transport under VLESS, like xray models it) rather than a patch.

crates/meow-transport/src/xhttp.rs
```diff
@@ -0,0 +1,258 @@
...
+    ) -> Poll<io::Result<usize>> {
+        let this = self.get_mut();
+
+        if this.pending_write.is_none() {
+            let data = Bytes::copy_from_slice(buf);
+            this.send.reserve_capacity(data.len());
+            this.pending_write = Some(data);
+        }
+
+        match this.send.poll_capacity(cx) {
```
madeye, line 178: **[correctness — confirmed]** `poll_write` discards the capacity actually granted by `poll_capacity` and always `send_data`s the full buffered chunk.

`reserve_capacity(data.len())` is called, but once `poll_capacity` returns *any* nonzero (possibly much smaller) capacity, the code ignores it and sends the entire buffer. h2 0.4.15's `SendStream` docs (share.rs:326-331) state this buffering is unbounded. Under a slow/congested peer with a small send window, every relay write buffers the excess in memory without bound — same failure class as the previously-fixed gRPC unbounded-growth DoS. The fix is to send at most the granted capacity and keep re-polling (note `h2.rs` has the identical bug — see the duplication comment).

crates/meow-transport/src/xhttp.rs
```diff
@@ -0,0 +1,258 @@
...
+
+impl AsyncWrite for XhttpStream {
+    fn poll_write(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &[u8],
+    ) -> Poll<io::Result<usize>> {
+        let this = self.get_mut();
+
+        if this.pending_write.is_none() {
```
madeye, line 172: **[correctness — plausible]** `pending_write` can survive a cancelled write future and send stale bytes.

If a caller's `write(buf1)` future is dropped while `poll_capacity` is `Pending` (e.g. a `timeout`/`select!` around a write), `pending_write` stays `Some(buf1)`. A later `write(buf2)` sees `pending_write` occupied, skips copying `buf2`, sends the stale `buf1` bytes once capacity arrives, and returns `Ready(Ok(buf2.len()))` — the caller believes `buf2` was transmitted. Silent AsyncWrite-contract violation / byte-stream corruption. Worth a targeted test with a timed-out write.

madeye (commented):
Automated deep review (8 analysis angles, findings cross-verified against the worktree — fmt/clippy gates actually run, h2 0.4.15 source checked, call sites traced).

**TL;DR: not mergeable as-is.** Two architectural problems, two correctness bugs in the h2 stream bridge, a per-dial resource leak, and the branch currently fails the repo's own Regression Bar (`cargo fmt --check` + `clippy -D warnings`).

**Most severe:**
1. **No TLS path, config keys silently dropped** — `tls`/`sni`/`skip-cert-verify`/`mode` parse without error but the adapter always dials cleartext h2c (ADR-0002 Class A).
2. **Destination never encoded** — `Metadata` is discarded; every dial sends an identical POST to the fixed server/path, so this cannot interoperate with real xray/mihomo xhttp servers.
3. **Unbounded h2 buffering** — `poll_write` ignores the capacity actually granted and `send_data`s the whole chunk (h2 docs: unbounded internal buffering) — same class as the previously-fixed gRPC DoS.
4. **Stale `pending_write` after a cancelled write future** can send old bytes while reporting the new buffer as written (silent stream corruption).
5. **Socket/task leak per dial** — the detached h2 connection driver is never shut down; compounds on every health-probe cycle.

Most of the stream-bridge issues come from copy-pasting `H2Stream`/`H2Layer::connect` instead of reusing them — factoring a shared type would fix several findings at once. Inline comments below have details.

crates/meow-transport/src/xhttp.rs
```diff
@@ -0,0 +1,258 @@
...
+
+        let request = req_builder
+            .body(())
+            .map_err(|e| TransportError::Config(format!("xhttp: invalid request config: {e}")))?;
+
+        let (mut h2, conn) = h2::client::handshake(inner)
+            .await
+            .map_err(|e| TransportError::H2(e.to_string()))?;
+
+        tokio::spawn(async move {
```
madeye, line 84: **[resource-leak — confirmed]** The detached h2 connection driver (`tokio::spawn(async move { let _ = conn.await; })`) and its TCP socket are never explicitly shut down — dropping `XhttpStream` only closes the individual h2 stream (implicit RST_STREAM), not the connection (no `graceful_shutdown`).

This also fires on the new non-200 error path, where the driver keeps owning the live socket. `health.rs` calls `dial_tcp` periodically for url-test/fallback probing of every proxy (default 300s, plus REST `/proxies/delay`), so each probe cycle leaks one TCP socket + one tokio task, bounded only by the remote server's idle timeout.

crates/meow-transport/src/xhttp.rs
```diff
@@ -0,0 +1,258 @@
...
+        Poll::Ready(Ok(()))
+    }
+}
+
+impl Unpin for XhttpStream {}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use tokio::io::{AsyncReadExt, AsyncWriteExt};
```
madeye, line 217: **[ci-gate — reproduced]** `cargo clippy -p meow-transport --all-targets --features xhttp -- -D warnings` fails here: `unused imports: AsyncReadExt and AsyncWriteExt`. This blocks the Regression Bar / CI clippy gate for any build enabling the `xhttp` feature.

crates/meow-proxy/tests/xhttp_integration.rs
```diff
@@ -0,0 +1,283 @@
...
+        let mut send = respond.send_response(resp, false).unwrap();
+        send.send_data(Bytes::new(), true).unwrap();
+        // Keep the connection alive until the client disconnects.
+        // connection.next() returns None when the client closes.
+        let _ = tokio::time::timeout(Duration::from_secs(5), connection.next()).await;
+    });
+    (addr, handle)
+}
+
+fn metadata_for(target: SocketAddr) -> Metadata {
```
madeye, line 67: **[ci-gate — reproduced]** `cargo clippy -p meow-proxy --all-targets --features xhttp -- -D warnings` fails: `function metadata_for is never used`. All 8 tests build an inline `Metadata` literal instead of calling this helper — prefer actually calling it at the 8 call sites (kills the warning *and* the duplication).

crates/meow-config/src/proxy_parser.rs
```diff
@@ -13,6 +13,8 @@ use meow_proxy::{
 };
 #[cfg(feature = "vless")]
 use meow_proxy::{TransportChain, VlessAdapter, VlessFlow};
+#[cfg(feature = "xhttp")]
+use meow_proxy::XhttpAdapter;
```
madeye, line 17: **[ci-gate — reproduced]** `cargo fmt --all -- --check` fails on this PR: this import's placement/ordering here, the appended `#[cfg(any(…, feature = "xhttp"))]` in `meow-proxy/src/lib.rs:90` (needs multi-line wrapping), and several over-width `XhttpAdapter::new(…)` calls in `xhttp_adapter.rs` / `tests/xhttp_integration.rs`. Run `cargo fmt --all` before pushing.

crates/meow-transport/src/xhttp.rs
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    async fn connect_stream_up(&self, _inner: Box<dyn Stream>) -> Result<Box<dyn Stream>> {
+        Err(TransportError::Config(
+            "xhttp: stream-up mode is not yet implemented".into(),
+        ))
+    }
+}
+
+struct XhttpStream {
```
madeye, line 115: **[simplification — confirmed]** `XhttpStream` + `connect_stream_one` are a near byte-for-byte duplicate of `H2Stream` + `H2Layer::connect` in `h2.rs` — same fields, identical `poll_read`/`poll_write`/`poll_flush`/`poll_shutdown` logic (including the same capacity-handling bug flagged above) and the same handshake/spawn/response flow; only error-message prefixes differ.

Any fix to this bridging now has to land in two places and will silently drift. `H2Config` already models path/host — adding a `headers` field (as `GrpcConfig`/`WsConfig` already do) and reusing `H2Layer`, or factoring a shared raw-h2-stream type, would remove ~135 duplicated lines and fix both copies at once.

crates/meow-proxy/tests/xhttp_integration.rs
```diff
@@ -0,0 +1,283 @@
...
+    let mut buf = vec![0u8; 13];
+    timeout(TIMEOUT, conn.read_exact(&mut buf))
+        .await
+        .expect("read timed out")
+        .expect("read response");
+    assert_eq!(&buf, b"echo-response");
+}
+
+#[tokio::test]
+async fn xhttp_large_response() {
```
madeye, line 262: **[test-coverage — confirmed]** `xhttp_large_response` is byte-identical to `xhttp_connect_ok` — it reads the same 13-byte `echo-response` and never sends a large or multi-frame payload, despite its name and the spec's test plan (X2, `xhttp_tcp_relay_large_payload`) promising that coverage.

`poll_read` is never exercised across multiple DATA frames, so a flow-control/buffering regression — including the capacity-ignored unbounded-buffering issue flagged in `xhttp.rs` — would pass this suite silently. Suggest an actual >64KB echo across multiple DATA frames with a constrained window.

madeye (changes requested):
Requesting changes per the detailed review above (https://github.com/madeye/meow-rs/pull/376#pullrequestreview-4874351895).

Blocking items before this can merge:
1. **TLS**: `tls`/`sni`/`skip-cert-verify` must not be silently dropped — either wire a TLS layer or hard-error on these keys (ADR-0002 Class A).
2. **Destination encoding**: decide the interop story — as written, the wire format never carries the target address and cannot work with real xray/mihomo xhttp servers.
3. **h2 bridge correctness**: respect the capacity granted by `poll_capacity` (unbounded-buffering risk) and fix the stale `pending_write` cancellation hazard.
4. **Connection lifecycle**: shut down the spawned h2 driver/TCP socket on drop and on the non-200 path (currently leaks per health probe).
5. **Gates**: `cargo fmt --all -- --check` and the three-way `clippy -D warnings` must pass with the `xhttp` feature enabled.

Deduplicating `XhttpStream` with `H2Stream` (shared type or `H2Config.headers`) would resolve item 3 in one place, and the `xhttp_large_response` test should actually exercise a multi-frame response.

PR #387, opened 2026-08-12
fix(dns): pool TCP exchanges and honor IPv6 policy

aimkiray:
## Summary

Fix two Android-visible DNS problems while preserving resolver correctness:

1. Reuse up to four idle plain `tcp://` DNS connections per `DnsClient`, avoiding repeated TCP handshakes and Android `protect(fd)` calls.
2. Prefer IPv4 by querying A first and falling back to AAAA only when A has no usable answer.
3. Make the top-level `ipv6` setting drive every resolver, including custom `direct` proxy DNS resolvers and provider-loaded proxies.

## Correctness

- Track which address families were queried so an A-only cache entry never becomes a false AAAA miss.
- Preserve NODATA, NXDOMAIN, and SERVFAIL as distinct DNS server results.
- Check TCP streams out of the pool and return them only after a complete validated response; cancellation and malformed short frames discard the stream.
- Bound plain TCP concurrency to four connections and share one 5-second deadline across A and AAAA fallback.
- Keep hosts and cache behavior consistent when IPv6 is disabled.
- Bootstrap resolution intentionally remains dual-stack so hostname-based DNS upstreams can still use an available IPv4 address.

Pooling is intentionally scoped to plain `tcp://`; DoT and DoH behavior is unchanged.

## Scope

The core changes are in `meow-dns` and DNS/config wiring. Additional touched files only propagate the required resolver IPv6 constructor argument through existing call sites and tests.

## Verification

- `cargo fmt --all --check`
- `cargo test -p meow-dns` (146 unit tests passed; integration and doc tests passed, opt-in network/soak tests ignored)
- `cargo test -p meow-config` (all tests passed)
- `cargo clippy -p meow-dns -p meow-config --all-targets -- -D warnings`
- Android arm64 release build and emulator differential testing against `f7a11d4`

Device testing confirmed a shared ~5-second A/AAAA deadline, no AAAA query from custom direct DNS when `ipv6: false`, unchanged dual-stack fallback/cache/NXDOMAIN behavior, successful concurrent page loads, and no measurable throughput regression.


madeye (changes requested):
Thanks — both problems this targets are real (per-query TCP reconnect and the IPv6 policy leak), and fmt/clippy are clean. But review turned up several correctness bugs that block merge as-is:

**Blocking**

1. **AAAA resolution breaks for all dual-stack domains, even with `ipv6: true`.** The new A-first path (`crates/meow-dns/src/client.rs:388`) skips the AAAA query whenever A returns any address. The resolver keeps one merged per-domain cache entry, and `lookup_real_with_ttl`'s `ips.iter().find(is_ipv6).copied()?` (`resolver.rs:918-923`) then short-circuits to `None`, so the DNS server answers **NXDOMAIN** to an AAAA query for e.g. google.com (`server.rs:161→177`). Per the comment at `server.rs:183`, clients may negative-cache that NXDOMAIN against the bare name — poisoning subsequent A lookups too. It never self-heals: every TTL refresh re-runs the same A-first path. Fixing this needs per-family caching (or tagging which families were actually queried) so a family miss falls through to a real AAAA query.

2. **`tcp_exchange_pooled` is not cancellation-safe.** `query()` wraps the exchange in `tokio::time::timeout` (`client.rs:311`) and `query_pool`'s 2-client `select!` (`resolver.rs:369-400`) drops the losing future mid-exchange. Cancellation mid-`read_lp` leaves a framing-desynced stream in the pool slot — `*guard = None` (`client.rs:426`) is only reached on normal completion. The next caller reuses the desynced stream, `read_exact` on a garbage length prefix blocks, that caller is cancelled too and again doesn't evict — so every subsequent TCP DNS query on this client burns its full 5s timeout. Fix: `guard.take()` the stream out of the slot for the exchange and re-insert only on clean completion, so cancellation drops the connection.

3. **Panic on short frames.** `response[0]`/`response[1]` (`client.rs:418`) are indexed without checking the frame length; `read_lp` (`client.rs:701-708`) returns exactly the peer-declared length, which may be 0 or 1. A broken (or on-path-spoofable, for plaintext `tcp://`) upstream sending length prefix `0x0000`/`0x0001` panics the resolver task. The old path surfaced short buffers as `ClientError` via `decode_validated_response` — reusing Message-based id validation would also dedup the id check.

**Should fix**

4. The 5s timeout starts before `tcp_pool.lock()` is acquired, and all queries on a `tcp://` client serialize on one connection — under a page-load burst (the inflight map only dedups same-host), the k-th queued query waits (k−1)×RTT before its own I/O starts and spuriously times out on a healthy upstream. Scope the timeout to the exchange, or use a small pool (2–4 conns).
5. The per-adapter `Resolver` built for a `direct` proxy's custom `dns:` list never gets `set_ipv6`, and `Resolver::new` defaults to `ipv6: true` — so global `ipv6: false` silently doesn't reach it (`crates/meow-config/src/proxy_parser.rs:497`; `meow-proxy/src/direct.rs:80` can still dial v6). Making `ipv6` a constructor parameter (like `use_hosts`) turns omission into a compile error.
6. Hosts vs cache asymmetry in `resolve_ips` (`resolver.rs:825`): a v6-only hosts entry under `ipv6: false` returns `None` outright ("unresolvable"), while an identical v6-only *cached* entry falls through to a live upstream lookup. Whichever policy is intended, the branches should agree — the added test only covers a mixed v4+v6 entry.

**Minor**

7. Sequential A-then-AAAA doubles worst-case error latency to 2×timeout (10s vs the old parallel join's 5s), and when both fail, `last_err` reports the AAAA error, masking the original A failure.
8. Pooling covers only plain `Transport::Tcp` — DoT/DoH (in the default `full` bundle) still reconnect + full TLS handshake per query, so the docstring's per-query-reconnect/`protect(fd)` claim overstates the fix. Either extend pooling to the TLS transports or scope the claim.
9. Style: `absorb!` is defined and used once (AAAA arm) while the A arm hand-duplicates the same match/extend/min-ttl body — a small shared helper fn would delete both copies.

One candidate finding was checked and **refuted**: the server's qtype-28 early return bypassing hosts AAAA entries is deliberate, test-locked behavior.

aimkiray:
Addressed the review in `35f30ae` without rewriting the existing branch history:

- family-aware cache metadata prevents A-first lookup from poisoning AAAA
- cancellation-safe bounded 4-connection plain TCP pool
- short-frame validation before response decoding
- one 5-second deadline shared by A and AAAA fallback
- IPv6 policy is a required resolver constructor argument and reaches custom direct/provider resolvers
- hosts/cache misses are consistent; DNS server preserves NODATA/NXDOMAIN/SERVFAIL
- documentation now scopes pooling to plain `tcp://`

Local fmt, `meow-dns` (146 tests), `meow-config`, and strict clippy for both affected crates pass. The revised implementation also passed the arm64 emulator differential scenarios described in the PR body.

crates/meow-config/src/dns_parser.rs
```diff
@@ -105,6 +106,7 @@ pub async fn parse_dns(
         mode,
         hosts,
         use_hosts,
+        raw.ipv6.unwrap_or(false),
```
madeye: **[review] Silent dual-stack regression for configs that omit `ipv6:`.** Before this PR, `raw.ipv6` was display-only and `DnsClient::lookup_ip` always did `tokio::join!(A, AAAA)` — every existing deployment got dual-stack DNS regardless of the flag. Now `raw.ipv6.unwrap_or(false)` makes IPv4-only the default: any config without an `ipv6:` line silently loses AAAA everywhere (resolver `filter_enabled_ips`, `lookup_ipv6_result` → NoData, DirectAdapter dials to v6-only hosts fail with `MeowError::Dns`, GeoIP rules that matched on real v6 IPs fall through to MATCH). Matches the mihomo `dns.ipv6: false` default, but it is a breaking behavior change for existing meow-rs users upgrading with an unchanged config — worth defaulting to `true` for compat, or at least calling out loudly in release notes.

crates/meow-dns/src/resolver.rs
```diff
@@ -858,52 +954,106 @@ impl Resolver {
...
         }
-        let ips = self.lookup_actual_all(host).await?;
-        let ip = ips.into_iter().find(family)?;
-        let ttl = self
-            .cache
-            .get_with_ttl(host)
-            .map_or(HOSTS_ANSWER_TTL, |(_, remaining)| remaining);
-        Some((ip, ttl))
+
+        let Some(result) = self.lookup_actual_family(host, record_type).await else {
```
madeye: **[review] Per-family lookups lost single-flight dedup.** The DNS server path (`handle_query` → `lookup_ipv4_result`/`lookup_ipv6_result` → `lookup_real_with_ttl` → `lookup_actual_family`) never touches `self.inflight`, unlike the old path through `lookup_actual_all` (which coalesces concurrent lookups via the broadcast-channel guard). A page-load burst of N concurrent queries for the same cold domain now fans out N × (main+fallback) upstream queries instead of 1 — reintroducing the exact thundering-herd/upstream-starvation problem this PR sets out to fix, and multiplying load on the 4-permit TCP pool. Suggest routing the cache-miss path through the same inflight-keyed machinery (keyed by host+family).

crates/meow-dns/src/resolver.rs
```diff
@@ -404,26 +440,57 @@ async fn query_pool(clients: &[Arc<DnsClient>], host: &str) -> Option<PoolLookup
...
+    record_type: RecordType,
+) -> Option<PoolFamilyResult> {
+    let mut pending = FuturesUnordered::new();
+    for (index, client) in clients.iter().enumerate() {
+        pending.push(async move { (index, client.lookup_family(host, record_type).await) });
+    }
+
+    let mut empty = None;
+    let mut nxdomain = false;
+    while let Some((index, result)) = pending.next().await {
```
madeye: **[review] Negative answers stall on the slowest upstream.** `query_pool_family` drains every client future to completion before returning the remembered NODATA/NXDOMAIN — there is no early return for negatives. With 2+ upstreams and one dead/black-holed, an AAAA NODATA that a healthy upstream answers in ~10 ms still waits the dead client's full 5 s timeout. Since AAAA-NODATA is the common case for v4-only domains, one degraded upstream turns most AAAA queries into ~5 s responses (and with the `qtype==28` guard only active when ipv6 is disabled, dual-stack users hit this constantly). Consider returning on the first definitive negative once no remaining client can still produce a positive, or capping the straggler wait.

crates/meow-dns/src/cache.rs
```diff
@@ -231,6 +282,72 @@ impl DnsCache {
...
+            if existing.expire_at > now {
+                merged.extend(
+                    existing
+                        .ips
+                        .iter()
+                        .copied()
+                        .filter(|ip| !queried.contains_ip(*ip)),
+                );
+                merged_queried = existing.queried.union(queried);
+                expire_at = existing.expire_at.min(incoming_expire_at);
```
madeye: **[review] `min()` merge lets a short-TTL family answer expire the other family early.** `expire_at = existing.expire_at.min(incoming_expire_at)` applies to the whole merged entry. Cached A with TTL 3600 s + later AAAA NODATA (empty ips, TTL clamped to `clamp_ttl` MIN of 10 s) → the entire entry, including the still-fresh A answer, now expires in 10 s. That is a ~360× re-query amplification for dual-stack-queried domains — the same cache-thrash burst this PR is fighting. Per-family expiry (or keeping the existing expiry for IPs of families the incoming result did not query) avoids it.

crates/meow-dns/src/resolver.rs
```diff
@@ -801,14 +878,20 @@ impl Resolver {
     pub async fn resolve_ips(&self, host: &str) -> Option<Vec<IpAddr>> {
         if self.use_hosts {
             if let Some(ips) = self.hosts.search(host) {
-                if !ips.is_empty() {
-                    return Some(ips.clone());
-                }
+                return self.filter_enabled_ips(ips);
```
madeye: **[review] Hosts branch returns `None` with no fallthrough once filtered empty.** Unlike the cache branch below (which falls through to `lookup_actual_all` when `filter_enabled_ips` returns `None` and the required families were not queried), the hosts branch always returns. A hosts entry with only IPv6 addresses while `ipv6: false` (or an empty list) now makes `resolve_ips` return `None` outright — DirectAdapter's `resolve_targets` treats that as a hard `MeowError::Dns` failure, so the host becomes unreachable even though upstream DNS could supply a usable A record. Old code only returned early when `!ips.is_empty()`.

crates/meow-dns/src/resolver.rs
```diff
@@ -299,6 +302,47 @@ struct PoolLookupResult {
...
+    NxDomain,
+    Failed,
+}
+
+enum PoolFamilyResult {
+    Response(PoolLookupResult),
+    NxDomain,
+}
+
+fn family_answer_or_remember(
```
madeye: **[review] Asymmetric negative bookkeeping makes NXDOMAIN-vs-NODATA tier-order dependent.** The empty-`Response` arm overwrites `*negative` unconditionally, while the `NxDomain` arm only writes `if negative.is_none()`. Across the policy → main → fallback sequence in `lookup_actual_family`, an earlier tier's NXDOMAIN is silently clobbered by a later tier's NOERROR-empty (but not vice versa), so the rcode the client sees depends on which tier answered, not on a deliberate priority. If NODATA is meant to beat NXDOMAIN, make that an explicit documented rule; otherwise use the same first-wins guard in both arms.

crates/meow-dns/src/cache.rs
```diff
@@ -36,6 +36,40 @@ struct CacheEntry {
     ips: Box<[IpAddr]>,
     expire_at: Instant,
     source: Option<Arc<str>>,
+    queried: QueryFamilies,
```
madeye, line 62: **[review] CLAUDE.md architecture-invariant violation: no byte counts for `CacheEntry`.** Repo CLAUDE.md: "Any PR touching these types **must** include before/after byte counts (from `-Zprint-type-sizes`) in the commit body: … DNS `CacheEntry` / `ReverseEntry` (crates/meow-dns/src/cache.rs) — M2 exit 72 B per LruEntry slot." This PR adds the `queried: QueryFamilies` field — measured impact 48 B → 56 B per entry (the u8 forces a new 8-byte word; padding does not absorb it) — and none of the four commit bodies in the range mention byte counts. Please add the `-Zprint-type-sizes` before/after to the commit body (or pack the 2 bits into an existing niche).

crates/meow-dns/src/client.rs
```diff
@@ -297,42 +343,153 @@ impl DnsClient {
...
+        wire: &[u8],
+        expected: &ExpectedResponse,
+    ) -> Result<Message, ClientError> {
+        let _permit = self
+            .tcp_pool
+            .permits
+            .acquire()
+            .await
+            .map_err(|_| ClientError::Protocol("TCP pool closed"))?;
+        let pooled = self.tcp_pool.idle.lock().await.pop();
```
madeye, line 566: **[review] Stale pooled stream can burn the whole shared deadline before the fresh-connect retry.** On a reused stream, a half-closed connection (write succeeds, peer never answers) makes `tcp_message_exchange` block in `read_lp` until the outer `self.timeout` from `query()` is nearly exhausted; the fallback `factory().connect_tcp(...)` + full exchange then has almost no budget, so a query that would succeed in ~30 ms on a fresh connection spuriously times out. RST/EOF failures fail fast, but idle-timeout half-closes by upstreams/NAT are exactly what a keep-alive pool accumulates. Consider a short read deadline on the pooled attempt (or budget the retry separately).

crates/meow-dns/src/resolver.rs
```diff
@@ -858,52 +954,106 @@ impl Resolver {
...
         &self,
         host: &str,
-        family: fn(&IpAddr) -> bool,
-    ) -> Option<(IpAddr, Duration)> {
-        if let Some((ips, remaining)) = self.cache.get_with_ttl(host) {
-            let ip = ips.iter().find(|ip| family(ip)).copied()?;
-            return Some((ip, remaining));
+        record_type: RecordType,
+    ) -> AddressLookupResult {
+        let queried = match record_type {
```
madeye: **[review] Duplicated family dispatch — reuse `QueryFamilies`.** This `match record_type { A => IPV4, AAAA => IPV6, … }` is byte-for-byte the same mapping as in `DnsClient::lookup_family` (client.rs), and the `matches_family` closure below re-derives per-IP family membership that `QueryFamilies::contains_ip` (added in this PR, cache.rs — currently private) already implements. A `QueryFamilies::from_record_type()` helper plus a `pub(crate) contains_ip` gives one source of truth; today a change to one of the three sites silently desyncs the others.

crates/meow-dns/src/resolver.rs
```diff
@@ -971,35 +1130,97 @@ impl Resolver {
...
+                Some(&result.source),
+                result.queried,
+            );
             return Some(result.ips);
         }
 
         self.try_fallback(host).await
     }
 
+    async fn lookup_actual_family(
```
madeye: **[review] Second parallel resolution pipeline duplicates the policy/fallback/gating flow.** `lookup_actual_family`/`query_pool_family`/`try_fallback_family` re-implement the domain-gate → policy → main → fallback structure of `do_lookup`/`query_pool`/`try_fallback`, including two verbatim copies of the `family_answer_or_remember` + `ip_gated` + fallback block within this function alone. Future changes to fallback/policy semantics now have to land in both pipelines (they already diverge: single-flight and negative bookkeeping exist in only one each). Consider one pipeline parameterized by the queried record set.

aimkiray:
Closing in favor of two smaller, independent PRs split from this one:

- #462 — TCP exchange pooling for plain `tcp://` upstreams (1 file, +251/-4). Self-contained keep-alive pool that avoids the per-query reconnect (and Android `protect(fd)` cost) that caused the intermittent timeouts. Based on current `main`, regression-green.
- #463 — prefer-IPv4 A-first lookup, per-family cache expiry, top-level `ipv6` driving every resolver, single-flight, first-definitive-negative (24 files, +1894/-348). The cache/pipeline rework that drew the most review feedback here, now isolated so it can be reviewed on its own. ADR-0011 `CacheEntry` byte delta documented in the commit body. Based on current `main`, regression-green.

Both are filed issue-first (#460, #461) per the contributing guide, rebased onto the latest `main` (this branch had fallen 32 commits behind and was `DIRTY`/`CONFLICTING`), and touch disjoint regions of `client.rs` so they can land in either order. The original 5-commit history and the round-2 review fixes are preserved across the two new PRs.

Thanks for the review feedback — it shaped the split (e.g. the per-family short-read deadline and cancellation-safety cases now live in the pooling PR, and the `min()`-TTL-collapse and `ipv6`-propagation fixes are scoped to the per-family PR).

PR #389, opened 2026-08-13
fix(config): accept short-id: null and bind-address '*' for mihomo compat

### LINKED ISSUE #388: config compat: short-id: null silently drops VLESS nodes; bind-address: '*' is fatal (mihomo tolerates both)
These two config constructs are common in real-world subscription / Clash Verge
profiles and are accepted by mihomo, but meow-rs currently rejects them — one
silently drops nodes, the other kills the process. Both repros below are sanitised
(fake server/uuid/key).

## 1. `reality-opts.short-id: null` (YAML null) drops every VLESS node

Many subscription generators emit `short-id: null` when the server has no
short-id. mihomo treats a null `short-id` as absent/empty and the node works.
meow-rs rejects it and **silently drops the node**, which then cascades.

### Repro (`meow -t`)

```yaml
proxies:
  - { name: t, type: vless, server: example.com, port: 443, uuid: 00000000-0000-0000-0000-000000000000, udp: true, flow: xtls-rprx-vision, tls: true, servername: www.example.com, reality-opts: { public-key: "AAA...", short-id: null }, client-fingerprint: chrome, network: tcp }
```

### Actual

```
WARN meow_config: Failed to parse proxy: vless: reality-opts.short-id must be a hex string
WARN meow_config::proxy_parser: Proxy 't' not found for group 'PROXY', skipping
WARN meow_config: Failed to parse proxy group 'PROXY': group 'PROXY' has no valid proxies or providers
```

Omitting `short-id` entirely works — only the explicit `null` is rejected. On my
own subscription this dropped all 10 VLESS nodes.

### Expected / suggested fix

Match mihomo: treat `short-id: null` (and `~`) like an absent/empty short-id. E.g.
deserialize as `Option<String>` (null → `None`) and only hex-validate when a value
is actually present.

## 2. `bind-address: '*'` is fatal (whole process exits)

`bind-address: '*'` is the idiomatic wildcard in clash/mihomo configs (Clash Verge
writes it). mihomo accepts `'*'` (and `0.0.0.0`) as "all interfaces". meow-rs
accepts `0.0.0.0` but treats `'*'` as a hard error and **the process exits**.

### Repro

```yaml
mixed-port: 7890
allow-lan: true
bind-address: '*'
```

### Actual

```
INFO  meow: Config loaded from ...
ERROR meow: meow-rs stopped with an error error=listener 'mixed': invalid bind address '*': invalid IP address syntax
Error: listener 'mixed': invalid bind address '*': invalid IP address syntax
```

### Expected / suggested fix

Match mihomo: accept `'*'` / empty as wildcard ≡ `0.0.0.0` / `::`. E.g. normalise
`bind-address` before parsing (`*` / `""` → `0.0.0.0`) instead of feeding the raw
string to `IpAddr::from_str`.

## Environment

meow-rs v0.20.1 (`x86_64-pc-windows-msvc` release binary). Both configs load and
run unchanged on mihomo (Clash Meta).

gqf2008:
Closes #388.

Two config constructs that are common in real-world subscription / Clash Verge
profiles are accepted by mihomo but were rejected by meow-rs — one silently
dropped nodes, the other aborted the process. This PR makes meow-rs tolerate
both, matching mihomo.

## 1. `reality-opts.short-id: null` (YAML null) no longer drops VLESS nodes

mihomo decodes `short-id` into a Go `string` (`adapter/outbound/reality.go`):
a YAML `null` becomes the zero value `""`, and `hex.Decode(dst, "")` succeeds
with `n == 0`, yielding `ShortID == [8]byte{0}` — no error. Many subscription
generators emit `short-id: null` when the server has no short ID, so mihomo runs
these nodes fine while meow-rs rejected them with
`vless: reality-opts.short-id must be a hex string` and dropped the node
(cascading into `group has no valid proxies`).

**Fix:** treat `Value::Null` the same as an absent key → empty short ID, instead
of erroring. Explicit values are unchanged: a present string is still required
to be valid hex (`parse_reality_short_id`), and a present non-string still
errors.

## 2. `bind-address: '*'` no longer aborts the process

mihomo maps the wildcard to "all interfaces" only when `allow-lan` is on
(`listener/listener.go` `genAddr`): `allow-lan: true` + `bind-address: '*'` →
`:port`; `allow-lan: false` → `127.0.0.1:port` regardless. meow-rs fed the raw
`'*'` to `IpAddr::from_str`, which failed and exited the whole process at
listener startup.

**Fix:** normalise `'*'` → `0.0.0.0` in `bind_socket_addr`. This is only reached
when `allow-lan` is on, because `allow-lan: false` already forces `127.0.0.1`
upstream in `meow-config` — so the loopback-only default is unchanged.

> Deliberate deviation, flagging for review: mihomo binds `:port` (dual-stack on
> Linux); this maps to `0.0.0.0` (all-IPv4), which is the conventional,
> cross-platform-safe reading and matches meow-rs's existing `0.0.0.0` wildcard
> convention. `bind-address: '::'` remains the explicit dual-stack opt-in. Happy
> to switch to `::` if you'd rather mirror mihomo's dual-stack exactly.

## Testing

- New unit tests: `reality_short_id_null_treated_as_absent` (null → empty; hex
  still decodes; non-hex still rejected) and `wildcard_bind_address_is_ipv4_any`
  (`*` → `0.0.0.0`).
- `cargo fmt --all -- --check` clean; `cargo clippy --all-targets` clean under
  default / `--no-default-features` / `--all-features`.
- `cargo test` for `meow-config` (282) and `meow-app` (14): all pass.
- Functional check against a live 50-node mihomo subscription (40 AnyTLS + 10
  VLESS-with-`short-id: null`): previously the 10 VLESS nodes were dropped; with
  this patch all 65 proxies load and the null-short-id VLESS nodes complete the
  REALITY handshake (measured real latencies) and proxy end-to-end.

Note: one unrelated test, `meow-api`'s
`post_dns_query_unknown_name_returns_null_answer`, fails when run on a machine
whose DNS is intercepted by a fake-ip proxy (it resolves `no-such-host.invalid`
to a `198.18.x.x` fake-ip instead of NXDOMAIN). It fails identically on
unmodified `main` under that environment and is untouched by this PR.

### DIFF
--- a/crates/meow-app/src/main.rs
+++ b/crates/meow-app/src/main.rs
@@ -731,9 +731,16 @@ fn run_cmd(cmd: &str, args: &[&str]) -> Result<()> {
 /// of `format!("{listen}:{port}")`. The string form produces an unparseable
 /// `:::7890` for an IPv6 listen address like `::`, which silently broke
 /// dual-stack binding (`bind-address: '::'`); `SocketAddr::new` handles IPv4
-/// and IPv6 uniformly. `listen` is always an IP literal here (`0.0.0.0`, `::`,
-/// `127.0.0.1`, or a specific address).
+/// and IPv6 uniformly. `listen` is normally an IP literal here (`0.0.0.0`,
+/// `::`, `127.0.0.1`, or a specific address); the mihomo wildcard `'*'` is
+/// normalised to `0.0.0.0`.
 fn bind_socket_addr(listen: &str, port: u16) -> Result<SocketAddr> {
+    // mihomo treats `bind-address: '*'` as the wildcard "all interfaces" — it is
+    // the idiomatic value Clash Verge writes. This only surfaces when allow-lan
+    // is on, since allow-lan:off already forces 127.0.0.1 upstream, so map it to
+    // the IPv4-any literal instead of aborting the whole process (issue #388).
+    // `::` remains the explicit dual-stack opt-in.
+    let listen = if listen == "*" { "0.0.0.0" } else { listen };
     let ip: IpAddr = listen
         .parse()
         .map_err(|e| anyhow::anyhow!("invalid bind address '{listen}': {e}"))?;
@@ -1080,6 +1087,15 @@ mod tests {
     #[test]
     fn invalid_bind_address_errors() {
         assert!(bind_socket_addr("not-an-ip", 80).is_err());
+    }
+
+    #[test]
+    fn wildcard_bind_address_is_ipv4_any() {
+        // mihomo wildcard: `bind-address: '*'` binds all interfaces (0.0.0.0)
+        // instead of aborting the process (issue #388).
+        let a = bind_socket_addr("*", 7890).unwrap();
+        assert_eq!(a.to_string(), "0.0.0.0:7890");
+        assert!(a.is_ipv4());
     }
 
     #[cfg(target_os = "windows")]
--- a/crates/meow-config/src/proxy_parser.rs
+++ b/crates/meow-config/src/proxy_parser.rs
@@ -1474,11 +1474,14 @@ fn parse_vless_reality_opts(
     let mut public_key = [0u8; 32];
     public_key.copy_from_slice(&public_key_bytes);
 
+    // mihomo tolerates `short-id: null` — many subscription generators emit it
+    // when the server has no short ID — and treats it as absent. Map it to the
+    // empty short ID instead of rejecting (and silently dropping) the node.
     let short_id_str = match opts.get("short-id") {
+        Some(serde_yaml::Value::Null) | None => "",
         Some(value) => value
             .as_str()
             .ok_or_else(|| "vless: reality-opts.short-id must be a hex string".to_string())?,
-        None => "",
     };
     let short_id_vec = parse_reality_short_id(short_id_str)?;
     let mut short_id = [0u8; 8];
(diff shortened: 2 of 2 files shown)


madeye (changes requested):
Thanks for the PR — both fixes are correct for their stated cases (verified `short-id: null` parity against mihomo's zero-value decoding, and `'*'` correctly scoped to `bind-address` only, matching mihomo which also rejects `'*'` for `external-controller`/`dns.listen`). A few issues need addressing before merge:

**1. `'*'` bypasses `allow-lan: false` for explicit `listeners:` entries (security-relevant)**

`build_named_listeners` (`crates/meow-config/src/lib.rs:1431`) passes an explicit `listen:` through verbatim — the allow-lan forcing to `127.0.0.1` only applies when `listen` is omitted. So with `allow-lan: false` plus `listeners: [{name: x, type: mixed, port: 7891, listen: '*'}]` (a common Clash Verge template), the process previously aborted loudly with "invalid bind address '*'"; after this PR it silently binds `0.0.0.0`, exposing an unauthenticated proxy port despite `allow-lan: false`. At minimum the inline comment in `bind_socket_addr` ("this only surfaces when allow-lan is on") must be corrected — it's wrong for the named-listener path; ideally `'*'` there should honor allow-lan or at least log a warning.

**2. IPv4-only `'*'` diverges from mihomo (needs an explicit decision)**

mihomo's `genAddr` maps `'*'` to `":port"` — a Go dual-stack listen. This PR maps it to IPv4-only `0.0.0.0`, so with `allow-lan: true, ipv6: true, bind-address: '*'` (the exact idiom the PR targets), IPv6 LAN clients that can reach mihomo get connection-refused from meow-rs while startup reports success. You flagged this in the PR body — let's resolve it explicitly: map `'*'` to `'::'` (dual-stack) when `ipv6: true`, or document the v4-only choice.

**3. Numeric `short-id` still drops the node — on the same line this PR rewrote**

`short-id: 1234` unquoted is a YAML integer, so it hits the `Some(value) => value.as_str().ok_or_else(...)` arm (`crates/meow-config/src/proxy_parser.rs:1482`) and the node is rejected — the same #388 failure mode. mihomo's decoder (`WeaklyTypedInput: true`) coerces int→string (`"1234"` → `[0x12, 0x34]`). Worth handling `Value::Number` here too since generator tolerance is the stated goal.

**Same-class adjacencies (fine as follow-ups if you'd rather keep this PR small):**
- hysteria2: `port: null` hits the `Some` arm and errors ("missing port", node dropped) instead of falling through to the `ports` range (`proxy_parser.rs:733`) — mihomo skips nil and dials via the port range.
- snell: `version: null` hard-errors instead of defaulting to V4 like the absent-key branch (`proxy_parser.rs:265`); the consistent fix is `Some(Value::Null) | None => SnellVersion::V4`.

madeye:
Closing: superseded by c0f153c (already on main), which fixed #388 at the config-parse layer — including the empty-string wildcard spelling this branch doesn't handle — and this branch no longer merges cleanly (content conflict in `proxy_parser.rs`; keeping both sides would leave two divergent `'*'` → `0.0.0.0` normalizations).

Two useful things surfaced by reviewing this PR, tracked separately:

- The one genuinely new bug — unquoted numeric `short-id: 1234` still drops the node on main while mihomo accepts it — is filed as #408.
- The review also noted that four hysteria2 parsers in the same file share the `null`-treated-as-present shape that #388 was about (`port`, `hop-interval`, `up`/`down`, `alpn`), so the same generator artifacts can still drop hy2 nodes; worth a follow-up if that class of report recurs.

Thanks — the branch pointed at the right problem; main just got there first via a deeper fix.

Issue #377, opened 2026-08-01
反馈： 与 mihomo 几个行为不一致的地方


Ozarklake:
今天尝试使用 meow-rs 失败了，发现有几个点比较影响setup：

1. release 中的二进制默认没有编译 anytls， 现在这个协议在机场里比较主流，默认不编译的话比较影响入手的流畅度。并且其他平台可以自行编译，iOS 要自己搞就比较麻烦了。建议重新评估一下是否要默认编译 anytls。
2. 以下 vless 配置无法正常工作，不太确定具体的原因是什么，vless 也是目前非常流行的协议，无法使用的话影响也挺大
```yaml
    name: HK-1
    type: vless
    server: server.com
    port: 10000
    uuid: redact
    udp: true
    packet-encoding: xudp
    cipher: auto
    client-fingerprint: chrome
    tls: true
    skip-cert-verify: false
    servername: s0.awsstatic.com
    reality-opts:
      public-key: redact
      short-id: redact
    network: grpc
    grpc-opts:
      grpc-service-name: update
```
3. hysteria2 这个协议在 mihomo 的实现中，port 虽然声称是 Required 的，但是实际可以不填，meow-rs 要求是必填，某些机场提供的 proxies 配置可能并不那么规范，这也会导致摩擦。 建议评估一下是否要参考 mihomo 的实现。
4. rule-provider 进行下载时，默认会使用第一个 proxy 的配置，这里的设计有点奇怪，如果我没有配置 proxies(全从 proxy provider 中拿)就会导致直接失败，建议参考 mihomo 的实现，增加一个字段用于描述 rule provider 下载时的网络策略。

xmh0511:
@madeye The mentioned PRs seems to only fix the last two issues. The first two issues are significant for the experience of using `meow-rs`, so I think the first two issues also have high priority to be fixed.

madeye:
All four items are now fixed on main and shipped in v0.20.0 / v0.20.1:

1. **anytls in release binaries** — fixed in #381: `anytls` is now part of the `full` feature bundle, so every published binary (including the iOS builds) carries it. v0.20.1 additionally fixed the anytls build for the mips32 targets.
[...]

PR #394, opened 2026-08-15
feat(proxy): connection multiplexing — sing-mux (smux/yamux/h2mux) + Xray Mux.Cool

### LINKED ISSUE #386: perf: bulk-relay loopback throughput trails Go mihomo ~40% (19 vs 31 Gbps)
## Symptom

While refreshing the README benchmarks for #356 (PR #385), the side-by-side suite showed a consistent bulk-throughput gap vs Go mihomo v1.19.29. Three full `bash bench.sh` runs on an Apple M4 (macOS 26.5.2, loopback, `mode: direct`, SOCKS5), medians:

| Workload | mihomo (Go) | meow-rs v0.20.1 | Delta |
|---|---|---|---|
| TCP 64 MB×1 | 31.30 Gbps | 19.07 Gbps | −39% |
| TCP 1 MB×10 | 35.77 Gbps | 18.86 Gbps | −47% |
| TCP 4 KB×10000 | 2.14 Gbps | 1.99 Gbps | −7% |

Per-run spread was small (rust 18.8–19.4 Gbps on 64 MB×1 across runs), so this is a stable gap, not noise.

## Not a recent regression

meow-rs's numbers match its own July 2026 dev-harness baseline (`docs/benchmarks/baseline-2026-07-15.json`: 17.3–17.4 Gbps on the same workloads). The old README table showed near-parity (~5 Gbps both) because that host was slow enough that neither kernel was CPU-bound. On fast hardware the relative per-proxy CPU cost is now visible: mihomo does the same relay work in ~60% of the CPU.

Latency (p50 130 µs vs 135 µs), connection rate, and small-message throughput are all at or near parity — the gap is specific to bulk streaming.

## Prime suspect: 4 KiB relay buffer

`crates/meow-tunnel/src/relay.rs`:

- `RELAY_BUF_SIZE = 4 * 1024` — each fill+drain cycle moves at most 4 KiB, i.e. one `read` + one `write` syscall per 4 KiB. At 19 Gbps that is ~580k read/write pairs per second per direction; the relay is very plausibly syscall-bound.
- Go mihomo relays through 32 KiB pooled buffers (`pool.RelayBufferSize`), an ~8× lower syscall rate for the same byte count.
- `RELAY_CYCLES_PER_POLL = 64` bounds one poll at 256 KiB relayed before a voluntary yield; at 32 KiB buffers the same cap would allow 2 MiB per poll (or the cap could be rebalanced).

The 4 KiB size was chosen for footprint: buffers are stack-allocated `[0u8; RELAY_BUF_SIZE]` arrays in the caller's async frame (ADR-0011 T6, zero-per-relay-setup-allocation invariant, see CLAUDE.md "Architecture Invariants"). Two arrays per relay means buffer size directly multiplies per-connection task/frame memory — 4 KiB ↔ ~8 KiB per relay vs 32 KiB ↔ ~64 KiB per relay. That trade-off is the crux of this issue.

## Suggested investigation

1. **Cheap A/B first**: rebuild with `RELAY_BUF_SIZE = 32 * 1024` and rerun `bash bench.sh` (3 runs, medians per ADR-0006). If the gap mostly closes, this is a buffer-size/footprint trade, not a code-efficiency problem.
2. Profile the relay hot path under the 64 MB workload (Instruments/samply) to confirm syscall dominance vs in-process copy overhead.
3. If a larger buffer wins, decide the footprint trade explicitly: tiered sizes (small default, upgrade after N bytes relayed on a connection), or a measured bump with before/after RSS-under-load and per-connection frame sizes in the commit body (ADR-0008/0011 gates apply — HP reproducer counts must not increase).
4. Re-check relay fairness after any buffer change: starvation shows up in quiet-direction p99, not p50 (see `docs/benchmarks/` relay-fairness notes); the one-cycle-per-turn alternation in `relay.rs` interacts with buffer size.

## Context

- Bench harness: `bash bench.sh` / `crates/meow-bench` (`--only throughput` for quick iterations)
- Loopback bulk throughput measures per-proxy CPU overhead, not real-network capability — both kernels saturate real multi-Gbps links with headroom. This issue is about closing the efficiency gap, not a user-visible slowdown on typical links.

aimkiray:
> Stacking: this branch carries one independent harness fix that lands first and then drops out of this PR on rebase — `fix/bench-harness` (meow-bench deadline hardening + --binary-arg + echo-timeout reporting, 4 commits, PR #399). The 32 KiB relay-buffer throughput fix was superseded by #396 (tiered relay buffers, already merged into main; closes #386).

## Overview

Connection multiplexing for VLESS / Trojan / Shadowsocks / VMess outbounds, aligned with mihomo's `mux:` config — many logical streams over **one physical node connection**:

```yaml
proxies:
  - name: node
    type: vless   # or trojan / ss / vmess
    ...
    mux:
      enabled: true
      protocol: h2mux       # smux | yamux | h2mux | muxcool (default h2mux)
      max-connections: 4
      min-streams: 4
      max-streams: 4
      padding: false
      only-tcp: false
```

Two protocol families, chosen by server type:

| Protocol | Applies to | Server requirement |
|---|---|---|
| `smux` / `yamux` / `h2mux` (sing-mux) | VLESS / Trojan / Shadowsocks / VMess | sing-box / mihomo inbounds with multiplex enabled (plain ss-server does not speak sing-mux) |
| `muxcool` (Xray Mux.Cool) | **VLESS / VMess** | Xray-core, or sing-box / mihomo VLESS/VMess inbounds (CommandMux is handled natively, no inbound config needed) |

## Design highlights

- **MuxClient connection pool**: mirrors metacubex/sing-mux offer/offerNew semantics, min/max bounds, idle-session eviction (60s), concurrent opens never exceed max-connections;
- **Single-driver task discipline**: each session is owned by one task that polls both directions (same discipline as h2mux); outbound frames go through a bounded channel (PollSender poll-based backpressure), inbound frames are demuxed by the driver's state machine. Layered connections (VlessConn / Vision / TLS) must **not** be split-polled concurrently — that once caused duplicated stream bytes and TLS record desync;
- **Inbound delivery never blocks the driver**: a frame hitting a full queue parks (at most one) and pauses the connection read (session-wide flow control, TCP window semantics), eliminating the write-then-read consumer ↔ driver deadlock (dedicated regression test);
- **Mux.Cool teardown is End-echo independent**: sing-box's Mux.Cool server (sing-vmess/mux.go) closes a stream silently and never echoes the client's End frame (xray-core does echo). The client removes the stream-map entry on both the shutdown and drop paths, which closes the inbound channel and surfaces EOF — no consumer ever waits for a frame that will not arrive, and the per-stream map cannot accumulate under connection churn (unit-tested against a silent mock server). Data the server sent before processing the End is still delivered first, then EOF; note that Mux.Cool End is a full close on the wire (matching xray), so TCP-style half-close-then-read is not a supported pattern on this protocol.
- Mux.Cool frame codec as pure functions (field-by-field aligned with xray common/mux, tolerant of unknown trailing meta bytes); the reader errors — never panics — on malformed frames;
- UDP-over-mux: sing-mux uses flagUDP streams with length-prefixed datagrams; muxcool uses Keep frames with per-datagram destinations;
- Per-protocol dial closures: VLESS/Trojan target the reserved mux FQDN in their request headers, VMess seals it into the VMess request header, SS dials its cipher stream (all plugin layers) directly; the SS adapter shares its dial state with mux sessions through an Arc-wrapped core so SIP003 plugin handles stay alive;
- `mux` is gated behind a Cargo feature (on by default, excluded from `minimal`); no-mux builds warn loudly when an enabled mux block is parsed.

## Testing & interop

- Unit: frame-codec byte assertions, malformed-input fuzzing, session state machine, flow-control/deadlock regression, churn leak regression, and silent-server teardown regressions (shutdown EOFs the read side without a server End echo, drop removes the map entry, exactly-once End, post-End discard of buffered data, UDP PacketConn cleanup) — meow-proxy mux suites all green;
- Local sing-box v1.13.18 interop: VLESS (plaintext/TLS/Reality/Reality+Vision), Trojan (TLS), **Shadowsocks and VMess** inbounds with multiplex enabled — 4 sing-mux protocols × all transports pass TCP (204/200) + UDP echo + 6-way concurrency (scripts under `target/e2e/mux-interop/`; sing-box configs are not distributed with the repo);
- Real node (anonymized, VLESS Reality+Vision): `muxcool` passes 204 (h2 + http/1.1) with concurrent streams reusing one physical session; sing-mux against the same node fails as expected (Xray only speaks Mux.Cool);
- Physical device (arm64 Android): the release APK with a real subscription (5 nodes incl. Reality+Vision + muxcool) connects end-to-end — tun0 up, engine sessions to the real node, HTTPS 204 / browsing / multi-MB downloads all through the mux sessions; full-chain TCP/HTTPS/UDP with stable memory over long runs (no leaks).

## Performance (local sing-box v1.13.18, meow-bench, 64 workers)

Before the teardown fix, connection churn against a silent-End server collapsed to ~24/s: every half-closed relay waited out its 30s half-close linger (no End echo ever arrived), listener slots saturated, and subsequent connects stalled. After the fix:

| workload | meow (this PR) | xray-core 26.3.27 (same setup) |
|---|---|---|
| connrate (tls+vision muxcool, 6s window) | ~1380/s | ~1650/s (both are Windows-loopback TIME_WAIT-capped; meow 745/s vs xray 634/s on 10s windows) |
| connrate (reality+vision / reality / h2mux) | ~1320/s each | — |
| latency (tls+vision muxcool) | p50 704µs / p99 1.1ms | p50 2.1ms / p99 2.8ms |
| load RSS | ~18 MB | ~470 MB |

(Bulk-transfer throughput is covered by #396's tiered relay buffers, already in main.)

## Regression gates

`cargo fmt --check` / clippy three-way (default, no-default-features, all-features) with `-D warnings` / `RUSTDOCFLAGS=-D warnings cargo doc` / `cargo test --lib` all pass.

## Docs

Adds `docs/specs/proxy-mux.md` (wire protocol field-by-field, architecture mapping, interop boundaries); refreshes the stale mux references in `docs/specs/proxy-vless.md`, `docs/specs/proxy-vless-test-plan.md` and `docs/migration-from-go-mihomo.md`.


aimkiray:
> 注:本机 Windows 环境下 `meow-dns` 的 `udp_client_times_out_on_unroutable` 在 `main` 上同样失败(不可路由地址在本环境立即返回网络错误而非超时,环境相关;本 PR 未触及 meow-dns,该行为由 #387 修复)。

aimkiray:
> Note: on this Windows host, `meow-dns`'s `udp_client_times_out_on_unroutable` fails identically on `main` (the unroutable address returns an immediate network error here instead of timing out — environment-dependent; this PR does not touch meow-dns, and the behavior is fixed by #387).

crates/meow-proxy/src/mux/muxcool.rs
```diff
@@ -0,0 +1,2083 @@
...
+                        bytes,
+                        dest: std::mem::take(&mut state.dest),
+                    })
+                }
+            };
+            state.reset_header();
+            match event {
+                Some(event) => {
+                    let tx = {
+                        let map = streams.lock().await;
```
madeye: **[correctness — data loss]** The Data-phase delivery breaks the cancel-safety invariant documented at lines 552-555. By this point the payload has been moved out of persistent state (`std::mem::take(&mut state.buf)` at L749) and `state.reset_header()` (L759) has already run — so the frame now lives only in `read_step`'s locals. `read_step` is a `tokio::select!` branch recreated each driver iteration: if this `streams.lock().await` is Pending (the map is locked from other tasks by `open_stream_parts` at L442/444 on every stream open) and `out_rx.recv()` wins the race, the future is dropped and the payload is silently lost — framing stays in sync, so nothing errors; N bytes just vanish mid-stream (truncated HTTP bodies, TLS-in-mux MAC failures).

The STATUS_END path is safe because it locks *before* `reset_header`. The fix is the same ordering here: acquire the lock (or buffer the frame back into `state`) before consuming/resetting, so cancellation loses nothing.

crates/meow-proxy/src/vless/conn.rs
```diff
@@ -145,11 +240,60 @@ impl AsyncRead for VlessConn {
...
+            match Pin::new(&mut this.inner).poll_write(cx, &pending[*pos..]) {
+                Poll::Pending => return Poll::Pending,
+                Poll::Ready(Err(e)) => {
+                    this.write_pending = None;
+                    return Poll::Ready(Err(e));
+                }
+                Poll::Ready(Ok(n)) => {
+                    *pos += n;
+                    if *pos < pending.len() {
+                        return Poll::Pending;
```
madeye: **[correctness — regression for non-mux Vision users]** Two partial-write bugs in the new deferred-header path (both reachable by plain `flow: xtls-rprx-vision` configs, which `dial_tcp` now routes through `new_deferred`):

1. **Lost waker (this line):** after the inner stream returns `Ready(Ok(n))` with `*pos < pending.len()`, this arm returns `Poll::Pending` — but the inner returned Ready, so no waker was registered. Nothing ever wakes the task again; the connection hangs until idle timeout. The drain arm needs to loop and re-poll the inner stream instead of returning Pending after partial progress.

2. **`Ok(0)` → WriteZero (L288):** on a partial write of the combined `[header][payload]` buffer, `n.saturating_sub(header_len)` returns `Ok(0)` for a non-empty caller buf whenever the write ends inside the ~26-byte header. Tokio's `write_all`/`copy` treat `Ok(0)` as `ErrorKind::WriteZero` and abort a connection that was making progress.

Both fall out of the same fix: keep polling until either the header is fully flushed plus ≥1 payload byte written (then return the payload count), or the inner stream returns Pending (waker registered).

crates/meow-proxy/src/vless_adapter.rs
```diff
@@ -164,24 +298,63 @@ impl ProxyAdapter for VlessAdapter {
...
+            // sent: xray expects the VLESS request inside the first
+            // Vision-padded record (mihomo wires it the same way), so the
+            // header is deferred to the first write through the Vision
+            // layer.
             #[cfg(feature = "vless-vision")]
-            Some(VlessFlow::XtlsRprxVision) => Ok(Box::new(VisionConn::new(conn, self.uuid_bytes))),
-            _ => Ok(Box::new(StreamConn(Box::new(conn)))),
-        }
+            Some(VlessFlow::XtlsRprxVision) => {
+                let vless = VlessConn::new_deferred(
```
madeye, line 293: **[correctness — regression for non-mux Vision users]** Deferring the request header to first write breaks server-first protocols. Pre-PR, `dial_tcp` sent the VLESS request header eagerly at dial for all flows; now the Vision path uses `new_deferred`, and the header is only ever emitted from `poll_write` — `poll_read` (conn.rs:170-238) and `poll_flush` (conn.rs:299-301) never touch `header_pending`.

A Vision connection to SMTP/IMAP/FTP/MySQL (server speaks first): the app writes nothing, the server never receives the request so never dials the destination → mutual deadlock until idle timeout. sing-box/xray flush the handshake with empty payload on first Read for exactly this case; `poll_read` here should do the same (drain `header_pending` before reading).

crates/meow-proxy/src/mux/smux.rs
```diff
@@ -0,0 +1,490 @@
...
+            match this.rx.poll_recv(cx) {
+                Poll::Ready(Some(chunk)) => {
+                    let n = chunk.len().min(buf.remaining());
+                    buf.put_slice(&chunk[..n]);
+                    if n < chunk.len() {
+                        this.pending = chunk.slice(n..);
+                    }
+                    Poll::Ready(Ok(()))
+                }
+                Poll::Ready(None) => {
```
madeye: **[correctness — silent truncation]** Session death is indistinguishable from a clean server FIN. `mark_dead()` (L191-195) clears the streams map, dropping every per-stream sender; this `Poll::Ready(None) => { eof = true; Ready(Ok(())) }` arm then reports clean EOF — the identical path as a legitimate `CMD_FIN` (`streams.remove` at L212). A relay mid-transfer on a crashed physical connection sees a clean EOF and reports a truncated 100 MB download as *success*.

There's no dead-session check anywhere in the read path (only `send_frame` checks `dead`, L243). Compare muxcool's `poll_read`, which returns an error when `session.aborted()` — smux needs the equivalent: on `Ready(None)`, return an error if the session died rather than EOF.

crates/meow-proxy/src/mux/smux.rs
```diff
@@ -0,0 +1,490 @@
...
+    }
+
+    /// Start an smux client session over the IO.  A reader task owns the
+    /// read half; a writer task serialises outbound frames.
+    pub fn client<S>(io: S) -> io::Result<Self>
+    where
+        S: AsyncRead + AsyncWrite + Unpin + Send + 'static,
+    {
+        let (mut reader, mut writer) = tokio::io::split(io);
+        let (writer_tx, mut writer_rx) = mpsc::unbounded_channel::<Frame>();
```
madeye: **[correctness — unbounded memory]** smux has no flow control in either direction:

- **Outbound:** the writer channel is `mpsc::unbounded_channel::<Frame>()` (here) and `poll_write` never returns Pending — `send_push` enqueues and returns `Ready(Ok(buf.len()))` unconditionally (L321-332). A fast local uploader through a slow WAN link accumulates frames without bound → OOM.
- **Inbound:** per-stream channels are also unbounded (L169) and the reader task pushes every PSH (L206) regardless of consumer speed, so a fast server + stalled local consumer grows memory the same way.

Real smux (xtaci/sagernet) still bounds memory without UPD frames via `MaxReceiveBuffer` token accounting that pauses socket reads, pushing backpressure into TCP. h2mux (h2 windows) and yamux already have backpressure — smux is the outlier here, and it also voids the relay-backpressure assumptions the rest of the codebase maintains. Bounded channels + a byte cap that pauses the reader would close both directions.

crates/meow-proxy/src/mux/muxcool.rs
```diff
@@ -0,0 +1,2083 @@
...
+            retry_parked(&mut parked);
+            tokio::select! {
+                _ = done.notified() => {
+                    return Err(io::Error::other("mux.cool session closed"));
+                }
+                frame = out_rx.recv() => {
+                    let Some(frame) = frame else {
+                        return Err(io::Error::other("mux.cool outbound channel closed"));
+                    };
+                    conn.write_all(&frame).await?;
```
madeye: **[correctness — session wedge]** `conn.write_all(&frame).await` runs inline inside the select arm with no timeout and no concurrent read progress. If the peer stops draining while still sending bulk data (both TCP buffers full), `write_all` parks forever: the driver never demuxes inbound frames again, keepalive blocks once the 64-slot channel fills, and `done.notified()` can't fire while live streams hold session Arcs — every multiplexed stream on the session wedges permanently until the kernel kills the connection.

h2mux/yamux avoid this because their drivers interleave read and write progress in one state machine. The muxcool driver needs the write to be a resumable state polled alongside `read_step` (or at minimum a write timeout that aborts the session), rather than an inline await that starves the read side. (Related, minor: draining `out_rx.try_recv()` into one buffer before a single flush would also coalesce the current one-syscall-per-frame pattern.)

madeye (commented):
Thanks @aimkiray — this is a substantial, well-documented PR, and the interop matrix + regression tests clearly took real care. The sing-mux request/response framing, the muxcool wire codec, and the flow-control parking design all check out against the upstream sources, and the muxcool cancel-safety design note (lines 552-555) is exactly the right discipline — most of the serious findings below are places where an implementation detail escapes that discipline rather than design problems.

I ran a deep review (8 analysis passes + per-finding verification against upstream sing-mux/sing-box/mihomo sources). **12 verified correctness findings are attached as inline comments** — the highest-severity ones are the muxcool Data-phase cancellation data loss, the three regressions for existing non-mux `xtls-rprx-vision` users from the deferred-header refactor, and the smux flow-control/teardown gaps.

Also verified but kept out of the inline comments (cleanup-level, no behavior change required to merge):

- **h2mux duplicates meow-transport's H2 machinery**: `mux/h2mux.rs`'s `poll_read`/`poll_write` (incl. the pending-write stash) are near line-for-line `meow-transport/src/h2.rs::H2Stream`, and its local `RecvState` re-solves the issue-#377 lazy-response deadlock that `h2_common::RecvState` already solves with a different mechanism. A future h2 fix now has to land twice; consider making the meow-transport pieces `pub` and reusing them.
- **Mux plumbing copy-pasted across all four adapters**: identical `mux`/`mux_only_tcp` fields, `with_mux` builders, byte-identical `support_udp()` blocks (trojan.rs:459, shadowsocks_adapter.rs:512, vless_adapter.rs:246, vmess/mod.rs:197), and near-identical dial-intercept branches. A generic decorator (or shared helpers on `MuxClient`) would make protocol #5 a one-liner and keep the only-tcp/host-extraction logic from drifting.
- **smux per-write double copy**: `send_push` (smux.rs:264) copies into a `Bytes`, then the writer task copies again via `encode_into`'s `put_slice` (smux.rs:52) — one alloc + two memcpys per relay chunk; owned-chunk pass-through or vectored writes would drop one copy.
- **packet.rs per-datagram alloc**: `write_datagram` (packet.rs:66) allocates and copies the whole payload to prepend 2 length bytes; two writes under the existing writer mutex are safe (the underlying streams are byte-oriented) and avoid the alloc.
- **muxcool one-flush-per-frame**: the driver writes+flushes each outbound frame individually (muxcool.rs:816-817); draining `out_rx.try_recv()` into one buffer before a single flush would coalesce syscalls/TLS records under concurrency.

Two findings I checked and did *not* flag, for the record: the h2 send-beyond-granted-capacity pattern in h2mux is bounded by h2's own capacity accounting (and pre-exists in meow-transport), and the REALITY session-id version-byte change (1.8.2→1.8.1) matches what sing-box's `reality_client.go` actually advertises today — correct interop fix.

crates/meow-proxy/src/mux/client.rs
```diff
@@ -0,0 +1,633 @@
...
+
+    /// Dial a fresh physical connection and start a new session.  sing-mux
+    /// sessions write the mux request header on top; Mux.Cool needs none —
+    /// the dialer's VLESS CommandMux request already marks the connection.
+    /// Callers must hold the sessions lock.
+    async fn offer_new_locked(
+        self: &Arc<Self>,
+        sessions: &mut VecDeque<Arc<MuxSession>>,
+    ) -> Result<Arc<MuxSession>> {
+        let mut conn = (self.dial)().await?;
```
madeye, line 547: **[correctness — head-of-line blocking]** `offer()` holds the sessions mutex across this physical dial, and no timeout exists anywhere in the chain: the adapter dial closures → `connect_tcp_host` → TLS handshake are all unbounded, and the tunnel path is a bare `proxy.dial_tcp(&metadata).await` (meow-tunnel/src/tcp.rs:110).

If the node black-holes (SYN drops, tarpit): the first open holds the lock for the full OS SYN-retry duration (~1-2 min), and every other `open_stream`/`open_packet_stream` on the adapter — including the URLTest probes that would mark the node unhealthy — queues behind the mutex, so failover can't trigger. Pre-mux, each connection dialed independently and its caller's timeout bounded the damage.

sing-mux holds its lock across `offerNew` the same way, but wraps the dial in `TCPTimeout` — the port dropped the timeout. Wrapping `(self.dial)()` in a `tokio::time::timeout` here restores the bound while keeping the no-overshoot serialization.

crates/meow-proxy/src/mux/smux.rs
```diff
@@ -0,0 +1,490 @@
...
+            dead: AtomicBool::new(false),
+        });
+
+        // Writer task: drain frames, write full wire frames in order.  The
+        // encode buffer is reused across frames (hot path — avoid one
+        // allocation per frame).
+        let writer_inner = Arc::clone(&inner);
+        tokio::spawn(async move {
+            let mut buf = BytesMut::new();
+            while let Some(frame) = writer_rx.recv().await {
```
madeye: **[correctness — task + fd leak]** Evicted/dead smux sessions leak the writer task and the physical socket forever. The writer task captures `writer_inner = Arc::clone(&inner)` (L113), and `SessionInner` holds `writer_tx` (L76) — so the task's own Arc keeps a sender alive and `writer_rx.recv().await` can never return `None`; the loop exits only on a write error. There's no `Drop` impl, no abort handle, and `mark_dead()` only clears the streams map.

`offer()` evicts idle sessions after 60s by simply dropping the pool's Arc, so every idle-evicted session (routine with URLTest probes recreating sessions each cycle) permanently pins one tokio task + the WriteHalf fd — accumulating to fd exhaustion over hours. The writer even self-sustains after the server closes and the reader exits.

yamux gets this right: dropping `Session` drops `open_tx` and the driver exits via `None => return Poll::Ready(())` (yamux.rs:64). smux needs an equivalent teardown signal (e.g. keep `writer_tx` outside `SessionInner`, or an explicit close on eviction).

crates/meow-proxy/src/mux/packet.rs
```diff
@@ -0,0 +1,464 @@
...
+        writer.write_all(&frame).await.map_err(MeowError::Io)?;
+        writer.flush().await.map_err(MeowError::Io)?;
+        Ok(data.len())
+    }
+
+    /// Read one datagram: `[len u16 BE][data]`.  The stream response status
+    /// byte is consumed lazily on the first read by `MuxStream`.
+    pub async fn read_datagram(&self, buf: &mut [u8]) -> Result<usize> {
+        let mut reader = self.reader.lock().await;
+        let mut len_buf = [0u8; 2];
```
madeye: **[correctness — UDP framing desync, TUN path]** `read_datagram` is two sequential `read_exact` awaits (length prefix, then payload) with no persisted partial-frame state — it is not cancellation-safe. The TUN UDP pump races it in a `tokio::select!` and recreates the future each iteration (tun/udp.rs:246): if the outbound branch wins after the 2 length bytes were consumed, those bytes are lost and the next read interprets payload bytes as a length — every subsequent datagram on the flow is garbage. The pump's comment "may drop one datagram — acceptable under UDP delivery semantics" (tun/udp.rs:229-230) holds for real UDP sockets, not for length-prefixed stream framing.

With mux enabled, `support_udp` is now true for SS/Trojan/VMess/VLESS, so QUIC/DNS over TUN rides exactly this path. SOCKS5 UDP is unaffected (dedicated task, no select), and muxcool's `PacketConn` is safe by construction (driver-owned state + cancel-safe mpsc recv) — the fix is to persist partial-frame state in `MuxPacketConn` (small state machine like muxcool's `ReadState`) or give the TUN pump a dedicated read task.

crates/meow-config/src/proxy_parser.rs
```diff
@@ -1375,9 +1395,139 @@ fn parse_vless(
...
+        feature = "vless",
+        feature = "ss",
+        feature = "vmess"
+    )
+))]
+fn parse_mux_options(
+    name: &str,
+    config: &HashMap<String, serde_yaml::Value>,
+) -> std::result::Result<Option<meow_proxy::mux::MuxOptions>, String> {
+    let Some(mux_cfg) = config.get("mux") else {
```
madeye: **[correctness — mihomo parity]** mihomo's per-proxy key for sing-mux is `smux:`, not `mux:` (`adapter/outbound/singmux.go` — `SingMuxOption`, wired via `mapping["smux"]` in adapter parsing). `parse_mux_options` only ever reads `config.get("mux")` (here, and the no-mux-feature warn path at L1517), so a real mihomo config or subscription using `smux:` silently gets no multiplexing — the proxy is built unmuxed with no warning, and the divergence is invisible until the node is a sing-box inbound with multiplex *required*, where connections fail confusingly.

Suggest accepting `smux` as an alias of `mux` (and perhaps warning if both are present) so migrated configs behave as the PR intends.

crates/meow-proxy/src/mux/muxcool.rs
```diff
@@ -0,0 +1,2083 @@
...
+
+    /// Client-assigned ids start at 1 and increase by 1
+    /// (xray common/mux/session.go::Allocate).  Once exhausted the session
+    /// keeps failing: ids never wrap around, because a recycled id would
+    /// collide with a live stream's frames.  The pool dials a fresh session
+    /// for later opens; existing streams keep working.
+    fn alloc_sid(&self) -> io::Result<u16> {
+        let prev = self.next_sid.fetch_add(1, Ordering::SeqCst);
+        if prev > u16::MAX as u32 {
+            return Err(io::Error::other("mux.cool session id space exhausted"));
```
madeye, line 497: **[correctness — eventual persistent failure]** `alloc_sid` errors forever once the u16 sid space is exhausted (no wraparound/reuse of freed sids), but it never sets `closed` — so `aborted()`/`is_dead()` stay false and the pool's `retain` keeps the session. The doc comment above promises "the pool dials a fresh session for later opens", but no code implements that: an exhausted session ends at 0 live streams, `offer()` returns a zero-stream session unconditionally (client.rs:401-403), and `open_stream_flags` retries only twice — so after ~65k cumulative streams on one physical connection, every new open fails for up to 60s (idle eviction), or indefinitely if the session still holds a long-lived stream.

Simplest fix: mark the session closed/aborted on sid exhaustion so `offer()` prunes it and dials fresh — which also makes the doc comment true.

crates/meow-proxy/src/shadowsocks_adapter.rs
```diff
@@ -371,107 +508,72 @@ impl ProxyAdapter for ShadowsocksAdapter {
...
+                ));
+            };
+            let conn = mux.open_stream(&host, metadata.dst_port).await?;
+            return Ok(Box::new(conn));
+        }
+
+        self.core.dial_tcp_stream(addr).await
+    }
+
+    async fn dial_udp(&self, metadata: &Metadata) -> Result<Box<dyn ProxyPacketConn>> {
```
madeye, line 533: **[build hygiene]** `metadata` is unused when built with `ss` but without `mux` — reproduced with `cargo check -p meow-proxy --no-default-features --features ss`:

```
warning: unused variable: `metadata` --> crates/meow-proxy/src/shadowsocks_adapter.rs:548:30
```

The `minimal` bundle is exactly this combo (`minimal = ["ss", "trojan", ...]`, mux excluded per ADR-0007), so any `-D warnings` build of it fails. Needs `#[cfg_attr(not(feature = "mux"), allow(unused_variables))]` or an underscore rebinding inside the non-mux path.

aimkiray:
## Reply to the PR #394 review (posted)

Thanks for the thorough review! All comments are addressed — the first-round
themes below, plus a final deep-review round (4 independent passes) and a
post-fix performance/interop verification. The fixes are pushed to
`aimkiray:muxcool` (this PR); the branch was rebased onto the latest main after
upstream advanced, and the harness hardening was split out as PR #399.

**1. Cancellation safety**
- Mux.Cool reader: length-prefix read progress now survives a cancelled
  `read_step` (persistent `pos` + pre-sized buffer, stream-map entry acquired
  before frame state is consumed). Tests:
  `length_prefix_read_survives_cancellation`,
  `data_delivery_survives_cancellation_while_stream_map_is_locked`.
- Mux.Cool writer: partial writes persist in a `WriteState` with frame
  coalescing (no stored-future duplication). Test:
  `blocked_physical_write_does_not_starve_reads`.
- sing-mux packet reader: datagram length + payload progress survive a
  dropped read future. Test:
  `cancelled_read_after_length_prefix_resumes_payload`.

**2. Backpressure / flow control**
- smux outbound is a bounded channel (64 frames) with `PollSender`
  backpressure and partial-write progress per the AsyncWrite contract.
  Test: `outbound_queue_applies_backpressure`.
- smux inbound: 4 MiB session receive budget via semaphore permits released
  as chunks are consumed or dropped.
- Mux.Cool keeps its parked-frame read pause: inbound delivery never blocks
  the driver.

**3. Session lifecycle / cleanup**
- smux: `Session::drop` cancels both IO tasks and releases the socket.
  Test: `dropping_idle_session_releases_physical_io`. Physical EOF now
  surfaces as `BrokenPipe` instead of looking like a clean FIN. Test:
  `physical_eof_is_reported_as_an_error`.
- Mux.Cool: sid-space exhaustion retires the session (the pool replaces it)
  without aborting existing streams. Test:
  `sid_space_exhaustion_errors_without_reuse`.
- h2mux: EOS sent on stream drop; setup bounded by the 5 s TCPTimeout; the
  pool prunes dead sessions and evicts zero-stream sessions idle past 60 s,
  and a stalled dial fails after 5 s. Test:
  `session_setup_timeout_bounds_a_stalled_dial`.
- yamux driver exits when its request channel closes.

**4. VLESS deferred header**
- Rewritten as `poll_write_deferred`: survives 1-byte partial writes and
  flushes before the first read so server-first protocols (SMTP/IMAP/MySQL)
  are not deadlocked. Tests: `deferred_header_handles_one_byte_partial_writes`,
  `deferred_header_is_flushed_before_server_first_read`.

**5. Deduplication / cleanup**
- The raw h2 stream moved to `meow_transport::h2_common::H2Stream`, shared by
  the h2 transport and h2mux (h2mux keeps its RST_STREAM(NO_ERROR)-as-EOF
  behavior behind a builder flag).
- The four adapters share `open_stream_for` / `open_packet_stream_for`; the
  duplicated `mux_only_tcp` field is gone in favor of `supports_udp()`.

**6. mihomo config parity**
- Canonical `smux:` key now accepted (mihomo's actual key), `mux:` kept as a
  legacy alias, and setting both warns and prefers the canonical `smux:`;
  malformed field types (e.g. `enabled: "true"`) reject the node instead
  of silently downgrading. Docs updated.
- muxcool remains a meow-rs superset (VLESS/VMess-only, rejected for ss/
  trojan with a clear message).

**7. Second review pass (reservations, h2 flow control, cancellation)**
- Pool: each open reserves a slot under the pool lock (pending counter) so
  concurrent offers cannot overshoot max-streams / min-streams; the
  reservation is released on cancellation. Test:
  `concurrent_opens_respect_max_streams`.
- h2: `poll_ready` before `send_request` in both h2mux and the h2 transport;
  `poll_write` sends only the flow-control window actually granted (real
  backpressure). h2mux session drop aborts the connection driver task.
- VLESS UDP: the plain path has the same persistent, cancellation-safe
  reader as the mux packet conn. Tests:
  `udp_read_survives_cancellation_mid_response`,
  `udp_read_survives_cancellation_mid_payload`.
- smux: `poll_flush` round-trips a Flush request through the FIFO writer
  channel; cancelled opens remove their stream-map entry.
- Mux.Cool shutdown claims the End only after it is enqueued - a shutdown
  cancelled on a full channel is re-sent by the EndGuard. Test:
  `cancelled_shutdown_still_sends_end_on_drop`.
- Zero-capacity reads return `Ready(Ok(()))`; overlong FQDNs are rejected;
  mux-only symbols are feature-gated (the no-mux build is clippy-clean).

**8. Final deep-review round (4 independent passes: driver, pool/adapters,
transport/docs, PR hygiene) — all findings fixed**
- Mux.Cool teardown is End-echo independent: sing-box's Mux.Cool server
  closes streams silently and never echoes End frames, so the client removes
  the stream-map entry on shutdown and drop (inbound channel closes -> EOF)
  instead of waiting for an echo that never arrives. This was the root cause
  of a churn collapse (24/s) that the benchmark matrix exposed; after the fix
  the data the server already sent is delivered before EOF, and Mux.Cool End
  remains a full close on the wire (no TCP-style half-close, matching xray).
  Tests against a silent mock server:
  `shutdown_eofs_read_side_without_server_end_echo`,
  `drop_removes_map_entry_without_server_end_echo`,
  `stream_churn_leaves_no_entries_behind` (silent server),
  `shutdown_delivers_buffered_data_then_eofs`.
- smux: zero-length PSH frames are skipped instead of reading as a
  premature EOF. Test: `zero_length_psh_frame_does_not_read_as_eof`.
- reality_tls: `poll_write` propagates drain errors instead of reporting
  `Ok` on a failed inner write (both branches); the REALITY auth payload
  comments now name the ClientVer triple correctly.
- vision: empty-buffer guard on `poll_write`.
- muxcool `poll_write` surfaces an error once the session aborted (no
  Ok(n)-then-drop window during driver teardown).
- EndGuard.ended is a plain bool (was an unshared Arc<AtomicBool>).

**Performance & interop (local sing-box 1.13.18, meow-bench, 64 workers)**
- connrate (tls+vision muxcool): 24/s before the teardown fix -> 745/s
  (10 s window; xray-core 634/s) and ~1380/s in a 6 s window;
  reality+vision / reality / h2mux rows ~1320/s each.
- latency: p50 704us / p99 1.1 ms (xray-core 2.1/2.8 ms).
- bulk 64 MB throughput is fixed separately by the 32 KiB relay-buffer PR
  (see the split below), not by this PR.
- Physical Android device (arm64): release APK with a real (anonymized)
  subscription - 5 nodes incl. Reality+Vision + muxcool - connects
  end-to-end: tun0 up, browsing/HTTPS/downloads all multiplexed over the
  mux sessions to the real node.

**PR split (per review feedback)**
- `fix/bench-harness` (4 commits): meow-bench phase deadlines, --binary-arg,
  echo-timeout reporting, Windows teardown. Land first.
- `fix/relay-buffers` (1 commit): tunnel 32 KiB relay buffers (generic bulk
  throughput fix for every protocol, not mux-specific). Land second.
- This PR rebases onto them and drops those commits.

**Verification (current tree)**
- meow-proxy: 375 passed / 3 ignored (43 muxcool, 12 smux); meow-config
  149; meow-tunnel 62; meow-transport 4.
- `cargo fmt` + three-way clippy (`-D warnings`) clean.
- Local sing-box interop matrix: 28/28 combos green (vless x {plain, tls,
  reality, reality+vision} x {smux, yamux, h2mux, muxcool}, trojan/ss/vmess
  sing-mux, vmess muxcool; each: TCP h2 + h1.1 + UDP echo + 12-way
  concurrent requests).
- meow-bench memleak: h2mux / smux / muxcool all OK; RSS and handle counts
  return to baseline after idle session eviction.

Known unchanged-by-design (matching mihomo/sing-mux): `max-streams` is a
threshold only when `max-connections=0`; the pool dials under the lock to
prevent connection overshoot; muxcool PacketConn teardown is Drop-based
(close() is a no-op, same as the sing-mux packet conn).

madeye:
hmm, this PR becomes too large to review now. can you split it into multiple small ones? For example, you can open PRs for each mux implementation  one by one.

aimkiray:
> hmm, this PR becomes too large to review now. can you split it into multiple small ones? For example, you can open PRs for each mux implementation one by one.

Good call, thanks for flagging it. I'll break this out into separate PRs, one per mux.

aimkiray:
Closing in favor of the stacked split: this large change was reorganized into six smaller, individually reviewable PRs (sing-mux smux → SS/VMess → yamux → h2mux → Mux.Cool), tracked under an umbrella issue. The rebased and reviewed versions are being submitted as a stack.

PR #399, opened 2026-08-16
bench: bound every connection phase; drive non-meow binaries; report echo timeouts

aimkiray:
# bench: bound every phase, drive non-meow binaries, report echo timeouts

## Overview

meow-bench hardening so a wedged proxy can never hang the harness and cross-binary
comparisons stay apples-to-apples:

- Every connection phase has a deadline: socks5 connect (10s), warmup (30s cap), and the
  connrate echo phase (10s). Previously a proxy that accepted CONNECT but never returned
  the echo blocked a worker in read_exact forever and the whole run hung.
- `--binary-arg` (default `-f`, xray uses `-c`) is reused by both the SOCKS5 and the DNS
  child so the harness can drive xray-core/mihomo binaries directly.
- `ConnRateResult` gains `echo_timeouts` so a proxy that stalls mid-connection surfaces as
  a count instead of an unexplained floor rate.
- Windows teardown: reap children via `child.kill()` on every path (ChildGuard), the
  memleak proxy included; earlier leaked children held the port and poisoned later runs.
- Spawned proxy stderr is inherited so protocol diagnostics land in the run log.

## Commits

- `d6a3ace` bench: terminate the memleak proxy via child.kill on Windows
- `11b49ae` bench: reap proxy children on every path; add workload context
- `0991c5c` bench: bound every connection phase so a stalled proxy never wedges the harness
- `cd7419f` bench: reuse --binary-arg for the DNS child; report echo timeouts

### DIFF
--- a/crates/meow-bench/src/bench_connrate.rs
+++ b/crates/meow-bench/src/bench_connrate.rs
@@ -16,6 +16,9 @@ pub struct ConnRateResult {
     pub duration_secs: f64,
     pub total_connections: u64,
     pub connections_per_sec: f64,
+    /// Connections whose echo phase hit the deadline (a proxy that
+    /// accepted the CONNECT but never answered).
+    pub echo_timeouts: u64,
 }
 
 pub async fn bench_conn_rate(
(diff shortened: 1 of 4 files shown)


aimkiray:

Verified every finding in the review against this PR's code and pushed targeted fix-ups. All commits are fast-forward pushes to the PR head; no history rewritten.

### Findings verified & fixed (`ae9cad7`)

- **M7 (confirmed, fixed):** termination went straight from SIGTERM to SIGKILL, giving no graceful-exit window. Added `SHUTDOWN_GRACE` (5 s): SIGTERM → poll `try_wait()` every 100 ms (non-blocking, workers are never stalled) → `kill()` + `wait()` only as the last resort. All three stop sites share `shutdown_child`.
- **M8 (confirmed, fixed):** `std::mem::forget` leaked the `Child` (an OS handle on Windows), and the memleak-benchmark path held a bare `Child` with no cleanup at all. Replaced with `ChildGuard(Option<Child>)`: `shutdown()` performs the graceful stop and disarms the guard so `Drop` is a no-op on the success path; the error paths now go through the guard too.

### Round-2 re-review (`dc9982f`, docs only)

- **N6 (confirmed, accepted as-is):** `ChildGuard::drop` uses a blocking `wait()`, but `kill()` precedes it so the reap returns essentially immediately — and `Drop` only runs on error paths now that `shutdown()` disarms. Documented in code, including the bounded worst case: a SIGTERM-ignoring target costs up to `SHUTDOWN_GRACE` (5 s) per target, sequential across targets.

### Validation

- `cargo clippy -p meow-bench --all-targets` clean; the crate exposes no unit tests (bench binary). Shutdown paths were exercised manually via the bench harness.

Issue #412, opened 2026-08-18
Track stacked mux implementation split from PR #394


aimkiray:
## Background

PR #394 introduced connection multiplexing for VLESS, Trojan, Shadowsocks, and VMess, including sing-mux protocols and Xray Mux.Cool.

Because #394 is a large pull request rather than an issue, this issue is the long-lived tracking item for splitting, reviewing, and landing that work. It is not a replacement for the design discussion in #394.

The bench harness changes are tracked separately in PR #399 and are intentionally outside the six mux PRs below.

## Goal

Land the mux implementation as a reviewable stacked series with one focused protocol or prerequisite per pull request.

The final result should preserve the complete mux functionality from the rebased #394 implementation while keeping unrelated benchmark harness changes separate.

## Stacked pull requests

The branches are prepared locally in this order:

- [ ] PR1: `pr1-vless-deferred` — VLESS deferred Vision header and REALITY prerequisite
- [ ] PR2: `pr2-mux-smux` — sing-mux scaffold and smux for VLESS/Trojan
- [ ] PR3: `pr3-mux-ss-vmess` — sing-mux for Shadowsocks and VMess
- [ ] PR4: `pr4-mux-yamux` — yamux sing-mux protocol
- [ ] PR5: `pr5-mux-h2mux` — h2mux and shared HTTP/2 stream
- [ ] PR6: `pr6-mux-muxcool` — Xray Mux.Cool for VLESS and VMess

Stack order:

```text
main
  -> pr1-vless-deferred
  -> pr2-mux-smux
  -> pr3-mux-ss-vmess
  -> pr4-mux-yamux
  -> pr5-mux-h2mux
  -> pr6-mux-muxcool
```

Each pull request should use the previous branch as its base so reviewers see only the incremental layer.

## Acceptance criteria

- [x] Each stacked PR has a focused description and review scope.
- [x] Each branch passes formatting, three-way clippy, locked workspace check, and relevant tests; the final stack passes the workspace documentation warning check.
- [x] The final stack preserves the mux code, configuration behavior, tests, and documentation from the rebased #394 implementation (tree-equivalent to the rebased source).
- [x] `meow-bench` harness changes remain in #399 and are not duplicated in the mux stack.
- [x] The cleaned `Cargo.lock` contains only the dependency changes required by the mux implementation, without unrelated registry version churn.
- [x] The stack is rebased onto current `main` before submission.
[...]

PR #413, opened 2026-08-18
fix(vless): defer Vision request header and fix REALITY auth

aimkiray:
### Description

This is the first layer of the mux work from #394. It isolates the VLESS and REALITY transport prerequisites from the protocol implementations that follow.

VLESS Vision connections now defer the request header until the first application write. Deferred response handling is cancellation-safe across partial writes and cancelled polls. REALITY authentication payload construction is corrected, and oversized writes are chunked without violating the transport write-size invariant.

This PR contains no mux protocol implementation.

This branch is the base of the stack:

- Base: `main`
- Next: `pr2-mux-smux`

Given a Vision connection whose request body has not been written, when the connection is established, the complete VLESS request header is deferred until application data is written.

Given a deferred header write is cancelled and retried, the next poll does not duplicate the header or associate it with the wrong payload.

Given concurrent REALITY reads and writes, oversized application data is split into valid writes while the authentication payload remains intact.

The branch was also checked with repository formatting, all three clippy feature modes, a locked workspace check, affected crate tests, and configuration integration tests before preparing this stack entry.

### Key points

The deferred header behavior is kept in the prerequisite PR because every later VLESS mux implementation relies on the same first-write and response lifecycle. Keeping it separate makes the mux protocol reviews smaller and avoids mixing transport correctness with protocol dispatch.

### Notes for reviewers

Start with the VLESS connection lifecycle and the REALITY payload change. The following stacked PR adds the shared mux client and smux; yamux, h2mux, and Mux.Cool are intentionally not part of this diff.

This branch is based on the current `origin/main`, not the historical base of #394.

### Authorship and follow-up

Who wrote this:

- [ ] A human wrote it.

Who answers review comments:

> Note: this PR also bumps `h2` 0.4.14 → 0.4.16 in `Cargo.lock` to clear advisory [RUSTSEC-2026-0258](https://rustsec.org/advisories/RUSTSEC-2026-0258) (unbounded empty DATA frames, DoS; patched in 0.4.16). h2 0.4.14 is inherited from `main` via `meow-transport`'s h2 transport feature, so the bump sits at the stack base and clears the security-audit for every layer. No code change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,7 +118,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -129,7 +129,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -1177,7 +1177,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -1506,9 +1506,9 @@ checksum = "0cc23270f6e1808e30a928bdc84dea0b9b4136a8bc82338574f23baf47bbd280"
 
 [[package]]
 name = "h2"
-version = "0.4.14"
+version = "0.4.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "171fefbc92fe4a4de27e0698d6a5b392d6a0e333506bc49133760b3bcf948733"
+checksum = "a9f37a958b41b3b19ee2707c06439c0e9e547e847223eb791ecb0cb821c65e27"
 dependencies = [
  "atomic-waker",
  "bytes",
@@ -1773,7 +1773,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.3",
+ "socket2 0.5.10",
  "tokio",
  "tower-service",
  "tracing",
@@ -1992,7 +1992,7 @@ checksum = "3640c1c38b8e4e43584d8df18be5fc6b0aa314ce6ebf51b53313d4306cca8e46"
 dependencies = [
  "hermit-abi",
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -2917,7 +2917,7 @@ version = "0.50.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -3298,7 +3298,7 @@ dependencies = [
  "quinn-udp",
  "rustc-hash 2.1.2",
  "rustls",
- "socket2 0.6.3",
+ "socket2 0.5.10",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
@@ -3335,7 +3335,7 @@ dependencies = [
  "cfg_aliases",
  "libc",
  "once_cell",
- "socket2 0.6.3",
+ "socket2 0.5.10",
  "tracing",
  "windows-sys 0.60.2",
 ]
@@ -3672,7 +3672,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.12.1",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -4052,7 +4052,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3a766e1110788c36f4fa1c2b71b387a7815aa65f88ce0229841826633d93723e"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


aimkiray:

https://github.com/madeye/meow-rs/compare/main...aimkiray:pr1-vless-deferred

This layer: 2 commits / 6 files — the VLESS deferred-header fix plus a one-line `Cargo.lock` bump of h2 0.4.14 → 0.4.16 clearing RUSTSEC-2026-0258. See #412 for the stack table and merge order.

madeye:
## Review — PR 1/6 of the mux stack

Reviewed as the incremental diff against `origin/main`. Stack ancestry verified linear (413 → 414 → 415 → 416 → 417 → 418). Security Audit is green on all six branches, so the `h2 0.4.16` bump at the stack base does its job.

The deferred-header state machine is the good part: `poll_write_deferred` / `poll_flush_deferred_header` handle partial header writes, the server-first (SMTP/IMAP/MySQL) path, and read/shutdown ordering correctly, and the new tests cover one-byte partial writes, flush-before-first-read and mid-datagram cancellation. The u16 record-wrap fix in both the REALITY and Vision write paths is a real bug fix.

Four findings.

### 1. `reality_tls.rs:441` — the new `Pending` return introduces a changed-buffer hazard (high)

```rust
if self.write_pending.is_some() {
    let consumed = self.write_pending.as_ref().expect("checked above").consumed;
    return match self.drain_pending_write(cx) {
        Poll::Pending => Poll::Pending,
        Poll::Ready(Ok(())) => Poll::Ready(Ok(consumed)),
        ...
```

Before this PR `poll_write` never returned `Pending` after consuming: it sealed the record, buffered it, and reported `Ok(buf.len())` immediately. That is safe under any caller, because the buffered frame is drained by the next `poll_write` / `poll_flush` / `poll_shutdown`.

The new version returns `Pending` with a sealed record buffered, and on the next poll reports that record's plaintext length — against whatever buffer arrives. `AsyncWrite` does not guarantee the same buffer is presented after a `Pending`. If a write future is dropped while `Pending` and recreated with different data (precisely the stored-write-future pattern this PR's own test comment cites as the source of the stress-era mux corruption), the *old* buffer's bytes go on the wire while the caller is told that many bytes of the *new* buffer were consumed.

Simplest fix, and it keeps the pre-existing semantics:

```rust
match self.drain_pending_write(cx) {
    Poll::Ready(Err(e)) => Poll::Ready(Err(e)),
    _ => Poll::Ready(Ok(chunk_len)),   // buffered; flush/shutdown drain it
}
```

Alternatively add the same changed-buffer rejection that #407 adds for h2. Whichever you pick, please apply it consistently with PR 5 of this stack — `h2_common::H2Stream::poll_write` has the identical shape there (see my comment on #417).

### 2. `vless/conn.rs:377` — `VlessUdpReader` addon discard loses cancellation progress (medium)

```rust
if addon_length > 0 {
    self.payload.resize(addon_length, 0);
    self.payload_pos = 0;          // reset on every re-entry
    while self.payload_pos < self.payload.len() { ... }
```

`header_done` is still `false` while this loop runs, so a `read_packet` future dropped mid-addon re-enters here and resets `payload_pos` to 0 — the resumed call reads `addon_length` bytes again, eating the head of the first datagram and desyncing the UDP-over-TCP framing permanently. This contradicts the invariant the struct doc claims ("the response header, the per-datagram length prefix and the payload all persist their read progress across dropped `read_packet` futures"), and the TCP path in `poll_read` gets it right by counting against the persistent `response_pos`.

```rust
if self.payload.len() != addon_length {
    self.payload.resize(addon_length, 0);
    self.payload_pos = 0;
}
```

Latent in practice — VLESS responses carry `addon_length == 0` — and the new `udp_read_survives_cancellation_*` tests exercise the header and payload phases but not the addon phase.

### 3. `reality_tls.rs:582` — REALITY ClientVer `[1, 8, 2]` → `[1, 8, 1]`: please cite the upstream line

The two comments added here disagree in emphasis. The code comment says sing-box's utls fork validates the exact value; the test comment says xray servers tolerate the previous `[1, 8, 2]` while sing-box validates. This changes bytes on the wire for every existing REALITY user, so a link to the validating code (xray `reality.go` / sing-box's fork) would settle whether any deployment can regress.

### 4. `reality_tls.rs:1090` — `debug_assert!` on the record length (minor)

Release builds silently truncate `record_len as u16`. Now that `poll_write` chunks at 16 KiB the only reachable caller is safe, but returning `TransportError` instead of asserting would make that structural rather than incidental.

aimkiray:
## Review fixes applied

Thank you for the thorough review. All four findings are fixed in this update.

### 1. `reality_tls.rs:441` — changed-buffer hazard (high) ✅

Restructured `poll_write` with a unified invariant: **"never report bytes consumed from a buffer that was not actually read."**

- **Top path** (pending record exists): drain first. If `Pending`, return `Pending` — zero bytes consumed from the incoming buffer, so the caller retries safely. If drained, fall through to seal the new buffer (never report the old record's `consumed` against the new buffer).
- **Bottom path** (after sealing): always return `Ready(Ok(chunk_len))` regardless of whether the drain completes. The sealed record is buffered and drains on the next `poll_write` / `poll_flush` / `poll_shutdown`.

The `consumed` field is removed from `StreamPendingWrite` — it is no longer needed.

This is the "buffered Ready" approach (your simplest fix). It differs from the h2_common guard in #417 because the constraints differ: REALITY seals whole records (no partial offset, encrypted bytes can't be compared against incoming plaintext), so buffering + `Ready` is clean. h2 tracks partial sends with flow control, so the changed-buffer rejection guard is necessary there. Both satisfy the same invariant.

New test: `poll_write_reports_new_chunk_len_after_pending_drain` — a mock inner stream returns `Pending` on the first `poll_write`, then accepts on the second. Verifies the first write returns `Ok(10)`, the second returns `Ok(5)` (the new buffer's length, not the old record's 10).

### 2. `vless/conn.rs:377` — addon discard loses cancellation progress (medium) ✅

Guarded the reset: `payload_pos` is only reset when `payload.len() != addon_length`. A cancelled `read_packet` future re-enters with `header_done` still false, and the existing `payload` (sized to `addon_length`) is reused — progress resumes from `payload_pos` instead of re-reading `addon_length` bytes.

### 3. `reality_tls.rs:582` — ClientVer upstream citation (docs) ✅

Updated the code comment with specific upstream references:
- sing-box: `common/tls/reality_client.go:186-188` — hardcodes `[1, 8, 1]`
- xray-core: `transport/internet/reality/reality.go:143-145` — uses `core.Version_x/y/z` (currently 26, 7, 28)
- Neither server validates these bytes — they are part of the 16-byte AES-GCM auth payload; the server only checks `short_id` and `timestamp` after decryption.

The test comment is updated to match.

### 4. `reality_tls.rs:1090` — `debug_assert!` on record length (minor) ✅

Replaced `debug_assert!` with a `TransportError` return, making the u16 overflow check structural rather than incidental.

### Smoke tests

All 13 smoke tests pass (VLESS×{smux,yamux,h2mux,muxcool}, Trojan×{smux,yamux,h2mux}, SS×h2mux, VMess×{h2mux,muxcool} TCP+8-way concurrency, plus h2mux TCP/UDP and muxcool TCP+UDP live probes against sing-box 1.13.13).

PR #414, opened 2026-08-18
feat(mux): add sing-mux scaffold and smux for VLESS/Trojan

aimkiray:
### Description

This is the second layer of the #394 stack. It adds the shared mux client infrastructure and the first sing-mux protocol, smux, for VLESS and Trojan.

The change introduces the connection pool, session lifecycle, stream handling, UDP packet handling, request/address encoding, and cancellation-safe stream setup. smux uses sing-mux protocol byte `0`. Mux is gated behind the `mux` feature and remains excluded from `minimal` builds.

This stage intentionally keeps smux as the temporary default. Yamux, h2mux, and Mux.Cool are added by later stack entries.

Stack position:

- Base: `pr1-vless-deferred` (`bf83872`)
- Next: `pr3-mux-ss-vmess`

Given a VLESS or Trojan proxy with mux enabled, when several TCP streams are opened, they share one physical session while remaining independently readable and writable.

Given a mux UDP stream, when datagrams with different destinations are sent, packet boundaries and destination addresses are preserved.

Given an unknown mux protocol, when configuration is loaded, the proxy is rejected with a clear warning. Given mux is disabled, the existing non-mux connection path remains active.

The branch was checked with repository formatting, all three clippy feature modes, a locked workspace check, affected crate tests, and configuration integration tests. The lockfile contains only the required mux dependency addition; unrelated registry version churn was removed during review.

### Key points

The pool and stream abstractions are shared by all later protocol implementations. Keeping this PR limited to smux makes the initial lifecycle, backpressure, UDP, and cancellation behavior reviewable before adding additional wire protocols.

The temporary smux default is deliberate. PR5 changes the empty/default protocol to h2mux after all sing-mux protocol dispatch is available.

### Notes for reviewers

Review the pool lifecycle and stream/packet boundaries first, then the VLESS/Trojan adapters. This PR is based on PR1 and should be reviewed as the incremental diff against `pr1-vless-deferred`, not against `main`.

Yamux, h2mux, and Xray Mux.Cool are not supported by this branch yet. The bench harness changes from #399 are also not part of this stack.

### Authorship and follow-up

Who wrote this:

- [ ] A human wrote it.

Who answers review comments:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2606,6 +2606,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "tokio-rustls",
+ "tokio-util",
  "tracing",
  "tracing-subscriber",
  "webpki-roots 0.26.11",
@@ -4392,6 +4393,7 @@ checksum = "9ae9cec805b01e8fc3fd2fe289f89149a9b66dd16786abd8b19cfa7b48cb0098"
 dependencies = [
  "bytes",
  "futures-core",
+ "futures-io",
  "futures-sink",
  "pin-project-lite",
  "tokio",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,7 +118,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -129,7 +129,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -1177,7 +1177,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -1506,9 +1506,9 @@ checksum = "0cc23270f6e1808e30a928bdc84dea0b9b4136a8bc82338574f23baf47bbd280"
 
 [[package]]
 name = "h2"
-version = "0.4.14"
+version = "0.4.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "171fefbc92fe4a4de27e0698d6a5b392d6a0e333506bc49133760b3bcf948733"
+checksum = "a9f37a958b41b3b19ee2707c06439c0e9e547e847223eb791ecb0cb821c65e27"
 dependencies = [
  "atomic-waker",
  "bytes",
@@ -1773,7 +1773,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.3",
+ "socket2 0.5.10",
  "tokio",
  "tower-service",
  "tracing",
@@ -1992,7 +1992,7 @@ checksum = "3640c1c38b8e4e43584d8df18be5fc6b0aa314ce6ebf51b53313d4306cca8e46"
 dependencies = [
  "hermit-abi",
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -2917,7 +2917,7 @@ version = "0.50.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -3298,7 +3298,7 @@ dependencies = [
  "quinn-udp",
  "rustc-hash 2.1.2",
  "rustls",
- "socket2 0.6.3",
+ "socket2 0.5.10",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
@@ -3335,7 +3335,7 @@ dependencies = [
  "cfg_aliases",
  "libc",
  "once_cell",
- "socket2 0.6.3",
+ "socket2 0.5.10",
  "tracing",
  "windows-sys 0.60.2",
 ]
@@ -3672,7 +3672,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.12.1",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -4052,7 +4052,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3a766e1110788c36f4fa1c2b71b387a7815aa65f88ce0229841826633d93723e"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -4182,7 +4182,7 @@ dependencies = [
  "getrandom 0.4.2",
  "once_cell",
  "rustix 1.1.4",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -4998,7 +4998,7 @@ version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2a7b1c03c876122aa43f3020e6c3c3ee5c05081c9a00739faf7503aeba10d22"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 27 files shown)


aimkiray:

https://github.com/madeye/meow-rs/compare/aimkiray:pr1-vless-deferred...aimkiray:pr2-mux-smux

This layer: 1 commit / 24 files. See #412 for the stack table and merge order.

madeye:
## Review — PR 2/6 of the mux stack

Reviewed as the incremental diff against `pr1-vless-deferred`.

Several things here are done well and are worth saying out loud: the `offer()` CAS-reserve plus the `Reservation` / `mem::forget` handoff to `MuxStreamConn::drop` is a genuinely correct fix for the classic pool overshoot race; routing `OutMsg::Flush` through the *same* bounded channel as data frames is the right way to get the FIFO guarantee, and `write_since_flush` correctly handles the "data queued after the flush request went out" case; `MapEntryGuard` covers cancelled opens; and `address.rs` rejects overlong FQDNs instead of truncating the frame.

Seven findings, two of which I'd like to see fixed in this PR rather than deferred — mainly because PR 6 of this stack already implements the shape I'm asking for, so the divergence is internal to the series.

### 1. `smux.rs:36` — a per-stream queue of 2 head-of-line blocks the entire session (high)

```rust
const STREAM_QUEUE: usize = MAX_STREAM_BUFFER / MAX_FRAME_SIZE;   // 64 KiB / 32 KiB = 2
```

and `smux.rs:376`, in the reader task:

```rust
result = tx.send(chunk) => { let _ = result; }
```

`handle_frame` awaits a bounded 2-slot channel. One consumer that stops reading blocks the reader task, which halts frame dispatch for *every* stream on that physical connection. Frames are not necessarily 32 KiB — with small frames it takes two of them to stall the session. For a proxy this is user-visible: one paused download stalls every other tab sharing the mux session.

The 4 MiB `receive_budget` semaphore already bounds memory (permits ride on `InboundChunk` and are released when the stream consumes or drops the chunk), which is also what xtaci relies on. The 2-slot channel therefore buys no safety, only stalls. Compare `muxcool.rs` in PR 6: `STREAM_QUEUE = 32` plus a non-blocking `deliver_now` / `Parked` pair that pauses the connection read deliberately and, per its own doc, guarantees "a consumer stalled mid-write cannot deadlock writers sharing the session". Either adopt that shape here or raise the depth substantially and let the semaphore do the bounding.

### 2. `smux.rs:454` — stream-map entries leak (medium)

`Drop for SmuxStream` calls `best_effort_fin()` but never removes `state.streams[id]`. The entry is removed only on a server `CMD_FIN` or in `mark_dead`. A server that half-closes late, or never FINs a stream the client abandoned, leaves the entry for the session's lifetime; under stream churn on a long-lived session these accumulate.

Sends into the stale entry are harmless (the receiver is gone, `send` fails immediately), so this is a leak rather than a correctness bug — but PR 6's `EndGuard` removes its map entry explicitly, with the comment *"Mux.Cool servers (notably sing-box) never echo End frames, so the driver cannot clean this up itself, and the cleanup must not wait on outbound channel capacity."* The same reasoning applies here; `MapEntryGuard`'s detached-spawn pattern is already available to reuse.

### 3. `mux/packet.rs:162` — `PacketWriter::write_datagram` is not cancellation-safe (medium)

```rust
self.stream.write_all(&(data.len() as u16).to_be_bytes()).await?;
self.stream.write_all(data).await?;
self.stream.flush().await
```

A `write_packet` future cancelled between (or inside) these leaves a length prefix with no payload, or a partial payload, and the peer's framing desyncs permanently. `SmuxStream::poll_write` returns `Pending` whenever the outbound channel is full, so this is reachable whenever the session is under write pressure and the caller drives `write_packet` inside a `select!`.

`PacketReader` right above it is carefully built around persistent progress for exactly this reason; the writer needs the mirror image (a retained `pending: Option<(Bytes, usize)>` that the next call resumes). Note the same shape exists in the plain VLESS UDP writer, so if you'd rather fix both together that's fine — but the asymmetry inside this one file is worth resolving.

### 4. `client.rs:446` — `min_streams` is ignored when `max_connections == 0` (medium)

```rust
} else {
    // max-connections=0 with max-streams=0 keeps the mihomo semantics
    options.max_streams > 0 && load < options.max_streams
};
```

sing-mux's `offer()` checks `minStreams` *before* falling through to `maxStreams` in the `maxConnections == 0` branch. A mihomo config of `max-connections: 0, min-streams: 8` therefore reuses up to 8 streams per physical connection upstream, but gets meow's `max_streams` (default 4) here. The `max_connections == 0 && max_streams == 0` warning in the parser catches the pathological case, but not this one.

### 5. `vless_adapter.rs:144` — `flow: xtls-rprx-vision` + mux is accepted (medium)

```rust
#[cfg(feature = "vless-vision")]
if flow.is_some() {
    let vless = VlessConn::new_deferred(...)
    return Ok(Box::new(VisionConn::new(vless, uuid_bytes)) as Box<dyn ProxyConn>);
}
```

sing-box and Xray both reject XTLS combined with multiplex. Building a Vision-wrapped mux session gets the user a connection that the server tears down (or worse, mis-frames) with no diagnostic. This should be a hard config error in `parse_vless` alongside the existing Class A gates, rather than something the dial path silently constructs.

Separately, `flow.is_some()` rather than matching the Vision variant is fragile if a second flow is ever added.

### 6. `tun/udp.rs:234` — the replacement comment overstates the guarantee (docs)

```
// A downstream read cancelled by another branch loses nothing:
// packet reads persist their progress across cancellation (see
// VlessUdpReader / MuxPacketConn), so the next poll resumes the
// same datagram.
```

`relay_flow` is generic over `Box<dyn ProxyPacketConn>`. The property holds for the two implementations named, but Direct / Shadowsocks / Trojan UDP conns still can drop a datagram on cancellation. The original "acceptable under UDP delivery semantics" comment was accurate; this one asserts an invariant the type doesn't carry.

### 7. No smux keepalive (minor)

xtaci and sing-mux send `CMD_NOP` keepalives; without them a half-dead session is only detected on an IO failure or after the 60 s idle eviction, so a stream opened onto a silently-dead session pays the full `open_stream_flags` retry.

### Note on defaults

Worth stating in `docs/specs/proxy-mux.md`: mihomo's `smux` block defaults to `0/0/0` (sing-mux then applies `minStreams = 8`), while this PR defaults to `4/4/4`. That's a defensible choice, but a user copying a mihomo config gets different pooling behaviour, and right now that only lives in a code comment.

aimkiray:
## Review fixes applied

Thank you for the review. All seven findings plus the defaults note are addressed.

### 1. `smux.rs:36` — STREAM_QUEUE=2 head-of-line blocks session (high) ✅

Raised `STREAM_QUEUE` from 2 to **32**. The 4 MiB `receive_budget` semaphore already bounds memory — the 2-slot channel bought no safety, only stalls. A depth of 32 gives real headroom before a stalled consumer pauses the reader.

The target architecture is muxcool's `deliver_now` / `Parked` pattern (PR 6): a full queue parks the delivery and pauses the connection read (session-wide flow control) without blocking writers. Porting that to smux is a follow-up PR; this depth raise is the safe interim fix, as you offered ("either adopt that shape here or raise the depth substantially").

### 2. `smux.rs:454` — stream-map entries leak (medium) ✅

`SmuxStream::drop` now removes the map entry explicitly via a detached spawn (`state.streams.lock().await.remove(&id)`), mirroring `MapEntryGuard`'s pattern. The `best_effort_fin()` is still sent first; cleanup does not wait on outbound channel capacity. This matches muxcool's `EndGuard` reasoning: "servers never echo End/FIN frames, so the driver cannot clean this up itself."

### 3. `mux/packet.rs:162` — PacketWriter not cancellation-safe (medium) ✅

Added `pending: Option<(Bytes, usize)>` to `PacketWriter`. The combined frame (length prefix + data) is built once as a contiguous `Bytes` and written with individual `write` calls that update the offset before each await point. A cancelled `write_datagram` future re-enters and resumes from the stored offset.

**Key detail**: `write_all` is NOT used — its internal progress is lost on cancellation, which would duplicate already-sent bytes. Individual `write` calls with manual offset tracking ensure each successful write persists its progress.

### 4. `client.rs:446` — `min_streams` ignored when `max_connections == 0` (medium) ✅

The `max_connections == 0` branch now checks `min_streams` before `max_streams`, matching sing-mux's offer precedence:

```rust
load < options.min_streams || (options.max_streams > 0 && load < options.max_streams)
```

The `zero_bounds_dial_one_connection_per_stream` test is updated to explicitly set `min_streams: 0` (it previously inherited the default 4 via `..MuxOptions::default()`).

### 5. `vless_adapter.rs:144` — `flow: xtls-rprx-vision` + mux accepted (medium) ✅

Hard config error in `parse_vless`: if `flow == Some(VlessFlow::XtlsRprxVision)` and mux is enabled, returns an error message citing that sing-box and Xray reject XTLS + mux.

In `vless_adapter.rs`, changed `flow.is_some()` to `flow_str.is_some()` (which is `Some` only for the Vision variant) with a defense-in-depth comment. This is robust against a future second flow variant.

### 6. `tun/udp.rs:234` — comment overstates guarantee (docs) ✅

Reverted to an accurate comment acknowledging that `relay_flow` is generic over `Box<dyn ProxyPacketConn>` and Direct / Shadowsocks / Trojan UDP conns do not carry the persistent-progress invariant. "This is acceptable under UDP delivery semantics."

### 7. No smux keepalive (minor)

Deferred to a follow-up PR — `CMD_NOP` keepalive is a feature add, not a correctness fix.

### Defaults documentation (note) ✅

Added to `docs/specs/proxy-mux.md`: meow defaults to `4/4/4` vs mihomo's `0/0/0` (sing-mux applies `minStreams = 8`).

### Smoke tests

All 13 smoke tests pass.

PR #415, opened 2026-08-18
feat(mux): extend sing-mux to Shadowsocks and VMess

aimkiray:
### Description

This is the third layer of the #394 stack. It extends the existing sing-mux client to Shadowsocks and VMess.

Shadowsocks gains mux pooling and dial integration while preserving cipher and plugin state. VMess gains sing-mux request and stream integration. The configuration parser accepts the legacy `mux:` form and canonical `smux:` form for both proxy types.

Stack position:

- Base: `pr2-mux-smux` (`7b442ff`)
- Next: `pr4-mux-yamux`

Given a Shadowsocks proxy with mux enabled, when multiple streams are opened, the cipher and plugin state remain valid for the lifetime of the pooled session.

Given a VMess proxy with sing-mux enabled, when a logical stream is created, the VMess request carries the sing-mux setup and the stream can be used independently of other streams.

Given the canonical `smux:` configuration key, when an SS or VMess proxy is loaded, it is accepted without an unrelated mux warning. With mux disabled, the existing non-mux path remains available.

### Key points

This PR only extends the already-reviewed sing-mux infrastructure to two outbound types. It does not add a new mux wire protocol. Keeping protocol implementation and outbound coverage separate lets later yamux and h2mux changes reuse the same adapter wiring.

### Notes for reviewers

Review the Shadowsocks dial-state ownership and the VMess request/header changes first. This PR should be reviewed incrementally against `pr2-mux-smux`.

Yamux, h2mux, and Mux.Cool remain in later stack entries.

### Authorship and follow-up

Who wrote this:

- [ ] A human wrote it.

Who answers review comments:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2606,6 +2606,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "tokio-rustls",
+ "tokio-util",
  "tracing",
  "tracing-subscriber",
  "webpki-roots 0.26.11",
@@ -4392,6 +4393,7 @@ checksum = "9ae9cec805b01e8fc3fd2fe289f89149a9b66dd16786abd8b19cfa7b48cb0098"
 dependencies = [
  "bytes",
  "futures-core",
+ "futures-io",
  "futures-sink",
  "pin-project-lite",
  "tokio",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,7 +118,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -129,7 +129,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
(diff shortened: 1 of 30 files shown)


aimkiray:

https://github.com/madeye/meow-rs/compare/aimkiray:pr2-mux-smux...aimkiray:pr3-mux-ss-vmess

This layer: 1 commit / 5 files. See #412 for the stack table and merge order.

madeye:
## Review — PR 3/6 of the mux stack

Reviewed as the incremental diff against `pr2-mux-smux`.

No findings. This is the layer I'd merge with the least hesitation.

The `SsCore` extraction is the right call and gets the details right: pulling `server` / `port` / `server_config` / `context` / `plugin` behind one `Arc` keeps the SIP003 plugin handle alive for the mux session as well as the adapter, and moving the dial into `SsCore::dial_tcp_stream` preserves the `PluginKind::None | PluginKind::External` dispatch on `tcp_external_addr` — so the `SocketProtector` / `HostResolver` behaviour that the long comment there exists to protect is unchanged for both paths. Pointing the session's `ProxyClientStream` at `sp.mux.sing-box.arpa:444` is correct for AEAD-2022 too, since the address is bound once per physical stream.

On the VMess side, hoisting the header exchange into the free `dial_vmess` function keeps the plain and mux dial paths byte-identical, and threading the command through `HeaderCmd` rather than a bool leaves room for `CommandMux` in PR 6 without touching `seal_request_header`'s callers again.

One nit: `HeaderCmd` is introduced with an `Option<&Metadata>` shape documented as *"`None` for CommandMux (no port/address follows the command byte)"*, but both variants are `Some(m)` until PR 6 adds the Mux variant, so `build_header_plaintext`'s `if let Some(metadata)` is a dead branch for the duration of this PR. Harmless as stack scaffolding — just flagging it in case clippy's `--all-features` run in CI has an opinion.

aimkiray:
## Review fixes applied

Thank you — no findings on this layer. The `HeaderCmd` dead-branch nit was checked: `cargo clippy --all-features --all-targets -- -D warnings` passes clean (clippy has no opinion on the `if let Some(metadata)` branch that is dead until PR 6 adds the Mux variant). No code changes needed.

All smoke tests pass.

PR #416, opened 2026-08-18
feat(mux): add yamux sing-mux protocol

aimkiray:
### Description

This is the fourth layer of the #394 stack. It adds yamux as the second sing-mux protocol.

The yamux session and stream adapter use the workspace `yamux` dependency and dispatch on sing-mux protocol byte `1`. The implementation reuses the existing pool, request prefix, response preamble, stream lifecycle, and UDP packet path. Yamux-specific pool, UDP, and configuration coverage is included.

Stack position:

- Base: `pr3-mux-ss-vmess` (`fec2538`)
- Next: `pr5-mux-h2mux`

Given `protocol: yamux`, when several streams are opened through the supported sing-mux outbound paths, they use the shared pool and remain independently readable and writable.

Given a yamux UDP stream, when several datagrams are sent, the datagram boundaries match the existing sing-mux packet behavior.

Given the underlying yamux session reaches EOF, when the pool checks the session, it is treated as dead and is not reused.

### Key points

Yamux is added as a focused protocol adapter rather than a second pool implementation. This keeps lifecycle, backpressure, request encoding, and UDP behavior consistent with smux.

### Notes for reviewers

Start with `mux/yamux.rs`, then review the small protocol dispatch changes in the shared client. This PR should be reviewed incrementally against `pr3-mux-ss-vmess`.

H2Mux and Mux.Cool are intentionally deferred to the next stack entries. The bench harness changes from #399 are not included.

### Authorship and follow-up

Who wrote this:

- [ ] A human wrote it.

Who answers review comments:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2611,6 +2611,7 @@ dependencies = [
  "tracing-subscriber",
  "webpki-roots 0.26.11",
  "x25519-dalek",
+ "yamux",
 ]
 
 [[package]]
@@ -2865,6 +2866,12 @@ dependencies = [
  "libc",
  "memoffset",
 ]
+
+[[package]]
+name = "nohash-hasher"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2bf50223579dc7cdcfb3bfcacf7069ff68243f8c363f62ffa99cf000a6b9c451"
 
 [[package]]
 name = "nom"
@@ -4099,6 +4106,12 @@ name = "stable_deref_trait"
 version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6ce2be8dc25455e1f91df71bfa12ad37d7af1092ae736f3a6cd0e37bc7810596"
+
+[[package]]
+name = "static_assertions"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 
 [[package]]
 name = "strsim"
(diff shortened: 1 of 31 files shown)


aimkiray:

https://github.com/madeye/meow-rs/compare/aimkiray:pr3-mux-ss-vmess...aimkiray:pr4-mux-yamux

This layer: 1 commit / 8 files. See #412 for the stack table and merge order.

madeye:
## Review — PR 4/6 of the mux stack

Reviewed as the incremental diff against `pr3-mux-ss-vmess`.

No blocking findings. Adding yamux as a protocol adapter over the existing pool rather than a second pool implementation is the right shape, and the dispatch additions to `SessionKind` / `MuxStreamKind` are mechanical.

The hand-rolled driver in `yamux.rs` is the only interesting part, and it holds up: draining `open_rx` into a FIFO `pending_opens` before serving `poll_new_outbound` means no queued opener can be dropped by a newer one; every `Poll::Pending` path is reached only after all three sources (request channel, outbound ack, socket) have registered their wakers, so the final `return Poll::Pending` cannot lose a wake-up; and `Poll::Ready(Some(Err(_))) | Poll::Ready(None)` drops the pending oneshots, so queued openers get `BrokenPipe` rather than hanging. Session lifetime is safe because `MuxStreamConn` / `MuxPacketConn` hold the `Arc<MuxSession>`, so an evicted-from-pool session stays alive while its streams do.

Two notes, neither blocking:

- `yamux::Config::default()` drops sing-mux's 5 s open/close timeouts. The module doc calls this out as an acceptable divergence, which is fine — but it does mean a server that accepts the physical connection and then stalls will park `open_stream` indefinitely, since `SESSION_SETUP_TIMEOUT` only wraps `create_session`.
- Server-initiated streams are `drop(stream)`ped silently. A `debug!` there would make a misconfigured or hostile peer visible instead of invisible.

Good to see the yamux UDP round-trip test added in `packet.rs` alongside the smux one.

aimkiray:
## Review fixes applied

Thank you — no blocking findings. Both notes addressed:

- **Server-initiated stream drop**: Added `tracing::debug!("yamux: server-initiated stream dropped (reset)")` before `drop(stream)`, making a misconfigured or hostile peer visible in logs.
- **`yamux::Config::default()` timeout divergence**: Documented and acceptable — no change needed.

All smoke tests pass.

PR #417, opened 2026-08-18
feat(mux): add h2mux and shared HTTP/2 stream

aimkiray:
### Description

This is the fifth layer of the #394 stack. It adds h2mux, the third sing-mux protocol, and extracts the shared raw HTTP/2 stream state needed by the implementation.

The change adds h2mux session and stream handling, dispatches sing-mux protocol byte `2`, and wires the h2 feature and optional dependencies. The empty/default mux protocol becomes h2mux, matching mihomo. The `minimal` build remains free of mux dependencies.

Stack position:

- Base: `pr4-mux-yamux` (`3023ca3`)
- Next: `pr6-mux-muxcool`

Given mux is enabled without an explicit protocol, when configuration is parsed, h2mux is selected as the default.

Given `protocol: smux` or `protocol: yamux`, when configuration is parsed, the explicit protocol remains respected.

Given an HTTP/2 peer that delays response headers until request DATA, when the first application write is made, the connection does not deadlock waiting for response headers.

Given an H2 write is Pending and the caller changes the buffer, when the write is polled again, stale bytes are not sent and the changed-buffer contract violation is rejected.

### Key points

The shared H2 state is extracted in `meow-transport` so h2mux and the existing H2 transport use the same response and write-cancellation invariants. Changing the empty/default protocol is intentionally delayed until h2mux exists in the stack.

### Notes for reviewers

Review the shared H2 state transitions before the h2mux session adapter. This PR should be reviewed incrementally against `pr4-mux-yamux`.

After this PR, smux, yamux, and h2mux complete the sing-mux family. Xray Mux.Cool is a separate VLESS/VMess-specific implementation in PR6.

### Authorship and follow-up

Who wrote this:

- [ ] A human wrote it.

Who answers review comments:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2611,6 +2611,7 @@ dependencies = [
  "tracing-subscriber",
  "webpki-roots 0.26.11",
  "x25519-dalek",
+ "yamux",
 ]
 
 [[package]]
@@ -2865,6 +2866,12 @@ dependencies = [
  "libc",
  "memoffset",
 ]
+
+[[package]]
+name = "nohash-hasher"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2bf50223579dc7cdcfb3bfcacf7069ff68243f8c363f62ffa99cf000a6b9c451"
 
 [[package]]
 name = "nom"
@@ -4099,6 +4106,12 @@ name = "stable_deref_trait"
 version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6ce2be8dc25455e1f91df71bfa12ad37d7af1092ae736f3a6cd0e37bc7810596"
+
+[[package]]
+name = "static_assertions"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 
 [[package]]
 name = "strsim"
@@ -5513,6 +5526,21 @@ dependencies = [
  "rusticata-macros",
  "thiserror 2.0.18",
  "time",
+]
+
+[[package]]
+name = "yamux"
+version = "0.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "767113d8f66a81e461362462aa71d8d0108cbf3430d4442fb88a04e31be81165"
+dependencies = [
+ "futures",
+ "log",
+ "nohash-hasher",
+ "parking_lot",
+ "pin-project",
+ "static_assertions",
+ "web-time",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2606,6 +2606,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "tokio-rustls",
+ "tokio-util",
  "tracing",
  "tracing-subscriber",
  "webpki-roots 0.26.11",
@@ -4392,6 +4393,7 @@ checksum = "9ae9cec805b01e8fc3fd2fe289f89149a9b66dd16786abd8b19cfa7b48cb0098"
 dependencies = [
  "bytes",
  "futures-core",
+ "futures-io",
  "futures-sink",
  "pin-project-lite",
  "tokio",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2578,6 +2578,7 @@ dependencies = [
  "crc32fast",
  "ctr",
  "futures",
+ "h2",
  "h3",
  "h3-quinn",
  "hex",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,7 +118,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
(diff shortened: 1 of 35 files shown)


aimkiray:

https://github.com/madeye/meow-rs/compare/aimkiray:pr4-mux-yamux...aimkiray:pr5-mux-h2mux

This layer: 1 commit / 12 files. See #412 for the stack table and merge order.

madeye:
## Review — PR 5/6 of the mux stack

Reviewed as the incremental diff against `pr4-mux-yamux`.

Extracting the raw H2 stream state into `h2_common` so h2mux and the h2 transport share one set of response/write invariants is the right move, and the lazy-response design (`RecvState` resolved on first read, parameterised by `StatusPolicy` and an optional timeout) is the correct fix for the sing-box deferred-headers deadlock described in the module docs. `Drop for Session { self._task.abort() }` is a good catch — without it an evicted idle session keeps its fd until the peer closes.

One finding is significant.

### 1. `h2_common.rs:227` — the changed-buffer guard is not in the code (high)

The PR description states:

> Given an H2 write is Pending and the caller changes the buffer, when the write is polled again, stale bytes are not sent and the changed-buffer contract violation is rejected.

The implementation does not do this:

```rust
// This relies on the AsyncWrite contract that a Pending poll is
// retried with the same buffer.
if this.pending_write.is_none() {
    let data = Bytes::copy_from_slice(buf);
    this.send.reserve_capacity(data.len());
    this.pending_write = Some((data, 0));
}
```

There is no comparison against `buf` anywhere in the file. When `pending_write` is set, the stash branch is skipped entirely and the stale `data[offset..]` is sent, with its length returned as the number of bytes consumed from the *new* `buf`. So stale bytes are sent, and the violation is not rejected.

#407 (`fix/h2-write-cancellation`) has exactly this guard on the older `Option<Bytes>` shape:

```rust
if let Some(data) = &this.pending_write {
    if buf != data.as_ref() {
        this.pending_write = None;
        return Poll::Ready(Err(io::Error::new(
            io::ErrorKind::InvalidInput,
            "h2: write buffer changed after Pending; \
             AsyncWrite requires retrying with the same buffer",
        )));
    }
} else { ... }
```

It was lost in the rewrite to `(Bytes, usize)`. The offset-aware form is `buf != &data[*offset..]` — worth keeping the partial-send offset, since that part is an improvement over #407, but the comparison has to come back with it. Worth adding a test that drops a `Pending` write and re-polls with different bytes; the current suite doesn't cover it.

Note this is the same defect class as my first comment on #413 (`reality_tls.rs:441` now returns `Pending` with a sealed record buffered and reports its length against whatever buffer arrives next). A consistent decision across the two would be good — either both reject the changed buffer, or both keep buffered `Ready` semantics and never return `Pending` after consuming.

### 2. `h2_common.rs:252` — `allowed == 0` returns `Pending` with no waker (minor)

```rust
let allowed = capacity.min(remaining);
if allowed == 0 {
    return Poll::Pending;
}
```

`poll_capacity` has already returned `Ready`, so nothing registered a waker on this path — if it were reachable the task would park forever. It isn't today (h2 only yields `Some(Ok(n))` with `n > 0`, and `remaining == 0` implies `pending_write` was already cleared), but the arm reads as if it expects to be reachable. Either make it unreachable explicitly or re-register.

### 3. Per-stream open has no timeout (minor)

`SESSION_SETUP_TIMEOUT` wraps `create_session` only. `send_request.ready()` in `Session::open_stream` can park indefinitely once the peer's `SETTINGS_MAX_CONCURRENT_STREAMS` is exhausted — reachable if the pool's own `max_streams` accounting is looser than the server's setting. `RESPONSE_TIMEOUT` covers the response, not the open.

Changing the empty/default protocol to h2mux here, after all three sing-mux protocols exist, is the right sequencing.

aimkiray:
## Review fixes applied

Thank you for the review. All three findings are fixed.

### 1. `h2_common.rs:227` — changed-buffer guard missing (high) ✅

Restored the guard from #407 with offset-aware comparison:

```rust
if let Some((data, offset)) = &this.pending_write {
    if buf != &data[*offset..] {
        this.pending_write = None;
        return Poll::Ready(Err(io::Error::new(
            io::ErrorKind::InvalidInput,
            "h2: write buffer changed after Pending; \
             AsyncWrite requires retrying with the same buffer",
        )));
    }
} else {
    let data = Bytes::copy_from_slice(buf);
    this.send.reserve_capacity(data.len());
    this.pending_write = Some((data, 0));
}
```

The offset-aware comparison `buf != &data[*offset..]` handles both cases correctly:
- **Pending retry** (offset=0): caller retries with same full buffer → `buf == &data[0..]` → passes.
- **Partial send + write_all advance** (offset>0): `write_all` advances by `chunk_len`, so `buf == &data[chunk_len..]` → passes.
- **Changed buffer**: mismatch → rejected with `InvalidInput`.

This is consistent with #413's reality_tls fix in invariant ("never report bytes consumed from a buffer not actually read") while differing in mechanism: h2 must sometimes return `Pending` for flow control, so the guard is necessary; reality_tls can always buffer-and-return `Ready`, so it never returns `Pending` after consuming.

New test: `poll_write_rejects_changed_buffer_after_pending` — fills the per-stream flow control window, interrupts a stalled `write` with a timeout (leaving `pending_write` set), then re-polls with a different buffer → asserts `InvalidInput`.

### 2. `h2_common.rs:252` — `allowed == 0` returns Pending with no waker (minor) ✅

Added `debug_assert!(allowed > 0, "h2: zero capacity grant with remaining data")` and `cx.waker().wake_by_ref()` as a defensive re-register, so a future code change reaching this arm cannot park the task forever.

### 3. Per-stream open has no timeout (minor) ✅

Wrapped `send_request.ready()` in `tokio::time::timeout(OPEN_TIMEOUT, ...)` where `OPEN_TIMEOUT = 5s` (mirrors `RESPONSE_TIMEOUT`). A peer with exhausted `SETTINGS_MAX_CONCURRENT_STREAMS` can no longer park `open_stream` indefinitely.

### Smoke tests

All 13 smoke tests pass (including h2mux TCP + UDP live probes).

PR #418, opened 2026-08-18
feat(mux): add Xray Mux.Cool for VLESS and VMess

aimkiray:
### Description

This is the sixth and final layer of the #394 stack. It adds Xray Mux.Cool for VLESS and VMess.

The change implements the Mux.Cool frame codec, session driver, stream lifecycle, flow control, UDP Keep frames, and teardown handling. VLESS gains CommandMux request encoding and deferred mux setup. VMess gains mux request-header signaling and Mux.Cool session setup. Configuration dispatch keeps Mux.Cool restricted to VLESS and VMess.

Stack position:

- Base: `pr5-mux-h2mux` (`80b601c`)
- Final stack entry.

Given a VLESS or VMess proxy configured with `muxcool`, when multiple logical streams are opened, they reuse one physical session.

Given a peer that silently closes a Mux.Cool stream without an End echo, when the client observes the teardown, the reader receives EOF and the stream-map entry is removed.

Given a malformed Mux.Cool frame, when the frame is decoded, the decoder returns an error instead of panicking.

Given `muxcool` is selected for Trojan or Shadowsocks, when configuration is loaded, the proxy is rejected because Mux.Cool is limited to VLESS and VMess CommandMux endpoints.

Given a Mux.Cool UDP Keep frame, when a datagram is sent, its destination metadata is preserved.

**Local end-to-end smoke (run 2026-08-18, sing-box 1.13.13 on macOS arm64).** A sing-box instance with VLESS / Trojan(+TLS, self-signed) / Shadowsocks / VMess inbounds, each with `multiplex.enabled: true`, was started on 127.0.0.1 together with local HTTP and UDP echo services. meow was launched with a one-proxy config (`rules: [MATCH,<proxy>]`) exposing the mixed listener on 127.0.0.1. Each scenario below passed: one request through the proxy returned `SMOKE-OK`, then 8 concurrent requests through the same mixed listener all returned `SMOKE-OK`, so the streams multiplex over shared physical sessions (sing-box logged each physical connection as an `inbound multiplex connection`):

- VLESS + smux, VLESS + yamux, VLESS + h2mux, VLESS + muxcool
- Trojan(TLS) + smux, Trojan(TLS) + yamux, Trojan(TLS) + h2mux
- Shadowsocks + h2mux, VMess + h2mux, VMess + muxcool

UDP relay was additionally exercised through the repository's live interop probes (`cargo test -p meow-proxy --lib live_singbox_vless -- --ignored`): `live_singbox_vless_probe` (h2mux TCP), `live_singbox_vless_udp_probe` (h2mux UDP echo), and `live_singbox_vless_muxcool_probe` (Mux.Cool TCP + UDP echo) — all passed against the same sing-box instance.

Not covered locally: SS/Trojan/VMess UDP paths, an Xray-core node (Mux.Cool was verified against sing-box's CommandMux handling, not Xray), and a real external node.

The branch was checked with repository formatting, all three clippy feature modes, doc warnings, a locked workspace check, affected crate tests, and configuration integration tests.

### Key points

Mux.Cool is kept separate from sing-mux because it has a different signaling path and frame model. Sing-mux uses protocol bytes and a shared request prefix; Mux.Cool uses VLESS/VMess CommandMux signaling and its own frame codec.

The final stack uses a cleaned lockfile rather than retaining unrelated registry version churn from the original PR. The four bench harness commits belong to #399 and are intentionally excluded.

### Notes for reviewers

Start with the frame codec and session driver, then review the VLESS/VMess signaling and teardown paths. This PR should be reviewed incrementally against `pr5-mux-h2mux`.

After this PR, the functional tree matches the rebased full mux implementation used as the split source. The intentional differences are the cleaned lockfile and the bench harness changes retained separately in #399.

### Authorship and follow-up

Who wrote this:

- [ ] A human wrote it.

Who answers review comments:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2611,6 +2611,7 @@ dependencies = [
  "tracing-subscriber",
  "webpki-roots 0.26.11",
  "x25519-dalek",
+ "yamux",
 ]
 
 [[package]]
@@ -2865,6 +2866,12 @@ dependencies = [
  "libc",
  "memoffset",
 ]
+
+[[package]]
+name = "nohash-hasher"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2bf50223579dc7cdcfb3bfcacf7069ff68243f8c363f62ffa99cf000a6b9c451"
 
 [[package]]
 name = "nom"
@@ -4099,6 +4106,12 @@ name = "stable_deref_trait"
 version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6ce2be8dc25455e1f91df71bfa12ad37d7af1092ae736f3a6cd0e37bc7810596"
+
+[[package]]
+name = "static_assertions"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 
 [[package]]
 name = "strsim"
@@ -5513,6 +5526,21 @@ dependencies = [
  "rusticata-macros",
  "thiserror 2.0.18",
  "time",
+]
+
+[[package]]
+name = "yamux"
+version = "0.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "767113d8f66a81e461362462aa71d8d0108cbf3430d4442fb88a04e31be81165"
+dependencies = [
+ "futures",
+ "log",
+ "nohash-hasher",
+ "parking_lot",
+ "pin-project",
+ "static_assertions",
+ "web-time",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2606,6 +2606,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "tokio-rustls",
+ "tokio-util",
  "tracing",
  "tracing-subscriber",
  "webpki-roots 0.26.11",
@@ -4392,6 +4393,7 @@ checksum = "9ae9cec805b01e8fc3fd2fe289f89149a9b66dd16786abd8b19cfa7b48cb0098"
 dependencies = [
  "bytes",
  "futures-core",
+ "futures-io",
  "futures-sink",
  "pin-project-lite",
  "tokio",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2578,6 +2578,7 @@ dependencies = [
  "crc32fast",
  "ctr",
  "futures",
+ "h2",
  "h3",
  "h3-quinn",
  "hex",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,7 +118,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -129,7 +129,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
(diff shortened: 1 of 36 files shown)


aimkiray:

https://github.com/madeye/meow-rs/compare/aimkiray:pr5-mux-h2mux...aimkiray:pr6-mux-muxcool

This layer: 1 commit / 12 files. See #412 for the stack table and merge order.

This PR also documents the full end-to-end smoke results (VLESS/Trojan/Shadowsocks/VMess × smux/yamux/h2mux/muxcool TCP + concurrency, plus h2mux and muxcool UDP via the live probes).

madeye:
## Review — PR 6/6 of the mux stack

Reviewed as the incremental diff against `pr5-mux-h2mux`.

This is the strongest layer in the stack; no blocking findings.

The frame codec is bounds-checked on every branch — `decode_meta` rejects a short meta, an `End` frame carrying the Data option (which would otherwise put a payload length where the next `meta_len` is expected and desync the stream), an unknown network type and a `New` frame with no destination, and `parse_address` length-checks each address family before indexing. `MAX_META` caps the meta allocation, and `MAX_PAYLOAD` chunks the TCP write path while `write_packet` explicitly rejects datagrams beyond u16 framing.

The driver design is better than the smux one in PR 2, and worth keeping as the reference shape: `deliver_now` / `retry_parked` never block on a stream queue, so a stalled consumer pauses the connection read (session-wide flow control) without ever blocking writers sharing the session, and `STREAM_QUEUE = 32` gives real headroom before that happens. The read state machine only advances `phase` on completion, so a cancelled `read_step` resumes mid-frame instead of desyncing. `StreamMapGuard` covers cancelled opens, and `EndGuard` handles the "Mux.Cool servers never echo End" teardown by removing the map entry *before* queueing the End frame, so cleanup never waits on outbound capacity — that comment is also what made me flag the missing equivalent in `SmuxStream::drop` over on #414.

Two cosmetic notes:

- `muxcool.rs:149` and `:177` — `out.put_u16(data.len() as u16)` is unchecked in both `encode_data_frame` and `encode_udp_data_frame`. Both callers bound it today (`MAX_PAYLOAD` in `Stream::poll_write`, the explicit `u16::MAX` check in `write_packet`), but `smux.rs`'s `encode_frame` locks the same invariant with a `debug_assert!`. A `u16::try_from` or an assert here would make it structural rather than dependent on the callers staying correct.
- `muxcool.rs:494` — `alloc_sid` keeps `fetch_add`ing past exhaustion, so after 2^32 opens the `u32` wraps back into the valid range and could recycle a live id. Unreachable in any real session, and the `retired` flag does the right thing at the boundary, but a saturating add is free.

Restricting Mux.Cool to VLESS and VMess at config-parse time (rather than letting Trojan/SS build a session the server can't interpret) is the right call, and the end-to-end smoke matrix in the description — 10 protocol × mux combinations against sing-box 1.13.13 plus the three live UDP probes — is exactly the evidence this stack needed. The gaps you list (no Xray-core node for Mux.Cool, SS/Trojan/VMess UDP paths) are the right ones to state explicitly.

aimkiray:
## Review fixes applied

Thank you — no blocking findings. Both cosmetic notes addressed:

### 1. `muxcool.rs:149,177` — unchecked `put_u16(data.len() as u16)` ✅

Added `debug_assert!(data.len() <= u16::MAX as usize, "muxcool: data frame payload exceeds u16")` in both `encode_data_frame` and `encode_udp_data_frame`, matching `smux.rs`'s `encode_frame` structural invariant. Both callers already bound the length (`MAX_PAYLOAD` in `Stream::poll_write`, the explicit `u16::MAX` check in `write_packet`), so this makes the invariant explicit rather than dependent on callers.

### 2. `muxcool.rs:494` — `alloc_sid` `fetch_add` past exhaustion ✅

Replaced `fetch_add` with a `compare_exchange` loop:

```rust
loop {
    let val = self.next_sid.load(Ordering::SeqCst);
    if val > u16::MAX as u32 {
        self.retired.store(true, Ordering::SeqCst);
        return Err(io::Error::other("mux.cool session id space exhausted"));
    }
    match self.next_sid.compare_exchange(val, val + 1, SeqCst, SeqCst) {
        Ok(_) => return Ok(val as u16),
        Err(_) => continue,  // CAS contention — retry, don't declare retired
    }
}
```

This prevents the u32 from wrapping past `u16::MAX + 1` back into the valid range (which `fetch_add` would do after 2³² opens). The loop retries on CAS contention from concurrent `open_stream` calls, rather than falsely setting `retired` (which `fetch_update` would do on CAS failure).

### Smoke tests

All 13 smoke tests pass (including muxcool TCP + UDP live probes).

PR #450, opened 2026-08-20
fix(mux): recover from cancelled partial datagram writes instead of erroring

madeye:
Follow-up to #442 review and #449 review — applies the minor/non-blocking findings from the adversarial review of those PRs ("mux-datagram-recovery" group).

## Findings addressed

**1. Recoverable different-data writes (#442 review, `mux/packet.rs`)**
`PacketWriter::write_datagram` rejected different data after a cancelled partial write with `InvalidInput` — and every production caller (meow-tunnel `udp.rs`, `socks5_udp.rs`, `tun/udp.rs` relay loops) tears the UDP flow down on write error, turning a recoverable state into a flow teardown. Since the pending frame owns its `Bytes`, the recoverable design is implemented:
- **same data** → retry: resume the pending frame from its offset (unchanged);
- **different data, offset 0** → nothing committed: replace the stale frame (unchanged);
- **different data after a partial write** → complete the stale frame on the wire first (late delivery, not truncation — peer framing preserved), then frame and send the new datagram. Two wire datagrams, no error, no drop.

No soundness hole found: the drain updates the stored offset after every successful `write` (so cancellation mid-drain of either frame resumes correctly), and an I/O error keeps the frame + offset so state stays consistent on retry.

**2. Content-based identity comment (#442 review)**
Documented that after a cancelled partial write, a *new* datagram carrying identical bytes is indistinguishable from a retry and coalesces into one wire datagram — acceptable under UDP semantics.

**3. Error attribution (#442 review)**
The "datagram changed after a cancelled partial write" caller-misuse error no longer exists. The only error that can surface from this path now is the underlying stream I/O error from draining the stale frame on a dead stream — attributed to the dead stream, not the caller — and the comment states this explicitly.

**4. `MuxClient::offer()` doc wording (#449 review, `mux/client.rs`)**
The doc said "Dead sessions are pruned" although the predicate was renamed `is_dead` → `is_unusable` precisely because retired-but-alive Mux.Cool sessions are also pruned. Reworded to describe both cases.

## Tests (deliberate pinned-behavior change)

`cancelled_partial_write_rejects_different_data` (from #442, pinning the old reject contract) is **deliberately replaced** by `cancelled_partial_write_different_data_completes_stale_then_sends_new`, which asserts the wire bytes of *both* datagrams (`\x00\x04AAAA` then `\x00\x02BB`) plus a clean frame boundary afterwards. The other two contract points were already pinned with wire-byte assertions and remain green unchanged:
- `cancelled_partial_write_resumes_with_same_data` — same-data retry → exactly one wire datagram;
- `cancelled_unwritten_frame_replaced_by_new_data` — offset-0 stale frame simply replaced → one wire datagram.

## Verification

Full regression bar green from the worktree (`CARGO_TARGET_DIR` unset):
`cargo fmt --all -- --check` · `cargo clippy --all-targets` (default / `--no-default-features` / `--all-features`, all `-D warnings`) · `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` · `cargo test --lib` (13/13 suites, 0 failures).

No touched types from the ADR-0011 byte-count list; no relay-path allocation changes (the pending frame's owned `Bytes` predates this PR — this change adds no per-datagram allocation).

### DIFF
--- a/crates/meow-proxy/src/mux/client.rs
+++ b/crates/meow-proxy/src/mux/client.rs
@@ -466,8 +466,10 @@ impl MuxClient {
     }
 
     /// Pick an existing session that can take a new request, or dial a new
-    /// one.  Dead sessions are pruned and zero-stream sessions idle past
-    /// `IDLE_TIMEOUT` are evicted.  The returned session has one slot
+    /// one.  Unusable sessions are pruned (`is_unusable`: dead transports,
+    /// but also retired-yet-alive Mux.Cool sessions that must not take new
+    /// streams) and zero-stream sessions idle past `IDLE_TIMEOUT` are
+    /// evicted.  The returned session has one slot
     /// reserved for the caller via a CAS on `streams` — the check and the
     /// increment are a single atomic step, so concurrent offers can never
     /// overshoot max-streams.
--- a/crates/meow-proxy/src/mux/packet.rs
+++ b/crates/meow-proxy/src/mux/packet.rs
@@ -170,26 +170,36 @@ impl<W: tokio::io::AsyncWrite + Unpin> PacketWriter<W> {
     }
 
     async fn write_datagram(&mut self, data: &[u8]) -> std::io::Result<()> {
-        // A pending frame may only be resumed by a retry carrying the
-        // *same* datagram — otherwise the new datagram would be silently
-        // dropped while its bytes are reported as sent. If the new call
-        // carries different data: before any byte hit the wire the stale
-        // frame is simply discarded; after a partial write the length
-        // prefix is committed, so the frame must be completed with the
-        // original bytes — reject the changed buffer instead, mirroring
-        // `H2Stream::poll_write`'s changed-buffer guard (`h2_common.rs`).
-        // The pending frame is kept so a retry with the original data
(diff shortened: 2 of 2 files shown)


madeye:
Reviewed and verified: merges clean with `main`, 94 mux tests pass including `cancelled_partial_write_different_data_completes_stale_then_sends_new`. No findings.

Completing the stale frame rather than truncating is the right call — the length prefix is already committed, so late delivery preserves peer framing where truncation would desync it.

I traced the cancellation windows specifically, since that's where this class of bug lives. The edge I most expected to be wrong is correct: cancelled after the final `write` but before the loop clears `pending`, the frame survives with `offset == len`, and a later call with *different* data drains (immediate no-op), clears, then sends the new frame — while a later call with the *same* data correctly does **not** re-send, because those bytes did reach the wire. `drain_pending` only returns `Ok` with `pending == None` or `offset >= len`, so the downstream `pending.is_none()` builder check is guaranteed to fire and no path can send new data under the stale frame's length.

Minor: the content-based-coalescing caveat in the comment is narrower than it reads — `pending` is always cleared on a successful call, so coalescing is only reachable inside the post-cancellation window.

PR #451, opened 2026-08-20
fix(transport): flush stashed frames at shutdown + gun/reality polish (PR #439/#440 review follow-ups)

madeye:
Follow-up to the reviews of #440 (gRPC gun transport) and #439 (REALITY pre-auth flight bounds). These findings were classified minor/non-blocking at merge time; this PR fixes them properly. All four findings were re-verified against current `main` and still applied.

### 1. `GunStream::poll_shutdown` flushes a stashed `pending_write` frame (follow-up to #440 review)
`poll_shutdown` sent an empty DATA+EOS while an encoded frame from a write cancelled at its `Pending` await still sat in `pending_write` — silently truncating the stream. It now takes the stashed frame and sends it together with the EOS flag.

Guard semantics preserved:
- **Cancellation-safe**: `send_data` queues beyond the granted flow-control window inside h2, so the flush completes in a single poll — no `Pending` state to lose.
- **Only a cancelled write is flushed**: the changed-buffer guard clears `pending_write` *before* returning `InvalidInput`, so shutdown after a rejection never resurrects the cleared frame (pinned by test I).

### 2. `H2Stream` had the same gap — fixed consistently
`H2Stream::pending_write` is owned `Bytes` (copied from the caller's buffer at stash time), so it can flush safely too. `best_effort_eos` (shared by `poll_shutdown` and `Drop`) now sends the stashed payload with the closing EOS frame. Pinned by new unit tests `shutdown_flushes_stashed_pending_write` and `shutdown_after_rejected_buffer_does_not_resurrect_payload`, which verify byte-exact delivery through a capacity-hoarding h2 server.

### 3. `GunStream::poll_write` empty-buffer early return (follow-up to #440 review)
Aligned with `H2Stream`'s `if buf.is_empty() { return Ok(0) }` ordering (checked before the changed-buffer guard). Previously `write(&[])` while a frame was stashed returned `InvalidInput`, and an empty write on an idle stream emitted a useless zero-payload gun frame. Pinned by test J, which also asserts the stash survives the empty write and nothing extra reaches the wire.

### 4. `grpc_test.rs` no longer leaks the h2 `RecvStream` (follow-up to #440 review)
`spawn_capacity_hoarding_server` used `std::mem::forget(body)` to keep the stream alive. It now drains the body to end-of-stream inside the task (holding it across the awaits) and reports the collected bytes over a oneshot — which tests H/I/J reuse to assert exactly which frames were sent.

### 5. `reality_tls.rs` bound-provenance comment (follow-up to #439 review)
Added a note on `MAX_PRE_AUTH_HS_MESSAGES` / `MAX_PRE_AUTH_TRANSCRIPT_LEN` (64 KiB) stating they are defense-in-depth: the load-bearing bound is `ServerFlightGuard`'s flight-ordering check, which admits at most 3 messages of ~36 KiB each (one 18 KiB record plus the previous record's buffered tail), so future changes to the per-message size math know what actually protects the loop.

## Tests
- New: gun tests H/I/J in `grpc_test.rs`; `H2Stream` tests `shutdown_flushes_stashed_pending_write` and `shutdown_after_rejected_buffer_does_not_resurrect_payload` in `h2_common.rs`.
- Adjusted (behavior unchanged): tests F/G bind the hoarding server's new byte-collection receiver.
- All existing tests green; no pinned behavior changed for uncancelled writes.

## Regression bar
Ran from the worktree root with `CARGO_TARGET_DIR` unset — all pass:
- `cargo clippy --all-targets -- -D warnings` (default / `--no-default-features` / `--all-features`)
- `cargo test --lib` (plus `cargo test -p meow-transport --features grpc,h2 --lib` and `--test grpc_test` explicitly)

No relay-path code or footprint-tracked types (`Metadata`, `ConnectionInfo`, `UdpSession`, DNS cache entries) touched.

### DIFF
--- a/crates/meow-transport/src/grpc.rs
+++ b/crates/meow-transport/src/grpc.rs
@@ -288,7 +288,9 @@ struct GunStream {
     /// exactly once per logical write.  A retry after `Pending` must pass
     /// the same buffer (the `AsyncWrite` contract); `poll_write` enforces
     /// this via [`gun_frame_matches`] and rejects a changed buffer with
-    /// `InvalidInput` instead of silently sending the stale frame.
+    /// `InvalidInput` instead of silently sending the stale frame.  A frame
+    /// still stashed at `poll_shutdown` (i.e. the parked write was cancelled)
+    /// is flushed together with the closing EOS frame, never discarded.
     pending_write: Option<Bytes>,
 }
 
@@ -399,6 +401,14 @@ impl AsyncWrite for GunStream {
         cx: &mut Context<'_>,
         buf: &[u8],
     ) -> Poll<io::Result<usize>> {
+        // Mirror H2Stream: an empty write is a no-op reported as `Ok(0)`,
+        // checked *before* the changed-buffer guard so `write(&[])` while a
+        // frame is stashed does not trip `InvalidInput` (it is a distinct
+        // logical write, not a retry).  It also avoids sending a useless
+        // zero-payload gun frame.
+        if buf.is_empty() {
+            return Poll::Ready(Ok(0));
+        }
         let this = self.get_mut();
 
         // Encode buf into a gun frame exactly once per logical write.
@@ -458,10 +468,18 @@ impl AsyncWrite for GunStream {
 
     fn poll_shutdown(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<io::Result<()>> {
         let this = self.get_mut();
-        // Send empty DATA+EOS to signal end of the request stream.
-        this.send
-            .send_data(Bytes::new(), true)
-            .map_err(io::Error::other)?;
+        // Flush a frame stashed by a write that was cancelled at its Pending
+        // await before signalling end-of-stream: the caller handed those
+        // bytes to `write()` and they were encoded, so discarding them here
+        // would silently truncate the stream behind an empty DATA+EOS.  Only
+        // a cancelled write can leave `pending_write` set — the
+        // changed-buffer guard clears it before returning `InvalidInput`, so
+        // shutdown never resurrects a rejected frame.  `send_data` queues
+        // beyond the granted flow-control window inside h2, so this completes
+        // in a single poll (cancellation-safe) and is bounded to one frame.
+        let frame = this.pending_write.take().unwrap_or_default();
+        // DATA (possibly empty) + EOS signals end of the request stream.
+        this.send.send_data(frame, true).map_err(io::Error::other)?;
         Poll::Ready(Ok(()))
     }
 }
--- a/crates/meow-transport/src/h2_common.rs
+++ b/crates/meow-transport/src/h2_common.rs
@@ -126,7 +126,10 @@ pub struct H2Stream {
     /// `Ready(Ok(n))` the caller may legally submit a different buffer
     /// (issue #423).  Only granted capacity is ever handed to the
     /// connection, so a peer that stops reading applies real
-    /// backpressure instead of growing h2's internal buffer.
+    /// backpressure instead of growing h2's internal buffer.  A payload
+    /// still stashed at half-close (the parked write was cancelled) is
+    /// flushed together with the closing EOS frame by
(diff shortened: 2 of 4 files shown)


madeye:
Reviewed and verified against current `main` (merges clean; `shutdown_flushes_stashed_pending_write` and `shutdown_after_rejected_buffer_does_not_resurrect_payload` both pass). Three notes.

**1. The new REALITY comment is half wrong.** It says both constants are "defense-in-depth, not the load-bearing bound". That holds for `MAX_PRE_AUTH_HS_MESSAGES` (32 is unreachable — `ServerFlightGuard` admits at most 3 messages, each type once), but **not** for `MAX_PRE_AUTH_TRANSCRIPT_LEN`: at the comment's own ~36 KiB-per-message figure, two messages already exceed 64 KiB, so the check at `reality_tls.rs:917` fires *before* the ordering guard on a large Certificate flight. Since this PR is a documentation correction, worth getting exactly right — suggest splitting the claim per-constant.

**2. `best_effort_eos` is also called from `Drop`,** so this makes a dropped stream emit bytes the caller never received an `Ok` for — `Pending` means "nothing written" under the `AsyncWrite` contract. For a relay this is the right call (those bytes were already consumed from the client socket, so dropping them *is* truncation), but it's a semantic change and the doc comment currently frames it only as "never discarded". Worth stating that shutdown/drop may now deliver bytes that were never acknowledged to the caller.

**3. The `write(&[])` early return in `GunStream::poll_write` is a real bug fix** — previously an empty write while a frame was stashed would trip `gun_frame_matches` and return `InvalidInput`. Good catch; it deserves a line in the PR description, it's currently buried as "mirror H2Stream".

**Conflicts with #452.** `git merge-tree` reports a content conflict in `h2_common.rs` — both PRs rewrite the same `pending_write` doc comment and add tests to the same module. Whichever lands second needs manual resolution plus a re-run. Semantically they compose fine: #452 caps the stash at 64 KiB, this flushes it at EOS, so shutdown flushes at most a 64 KiB prefix.

madeye:
Rebased onto current `main` and resolved the conflict with #452, plus the REALITY comment correction from the review above. Two commits pushed.

**Conflict resolution.** #452 landed `WRITE_STASH_CAP` and the prefix-based changed-buffer guard while this branch was adding flush-stashed-payload-at-EOS, and both sides appended tests to the same module. git split it into three hunks around the client-handshake boilerplate that *both* sides' new tests contain, interleaving two unrelated test bodies — so rather than splice that by hand, I reset `h2_common.rs` to main's version and re-applied this branch's four discrete changes on top (doc note, `best_effort_eos` flush, the two test helpers, and the two shutdown tests lifted verbatim from the branch tip).

The two changes compose rather than merely coexist, which neither side's comment could have said: #452 bounds the stash to one window's worth, and this flush is the single place bytes reach h2 *beyond* the granted window — so the cap is exactly what keeps the flush bounded. The merged doc comment records that interaction instead of concatenating both halves.

All ten transport lib tests pass together, including main's `capped_stash_accepts_same_full_buffer_retry` / `large_write_accepts_at_most_stash_cap_per_poll` and this branch's two shutdown tests, unchanged from either side.

**REALITY comment.** Split per-constant as described above: `MAX_PRE_AUTH_HS_MESSAGES` (32) is genuinely unreachable behind the 3-message ordering guard, but `MAX_PRE_AUTH_TRANSCRIPT_LEN` (64 KiB) is the binding limit — two ~36 KiB messages already exceed it, so the transcript check rejects an oversized Certificate flight before the count reaches 3. The note now also says it must be revisited if the per-message size or record cap changes, since this is pre-auth code and the old wording would have licensed exactly that.

Full bar green locally: fmt, clippy ×3 (default / no-default-features / all-features), rustdoc, and the default-features test list.

The remaining review point — that `best_effort_eos` runs from `Drop`, so a dropped stream now emits bytes the caller never received an `Ok` for — is unchanged and still worth a second opinion before merge. It's the right call for a relay (those bytes were already consumed from the client socket), but it is a semantic change to the `AsyncWrite` contract.

PR #452, opened 2026-08-20
fix(transport): follow-up h2 write-path cleanups from #428/#441 review

madeye:
Follow-ups from the adversarial review of PR #441 and PR #428, classified minor/non-blocking at merge time and now fixed properly. All five findings were re-verified against current `main` before applying — all still applied.

## Follow-up to #441 review — `crates/meow-transport/src/h2_common.rs`

1. **Cap the per-poll `pending_write` stash at 64 KiB** (`WRITE_STASH_CAP`). Large writes under backpressure re-copied the whole shrinking remainder on every `write_all` resubmission (a 1 MiB write through a 64 KiB window: ~16 allocations / ~8 MiB memcpy). The stash — and therefore the bytes accepted per poll — is now bounded to one window's worth.
   - **Behavioral change**: `poll_write` on a buffer > 64 KiB now returns partial `Ready(Ok(n <= 65536))`. All in-tree callers audited: `H2Layer` consumers use generic `AsyncWrite` loops, h2mux `Stream` is driven by `write_all` and `PacketWriter`'s offset-tracking `write` loop — all partial-write-safe. gRPC's `GunStream` has its own separate guard and is untouched.
   - The cap (65536) deliberately exceeds h2's default 65535-byte initial window, so writes that fit the default window keep single-poll behavior and the existing "must park" test setups (128 KiB payloads vs 65535 window) still park — verified, not assumed: D6/D7, the h2mux changed-buffer test, and all #441 unit tests pass unchanged.
   - The changed-buffer Pending guard now compares the stash against the new buffer's **prefix of the stash's length** (the stash may be a capped prefix of the caller's buffer); a retry buffer shorter than the stash is rejected as changed.
2. **`reserve_capacity(0)` in the changed-buffer guard path** before the `InvalidInput` return, so the stale stash's h2 window reservation is handed back instead of staying requested on the error-tainted stream. (No direct assertion — h2 exposes granted, not requested, capacity; the path is exercised by D7 and the h2mux guard test.)
3. **Deadline on `partial_write_then_different_buffer_is_accepted`'s first poll/yield loop** — wrapped in `tokio::time::timeout(5s)` like the drop test, so a regression fails with a diagnostic instead of hanging CI.

## Follow-up to #428 review — h2 tests

4. **D7's final `poll_fn` no longer awaits the raw `poll_write`** — it uses the one-shot `Poll::Ready(match ...)` pattern D6 and `write_until_pending` already use, so a guard regression to accept-and-park fails the test instead of hanging it.
5. **Stalled-h2-server harness consolidated** into one shared source, `tests/support/h2_stalled.rs`: the `h2_common` unit tests include it via `#[cfg(test)] #[path]`, the integration suite via `support::h2_stalled` (gated `#[cfg(any(feature = "grpc", feature = "h2"))]` in `support/mod.rs`). The helper returns the raw h2 halves so both contexts wrap `H2Stream` under their own crate path; no assertion was weakened in either suite.

## New tests (pinning the cap)

- `capped_stash_accepts_same_full_buffer_retry` — a parked > 64 KiB remainder retried with the same full buffer must keep parking; an exact stash comparison would falsely reject every `write_all` resubmission of a large buffer.
- `large_write_accepts_at_most_stash_cap_per_poll` — a 1 MiB write against a draining server with 2 MiB windows is accepted in ≤ 64 KiB chunks per poll and completes end-to-end.

Existing tests updated only where the finding explicitly targeted them (finding 3: timeout wrapper; finding 4: one-shot poll; finding 5: harness relocation) — no pinned behavior was changed.

## Verification

Full regression bar green from a clean worktree (`CARGO_TARGET_DIR` unset):
`cargo fmt --all -- --check` ✓ · `cargo clippy --all-targets` (default / `--no-default-features` / `--all-features`) `-D warnings` ✓ · `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` ✓ · `cargo test --lib` (13 suites, 0 failures) ✓

Additionally: `cargo test -p meow-transport --features h2,grpc --test h2_test --test grpc_test --test grpc_unbounded_test` ✓ and `cargo test -p meow-proxy --features mux --lib mux::` (94 passed) ✓.

No ADR-0008 hot paths, key-type structs, or relay code touched.

### DIFF
--- a/crates/meow-transport/src/h2_common.rs
+++ b/crates/meow-transport/src/h2_common.rs
@@ -115,18 +115,28 @@ impl RecvState {
     }
 }
 
+/// Per-poll cap on the bytes copied into `pending_write`.  Without a cap,
+/// a large write under backpressure re-copies the whole shrinking
+/// remainder on every `write_all` resubmission (a 1 MiB write through a
+/// 64 KiB window costs ~16 allocations / ~8 MiB of memcpy); capping the
+/// stash bounds every copy to one window's worth.  64 KiB is chosen to
+/// exceed h2's default 65535-byte initial window, so writes that fit the
+/// default window keep their single-poll behaviour.
+const WRITE_STASH_CAP: usize = 64 * 1024;
+
 /// Raw bidirectional bytes over one HTTP/2 request/response pair.
 pub struct H2Stream {
     send: h2::SendStream<Bytes>,
     recv: RecvState,
     read_buf: Bytes,
     /// Payload stashed while a `poll_write` waits for h2 send-window
-    /// capacity.  Only retained across a `Poll::Pending` return — every
-    /// `Ready` return (including a partial one) clears it, because after
-    /// `Ready(Ok(n))` the caller may legally submit a different buffer
-    /// (issue #423).  Only granted capacity is ever handed to the
-    /// connection, so a peer that stops reading applies real
-    /// backpressure instead of growing h2's internal buffer.
+    /// capacity — at most [`WRITE_STASH_CAP`] bytes, i.e. possibly only a
+    /// prefix of the caller's buffer.  Only retained across a
+    /// `Poll::Pending` return — every `Ready` return (including a partial
+    /// one) clears it, because after `Ready(Ok(n))` the caller may legally
+    /// submit a different buffer (issue #423).  Only granted capacity is
+    /// ever handed to the connection, so a peer that stops reading applies
+    /// real backpressure instead of growing h2's internal buffer.
     pending_write: Option<Bytes>,
     remote_no_error_is_eof: bool,
     eos_sent: bool,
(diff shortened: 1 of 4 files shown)


madeye:
Reviewed and verified against current `main`: merges clean, `capped_stash_accepts_same_full_buffer_retry` and `large_write_accepts_at_most_stash_cap_per_poll` pass, h2/grpc integration suites 9+9 green.

**The prefix-match relaxation is sound, but the comment doesn't say why.** Going from `buf != data` to `buf.len() < data.len() || &buf[..data.len()] != data` looks like a weakened guard on first read. It isn't, and the reason is load-bearing: only `data.len()` bytes are ever committed and reported, so a retry buffer that shares the stash's prefix but differs beyond it cannot corrupt anything — the bytes sent are identical to `buf[..n]` either way. The `buf.len() < data.len()` rejection is the other half, since accepting a shorter buffer would over-report `n`.

That invariant is exactly what a future change to the reported length would silently break, so it should be written down next to the guard rather than left to be re-derived.

**Throughput risk is lower than the behavioral-change note implies** — worth recording so it doesn't get re-litigated. The relay path writes 4 KiB buffers, well under the 64 KiB cap, so the cap never binds there; it only engages for callers submitting > 64 KiB, and h2 fragments into 16 KiB frames internally regardless. I don't think this needs an ADR-0006 bench.

**Conflicts with #451** — `git merge-tree` reports a content conflict in `h2_common.rs` (both rewrite the `pending_write` doc comment and add tests to the same module). Whichever lands second needs manual resolution and a re-run.

Note this PR could not have gone green on its own before #464: it adds `connection.accept()` calls in an in-src test module, which tripped the ADR-0001 scan in `crate_invariants_test`. That's fixed on `main` now — verified this branch merged with `main` passes the guard 9/9.

PR #454, opened 2026-08-20
fix(dns): harden fakeip store compare-and-delete symmetry

madeye:
## Summary

Follow-up to #436 review. Two minor/non-blocking findings from the adversarial review of that PR, fixed properly now:

1. **`MemoryStore::put`'s cap-eviction path lacked the compare-and-delete guard that `del_by_ip` has.** When the by_ip or by_host LRU exceeds `cap`, the evicted pair was popped from the *other* side unconditionally. If a host had been remapped to a new IP shortly before its old IP aged out of the LRU as the cap-eviction victim, that unconditional pop could destroy the host's live forward mapping. Applied the same "only pop the other side if it still points at the just-evicted value" guard in both eviction directions, symmetric with `del_by_ip`.

2. **`FileStore::del_by_ip` called `mark_dirty()` even when the compare-and-delete guard found the reverse entry stale** (persisted `entries` map left unchanged), scheduling a debounced rewrite of an identical snapshot. Moved `mark_dirty()` inside the branch that actually mutates `entries`.

Both findings were re-verified against current `main` before fixing — they still applied exactly as described in review.

## Changes

- `crates/meow-dns/src/fakeip.rs`:
  - `MemoryStore::put`: cap-eviction now compare-and-deletes on both sides (mirrors `del_by_ip`).
  - `FileStore::del_by_ip`: `mark_dirty()` only fires when `entries` is actually mutated.
  - New test `memory_store_cap_eviction_spares_remapped_host` — reproduces the cap-eviction hazard directly (cap=2, remap a host, force by_ip cap eviction of the stale IP, assert the live forward mapping survives).
  - New test `file_store_del_by_ip_stale_does_not_mark_dirty` — isolates a stale `del_by_ip` call and asserts the dirty flag stays clear.

No pinned behavior changed; all existing fakeip tests remain green as-is.

## Test plan
- [x] `cargo test -p meow-dns --lib` (22/22 fakeip tests pass, including the 2 new ones)
- [x] `cargo test --lib` (full workspace, all green)

### DIFF
--- a/crates/meow-dns/src/fakeip.rs
+++ b/crates/meow-dns/src/fakeip.rs
@@ -123,13 +123,23 @@ impl Store for MemoryStore {
         // Evict the LRU pair from both directions so the two maps stay
         // consistent — bounded LRUs would evict each side independently.
         if inner.by_ip.len() > inner.cap {
-            if let Some((_, evicted_host)) = inner.by_ip.pop_lru() {
-                inner.by_host.pop(&evicted_host);
+            if let Some((evicted_ip, evicted_host)) = inner.by_ip.pop_lru() {
+                // Compare-and-delete: only drop the forward entry while it
+                // still points at the just-evicted ip. Same hazard as
+                // `del_by_ip` — if `evicted_host` was since remapped to a
+                // different address, the stale `by_ip` entry we're evicting
+                // here must not destroy that live forward mapping.
+                if inner.by_host.peek(&evicted_host) == Some(&evicted_ip) {
+                    inner.by_host.pop(&evicted_host);
+                }
             }
         }
         if inner.by_host.len() > inner.cap {
-            if let Some((_, evicted_ip)) = inner.by_host.pop_lru() {
-                inner.by_ip.pop(&evicted_ip);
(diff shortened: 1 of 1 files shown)


madeye:
Reviewed and verified: merges clean with `main`, 146 dns tests pass including both new fakeip tests.

Confirmed the hazard is real rather than theoretical — `MemoryStore::put` (`fakeip.rs:118`) deliberately leaves the old `by_ip` entry behind on a remap, so a cap-eviction victim genuinely can be a stale reverse entry whose host is live under a different IP. The fix mirrors `del_by_ip` exactly, and using `peek` rather than `get` is the right detail: a `get` there would have quietly perturbed LRU recency.

Also checked the obvious follow-on worry — skipping the paired pop can't leak, because each map's cap is enforced by its own independent check and every `put` adds at most one entry per side.

One nit: rustfmt has glued the "Isolate the del_by_ip call…" comment onto the trailing comment of the `s.put("h.example", ip2);` line, so it reads as attached to the wrong statement. A blank line before it fixes the rendering.

PR #455, opened 2026-08-20
config: follow-up nits from PR #445/#446 review (docs wording, short-id warn)

madeye:
## Summary

Follow-ups from adversarial code review of #445 and #446, classified minor/non-blocking at merge time.

- **Follow-up to #445 review** — `crates/meow-config/src/proxy_parser.rs`: the `parse_vless` Errors-list doc bullet on Vision + sing-mux was grammatically garbled (`` `smux`/`mux` protocol `smux`/`yamux`/`h2mux` ``). Reworded to: Vision + a `smux`/`mux` block using sing-mux (`protocol: smux`/`yamux`/`h2mux`).
- **Follow-up to #446 review** — `crates/meow-config/src/proxy_parser.rs`: a bare numeric `reality-opts.short-id` (e.g. `short-id: 0x1f`, unquoted) is silently coerced through `Number::to_string()` before hex-decoding (matching mihomo's own lossy int-to-decimal-string coercion), with no indication to the operator that this happened. Now emits a `tracing::warn!` naming the coerced value and telling the operator to quote the field if a different literal was intended.

## Status: partial (one finding's premise didn't reproduce; fixed the underlying case)

The #446 finding as reported described the bug as "leading-zero all-decimal short-ids (e.g. `short-id: 0012`) silently lose the leading zero through the `Number::to_string()` coercion." I verified this specific scenario does **not** reproduce with this codebase's `serde_yaml`: its core-schema int resolver only matches decimal scalars *without* a leading zero (bare `0` itself aside), so `short-id: 0012` parses as the plain string `"0012"` and never reaches the `Number` branch at all — it takes the pre-existing `as_str()` arm untouched, preserving the leading zero and the exact hex-decoded bytes with zero coercion involved. I confirmed this with a standalone probe test before writing anything (`"0012" -> String("0012")`, `"1234" -> Number(1234)`, `"0x1f" -> Number(31)`, `"007" -> String("007")`).

The underlying general risk the finding was gesturing at is real, just for a different trigger: `short-id: 0x1f` (unquoted) *does* hit the `Number` branch — it parses as decimal `31`, gets coerced to `"31"`, and hex-decodes to a different byte than the literal hex digits `"1f"` the operator likely intended. I implemented the requested warning against that real, verified case (any bare-number short-id, which covers `0x1f`-style notation), rather than forcing a warning onto a code path (`0012`) that plain strings already handle correctly and that a Number-coercion warning would never actually fire for.

Documented the discrepancy in the code comments and added:
- `parse_vless_reality_opts_leading_zero_short_id_preserved_no_warn` — pins that `short-id: 0012` is unaffected (string path, no warn, bytes preserved).
- A warn-assertion added to the existing `parse_vless_reality_opts_hex_literal_short_id_decimal_coerced` (`0x1f`) test, and to `parse_vless_reality_opts_numeric_short_id_coerced_to_string` (`1234`).

No behavioral change to previously-pinned test expectations beyond adding warn assertions to tests that already exercised the `Number` coercion path; no existing test was weakened or removed.

## Regression bar (all green, from worktree root, `CARGO_TARGET_DIR` unset)

- `cargo test --lib` (workspace-wide)
- `cargo test -p meow-config` (all suites, including 45/45 `vless_config_test` cases)

No hot-path invariant type (`Metadata`/`ConnectionInfo`/`UdpSession`/DNS `CacheEntry`) was touched — no byte-count delta applicable.

## Test plan

- [x] `cargo test -p meow-config --test vless_config_test` — 45/45 passing, including new/updated short-id tests
- [x] Full regression bar above

### DIFF
--- a/crates/meow-config/src/proxy_parser.rs
+++ b/crates/meow-config/src/proxy_parser.rs
@@ -1095,14 +1095,19 @@ fn parse_lb_strategy(strategy: Option<&str>) -> std::result::Result<LbStrategy, 
 /// - `uuid` invalid
 /// - `server` domain > 255 bytes
 /// - `vless-vision` feature absent + `flow: xtls-rprx-vision`
-/// - `flow: xtls-rprx-vision` + `smux`/`mux` protocol `smux`/`yamux`/`h2mux`
-///   — sing-box and Xray reject XTLS + sing-mux (`protocol: muxcool` is fine)
+/// - `flow: xtls-rprx-vision` + a `smux`/`mux` block using sing-mux
+///   (`protocol: smux`/`yamux`/`h2mux`) — sing-box and Xray reject XTLS +
+///   sing-mux (`protocol: muxcool` is fine)
 ///
 /// # Warn-once (Class B per ADR-0002)
 ///
 /// - `tls: false` with plain VLESS — plaintext, but correct destination
 /// - `mux: { enabled: true }` — sing-mux multiplexing (server must be sing-box/mihomo)
 /// - `flow: xtls-rprx-vision` + `udp: true` — Vision is TCP-only; UDP uses plain VLESS
+/// - `reality-opts.short-id` given as a bare YAML number (e.g. `0x1f`) —
+///   coerced to its decimal digits before hex-decoding (matching mihomo),
+///   which can silently reinterpret the value; quote it to preserve the
+///   literal digits
 #[cfg(feature = "vless")]
 fn parse_vless(
     name: &str,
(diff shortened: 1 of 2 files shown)


madeye:
Reviewed and verified: merges clean with `main`, 45 config tests pass.

The warn matches the established Class B pattern — every existing warn in `parse_vless` is a plain per-parse `tracing::warn!` rather than a process-global `Once`, so "warn-once" here consistently means once per proxy at parse time.

I treated the leading-zero claim as verify-don't-trust, since `"0012".parse::<i64>()` succeeds and a naive resolver would coerce it. `parse_vless_reality_opts_leading_zero_short_id_preserved_no_warn` passes, so serde_yaml does reject leading-zero decimals and `0012` stays a string. Good that it's pinned — a serde_yaml bump is exactly the kind of change that would flip this silently and lose the leading zero.

PR #456, opened 2026-08-20
fix(listener): align tproxy accept-error log level, add unbounded-cap test

madeye:
## Summary

Follow-up to #443 review. Both findings verified against current `main` and still applicable:

- **Log level mismatch**: `bounded_accept_loop`'s accept-error branch in `crates/meow-listener/src/tproxy/mod.rs` logged at `debug!`, while `mixed.rs`'s equivalent accept-error branch (which it explicitly mirrors, per its own doc comment) logs at `error!`. At default log levels this made fd-exhaustion events (EMFILE) on the tproxy listener invisible. Changed to `error!` to match `mixed.rs`.
- **Missing test coverage for the unbounded sentinel**: no unit test exercised `max_connections == 0` (the "disable the cap" sentinel) through `bounded_accept_loop`. Added `accept_loop_unbounded_when_max_connections_is_zero`, which admits 20 concurrent handlers at once with `max_connections = 0` — more than any small numeric cap (e.g. the existing `CAP = 3` bounded test) would ever allow — proving `0` truly means unbounded rather than "some larger limit."

## Behavioral change

Log level only (`debug!` → `error!` on tproxy accept errors); no change to control flow, the accept-error branch still logs-and-continues rather than propagating.

## Checks

Full regression bar run from a clean worktree with `CARGO_TARGET_DIR` unset:
- `cargo test --lib` (all crates, including the new and existing `tproxy::tests` — 18/18 in `meow-listener`)

All green. No `Metadata`/`ConnectionInfo`/`UdpSession`/DNS `CacheEntry` byte-size changes; relay-buffer stack-allocation invariant untouched (not in the diff).

## Test plan

- [x] `cargo test -p meow-listener --lib tproxy:: --features listener-tproxy` — 18 passed
- [x] `cargo test --lib` (workspace) — all green
- [x] Three-way clippy (default / no-default-features / all-features) — clean
- [x] `cargo doc --workspace --no-deps` with `-D warnings` — clean

### DIFF
--- a/crates/meow-listener/src/tproxy/mod.rs
+++ b/crates/meow-listener/src/tproxy/mod.rs
@@ -12,7 +12,7 @@ use std::net::{IpAddr, SocketAddr, ToSocketAddrs};
 use std::sync::Arc;
 use tokio::net::{TcpListener, TcpStream};
 use tokio::sync::Semaphore;
-use tracing::{debug, info, warn};
+use tracing::{debug, error, info, warn};
 
 /// Default cap on in-flight inbound connections per listener when the
 /// listener config doesn't set `max-connections` (mirrors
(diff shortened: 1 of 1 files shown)


madeye:
Reviewed and verified: merges clean with `main`, `accept_loop_unbounded_when_max_connections_is_zero` passes.

Confirmed the parity claim — `mixed.rs:127` does use `error!` for accept errors, so this aligns tproxy rather than diverging.

Minor, non-blocking: in the new test the `release` / `notify_waiters` handshake is effectively decorative. A handler can increment the counter (satisfying `all_in_flight`) before it registers on `release`, so `notify_waiters` may miss it — harmless here, because the `done_rx` drain counts client connects rather than handler completions and is timeout-wrapped, and the final assertion reads `max_observed`, which is already at `CLIENTS` by then.

Issue #426, opened 2026-08-19
mux: record ADR-0007 binary-size delta for the default profile


madeye:
The #412 mux stack (#414 → #418) landed with no binary-size measurement in any
PR body. It adds the `mux` feature to `meow-app`'s `full` bundle — which is
`default = ["full", "boring-tls"]` — plus three new dependencies:

- `tokio-util` (#414)
- `yamux` 0.14 (#416)
- `h2` 0.4 + `meow-transport/h2` (#417)

and roughly 5.5 kLoC of new code across `crates/meow-proxy/src/mux/`.

ADR-0007 (`docs/adr/0007-m2-footprint-budget.md`) §Amendment requires a feature
landing to carry the delta and name the cap it threatens:

| Target | `minimal` cap | `default` cap |
|---|---|---|
| `aarch64-unknown-linux-musl` | 8 MiB | **18 MiB** |
| `mipsel-unknown-linux-musl` | 7 MiB | **16 MiB** |
| `x86_64-unknown-linux-musl` | 8 MiB | **20 MiB** |

`minimal` is unaffected (mux is excluded from it by design), but `default` is
directly affected and was never measured.

**Action:** measure stripped `default` size on the three targets at the current
main tip (3b57478), record the delta against the last baseline in
`docs/benchmarks/`, and open an ADR-0007 amendment if any cap is threatened.

Note #421 first — mipsel does not currently compile with `mux` enabled.

madeye:
@aimkiray — owner, please. Covers the whole #414–#418 stack rather than one PR. Please run it after #421, since `mipsel` will not build with `mux` enabled until that lands.

Issue #421, opened 2026-08-19
mux: use meow_common::atomic::AtomicU — AtomicU64 breaks the MIPS32 build


madeye:
The `mux` module uses `std::sync::atomic::AtomicU64` directly instead of the
workspace's platform-adaptive alias `meow_common::atomic::AtomicU`.

Sites:
- `crates/meow-proxy/src/mux/client.rs:22,300,544` — `MuxSession::last_used_ms`
- `crates/meow-proxy/src/mux/packet.rs` and `mux/muxcool.rs` (test constructors)

`mipsel-unknown-linux-musl` has no 64-bit atomics, so `AtomicU64` does not exist there:

```
$ rustc --print cfg --target mipsel-unknown-linux-musl | grep atomic
target_has_atomic="16"
target_has_atomic="32"
target_has_atomic="8"
target_has_atomic="ptr"
```

`mux` is in `meow-proxy`/`meow-config` `default` and therefore in `meow-app`'s
`default = ["full", "boring-tls"]`, so the whole default build fails to compile
for that target. Not caught by CI because the current release matrix (b4c7006)
contains no 64-bit-atomic-less target, but mipsel is still a named ADR-0007
size target and this is the exact failure fixed once before in
5ef7636 *"fix(anytls): don't assume 64-bit atomics (mips32 release build)"*.

**Fix:** switch to `meow_common::atomic::{AtomicU, Uint}`, and follow the alias
module's rule for the millisecond clock — compare in the truncated domain with
`wrapping_sub`, never widen to `u64` first (see `UdpSession::idle_for`).

**Related:** `MuxClient::now_ms()` reads `SystemTime`, so a backwards wall-clock
step makes `saturating_sub` return 0 and defers idle eviction. Moving to a
monotonic base would fix both issues at once.

Follow-up from the review of the #412 mux stack.

madeye:
@aimkiray — owner, please, as follow-up to #414 (`mux/client.rs`). This is the highest-priority item of the seven: `mipsel-unknown-linux-musl` currently fails to compile with the `mux` feature, which is on by default. The fix is a type-alias swap to `meow_common::atomic::{AtomicU, Uint}`, with the `wrapping_sub` comparison rule for the ms clock. 5ef7636 is the precedent from the anytls case.

PR #457, opened 2026-08-20
docs(benchmarks): re-check mipsel binary-size blocker after #421 fix

madeye:
## Summary

Follow-up to #448 review (group: mipsel-size-measurement, from adversarial code review of PRs #414–#418 / #448).

- Finding: the post-mux ADR-0007 measurement (PR #448, issue #426) skipped `mipsel-unknown-linux-musl` because #421 (mux used raw `std::sync::atomic::AtomicU64` instead of `meow_common::atomic::AtomicU`, breaking the mipsel32 build) was unmerged at the time. #421 has since been fixed and merged (#447), so that specific blocker is gone — verified `crates/meow-proxy/src/mux/` now uses `meow_common::atomic::AtomicU` exclusively, no raw `AtomicU64` remains.
- However, re-attempting the measurement with the repo's established method (`cargo zigbuild --release --target mipsel-unknown-linux-musl` on stable, same profile/flags as the other targets in `docs/benchmarks/binary-size.md`) still fails — for a *different* reason than #421: `mipsel-unknown-linux-musl` is a Tier 3 target and rustup ships no prebuilt `std` for it on any host (`rustup target add mipsel-unknown-linux-musl` → "no prebuilt artifacts available ... low-tier target").
  - Confirmed the only way to produce a `std`-linked binary at all is `-Zbuild-std` on nightly (locally verified `cargo +nightly zigbuild -Zbuild-std=std,panic_abort --release --target mipsel-unknown-linux-musl` at least starts compiling `core`/`alloc`/`std` from source). That's a materially different toolchain/build recipe than every other row in the doc (nightly vs. the workspace's pinned stable, locally-built std vs. distributed), so a size measured that way wouldn't be comparable — and it's not this repo's established cross-compile method. Per group guidance, documenting this rather than forcing it.
  - This also isn't a regression from #412/mux: CI's own `release.yml` doesn't build MIPS at all (excluded together with 32-bit musl/windows-gnu because they can't link `boring-sys`, part of the shipped `default` feature set), and ADR-0007 already documents mipsel as soft-gated for exactly this class of reason ("no mipsel cross-compile infra... no way to functionally validate a mipsel binary").

**Status: partial.** #421's blocker is confirmed resolved and documented; the mipsel `default`-profile size remains genuinely unmeasurable with the toolchain available here (and in CI), so the row stays `not measured` with the concrete blocker spelled out instead of reporting a non-comparable nightly-build-std number. The soft-gated mipsel cap (ADR-0007 §2, 16 MiB) is therefore still unverified — this was true before this follow-up too (blocked by #421) and remains true now (blocked by the Tier 3 toolchain gap); ADR-0007 already defers a hard gate to M3 for this reason.

Doc-only change, no code touched, no behavior change, no ADR amendment needed (no cap breached on the two targets that *are* measured).

## Test plan

- [x] Read current code: confirmed no raw `AtomicU64` remains under `crates/meow-proxy/src/mux/` (only `meow_common::atomic::AtomicU`)
- [x] Confirmed #421 closed / #447 merged via `gh issue view` / `gh pr view`
- [x] Reproduced the Tier 3 toolchain gap locally (`rustup target add mipsel-unknown-linux-musl` fails; `-Zbuild-std` on nightly at least starts building std, confirming it's the only path and it's out of scope for this doc's methodology)
- [x] `cargo test --lib` (all crates green)

### DIFF
--- a/docs/benchmarks/binary-size.md
+++ b/docs/benchmarks/binary-size.md
@@ -101,11 +101,50 @@ in the `full` bundle (`meow-app/Cargo.toml`), so it is part of every
 `default` build; it is excluded from `minimal` by design (ADR-0007 §1) and
 that profile is unaffected.
 
-**mipsel-unknown-linux-musl not measured**: blocked by #421 (`mux` uses
-`std::sync::atomic::AtomicU64` directly instead of
-`meow_common::atomic::AtomicU`, and mipsel has no 64-bit atomics — the
-`default` build does not compile for this target until #421 lands). Follow
-up once #421 merges.
+**mipsel-unknown-linux-musl still not measured, different reason now**:
+#421 (`mux` used `std::sync::atomic::AtomicU64` directly instead of
+`meow_common::atomic::AtomicU`) was fixed by #447, so the compile-time
+blocker is gone — `crates/meow-proxy/src/mux/` now uses
+`meow_common::atomic::AtomicU` exclusively (verified: no raw `AtomicU64`
+remains under `mux/`). But the target still cannot be measured with this
+repo's established method (`cargo zigbuild --release --target
+mipsel-unknown-linux-musl`, same as the other targets in this doc):
+`mipsel-unknown-linux-musl` is a **Tier 3** target
+([rustc platform support](https://doc.rust-lang.org/rustc/platform-support.html)) —
+rustup ships no prebuilt `std` for it on any host:
+
+```
+$ rustup target add mipsel-unknown-linux-musl
+error: toolchain 'stable-aarch64-apple-darwin' has no prebuilt artifacts
+available for target 'mipsel-unknown-linux-musl'
+note: this may happen to a low-tier target as per
+https://doc.rust-lang.org/nightly/rustc/platform-support.html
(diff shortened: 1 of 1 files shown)


madeye:
Reviewed and verified. Both load-bearing claims check out: no raw `AtomicU64` remains under `crates/meow-proxy/src/mux/`, and the cited `docs/adr/0007-m2-footprint-budget.md` resolves.

Reclassifying the blocker from #421 to the Tier 3 toolchain gap is the accurate call, and declining to publish a nightly `-Zbuild-std` number that isn't comparable to the other rows is the right instinct — a size measured under a different toolchain with a locally-built `std` would be worse than `not measured`, because it would look like data.

Unrelated, found while checking the reference: `CLAUDE.md`'s Architecture Invariants table has three dead ADR links — `0006-performance-targets.md`, `0007-binary-size-caps.md` and `0008-zero-alloc-invariants.md` don't exist; the real files are `0006-m2-benchmark-methodology.md`, `0007-m2-footprint-budget.md` and `0008-m2-allocator-audit.md`. Worth a one-line fix in a separate PR.

Issue #429, opened 2026-08-19
config: rule-provider `path:` is written without any containment check — `PUT /configs` gives arbitrary file write


madeye:
## Summary

`rule-providers[<name>].path` is used as a **write** target for HTTP-fetched payloads with no validation that it stays inside a provider cache directory. An absolute path is returned verbatim; a relative path with no `cache_dir` is likewise used as-is. Because the runtime rebuild path (`rebuild_from_raw_runtime`) hardcodes `cache_dir = None` and is fed a fully caller-supplied YAML document by `PUT /configs`, an API caller can make the process fetch an arbitrary URL and write the raw response bytes to any path the process can write — creating missing parent directories along the way.

Upstream mihomo guards exactly this with a "path is not subpath of home directory" check (`C.Path.IsSafePath` / `errSubPath`) for both rule- and proxy-providers; meow-rs has no equivalent (`grep -rn "is_safe\|subpath\|canonicalize" crates/meow-config/src` → no hits).

Reproduced with a throwaway integration test that calls the public `meow_config::rebuild_from_raw_runtime` against a local mock HTTP server: the server's response bytes landed verbatim in an absolute path outside any cache dir, with the missing parent directory auto-created.

## Affected code

`crates/meow-config/src/rule_provider.rs:530-543` — path resolution, no containment check:

```rust
fn resolve_path(cfg: &RawRuleProvider, cache_dir: Option<&Path>, name: &str) -> Option<PathBuf> {
    if let Some(p) = cfg.path.as_deref() {
        let path = PathBuf::from(p);
        if path.is_absolute() {
            return Some(path);          // <-- absolute path used as-is, cache_dir ignored
        }
        return Some(match cache_dir {
            Some(dir) => dir.join(path), // <-- still no traversal check ("../../etc/...")
            None => path,                // <-- relative path used as-is
        });
    }
    let dir = cache_dir?;
    Some(dir.join("rule-providers").join(format!("{name}.yaml")))
}
```

`crates/meow-config/src/rule_provider.rs:215-229` — that path becomes the cache/write target for `type: http`:

```rust
"http" => {
    let url = cfg.url.as_deref()
        .ok_or_else(|| anyhow!("http provider '{name}' requires a 'url'"))?;
    let cache_path = resolve_path(cfg, cache_dir, name);
    let prefer_cache = cfg.interval.unwrap_or(0) > 0;
[...]

PR #458, opened 2026-08-20
fix(config): provider-path polish follow-ups from PR #444/#437 review

madeye:
Follow-up to #444 review and follow-up to #437 review: a group of minor findings classified non-blocking at merge time, now fixed properly. All four findings were re-verified against current `main` before fixing and still applied.

### 1. Stray `path:` on inline rule-providers is ignored (#444 review)
`resolve_path`/`validate_paths` used to hard-error on a `path:` attached to an inline provider when no cache dir was available (and containment-check it when one was), even though inline providers never read the path. mihomo ignores a stray `path` on inline; we now warn-and-ignore it for any provider type that never reads it, so a harmless leftover field can't fail a whole rebuild. Test: `stray_path_on_inline_provider_is_ignored`.

### 2. `SKIP_SAFE_PATH_CHECK` escape hatch (#444 review)
The #429/#444 containment check shipped with no migration path for configs whose provider paths legitimately live outside the cache dir. Mirroring mihomo (`allowUnsafePath, _ := strconv.ParseBool(os.Getenv("SKIP_SAFE_PATH_CHECK"))`): when the env var holds a Go-ParseBool-truthy value (`1`/`t`/`T`/`true`/`TRUE`/`True`), a containment failure is downgraded to a prominent `warn!` (naming the path, the root, and how to restore the protection) and the out-of-root path is allowed.

**Security posture**: defaults OFF — requires the explicit env var; unset/empty/falsy values keep the containment guarantee fully intact. Both states are tested (`escape_hatch_off_rejects_out_of_root_paths`, `escape_hatch_on_allows_out_of_root_paths_with_warning`, `escape_hatch_on_allows_symlink_escape`, `escape_hatch_env_parsing_matches_go_parse_bool`) via bool injection so tests can't race on the process environment.

### 3. Proxy-provider out-of-root paths fail loudly (#444 review)
Previously an escaping proxy-provider path was warn-skipped in `load_proxy_providers`, silently degrading every group referencing the provider (each group only logged a generic "provider not found"). New `proxy_provider::validate_paths` runs in `rebuild_from_raw_impl` next to the rule-provider validation, so the rebuild fails hard with an error naming the provider — rule-provider parity. Scope is containment violations only; other per-provider problems (missing fields, unknown type, fetch failures) keep their historical warn-and-skip semantics. Tests: `validate_paths_rejects_escaping_provider_paths`, `validate_paths_accepts_contained_and_pathless_providers`, `rebuild_rejects_proxy_provider_path_escaping_cache_dir`.

**Behavioral note**: a config that previously *started degraded* (provider skipped, groups missing members) now fails the build/rebuild with a clear error. The new escape hatch (item 2) is the migration path for intentional out-of-root paths.

### 4. 'ignoring path' warn fires once per provider per rebuild (#444 review)
`validate_paths`, `prefetch_payloads`, and `load_http` all resolve the same path, so the "ignoring 'path' (no provider cache directory)" warn fired up to three times per http provider per rebuild. `resolve_path` now takes an `emit_warnings` flag; only `validate_paths` — which every (re)build runs exactly once, before any fetch/load pass — emits the warns.

### 5. End-to-end body-size-cap test (#437 review)
The existing cap tests exercised only the extracted `response_bytes_capped` helper. New `fetch_http_async_rejects_oversized_response_end_to_end` drives `fetch_http_async` against a local server declaring `Content-Length: MAX_BODY_BYTES + 1` (nothing near 256 MiB is transferred — the pre-check rejects on the header) to pin the actual wiring through `response_bytes_with_limit`.

## Tests
- No existing test was weakened; all pre-existing tests pass unchanged (only `resolve_path` test call sites gained the new `emit_warnings` argument).
- Full regression bar green locally: `cargo fmt --check`, three-way clippy (`-D warnings`), `cargo doc` (`-D warnings`), `cargo test --lib` (13/13 crates, 0 failures).

No hot-path, relay, or footprint-tracked types touched (ADR-0006/0007/0008/0011 not in scope).

### DIFF
--- a/crates/meow-config/src/lib.rs
+++ b/crates/meow-config/src/lib.rs
@@ -688,12 +688,17 @@ fn rebuild_from_raw_impl(
     // groups and provider-sourced proxies included (issue #377).
     let registry_lookup = |name: &str| proxies.get(name).cloned();
 
-    // Fail hard on any rule-provider path that would escape the provider
-    // cache directory — before any fetch or on-disk write happens, so a
-    // hostile `PUT /configs` is rejected without touching the filesystem
-    // (issue #429).
+    // Fail hard on any rule- or proxy-provider path that would escape the
+    // provider cache directory — before any fetch or on-disk write happens,
+    // so a hostile `PUT /configs` is rejected without touching the
+    // filesystem (issue #429). Proxy-providers get the same loud failure as
+    // rule-providers (PR #444 review follow-up) instead of being warn-skipped
+    // with every group referencing them silently degrading.
     if let Some(map) = raw.rule_providers.as_ref() {
         rule_provider::validate_paths(map, cache_dir)?;
+    }
+    if let Some(map) = raw.proxy_providers.as_ref() {
+        proxy_provider::validate_paths(map, cache_dir).map_err(|e| anyhow::anyhow!("{e}"))?;
     }
 
     // Fetch/read rule-provider payload bytes once — the parser-context build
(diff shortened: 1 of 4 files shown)


madeye:
Reviewed and verified: merges clean with `main`, all four `safe_path` escape-hatch tests pass.

The proxy-provider hard-fail is a straight improvement — rule-provider parity, and warn-skipping a containment violation while every group referencing it silently degrades was the worse failure mode.

**`SKIP_SAFE_PATH_CHECK` deserves an explicit decision rather than review-by-default,** because it re-opens the containment fix from #429 that merged a day earlier. The implementation is careful — defaults off, `go_parse_bool_truthy` correctly mirrors Go's `strconv.ParseBool` truthy set (`1|t|T|true|TRUE|True`), loud `warn!`, and the tests inject the bool instead of racing on process-global env. My reservation is scope, not correctness:

- The hatch is process-global, so it also covers the `PUT /configs` path. An operator who enables it for one legitimate local file provider silently re-opens arbitrary-path writes through the REST API — a different threat model from the config-file case it's meant to serve.
- It's documented only in a doc comment on a `pub(crate)` fn. Nobody who needs it will find it, and nobody who enables it will find the caveat. It should be in user-facing docs alongside the provider path rules.

mihomo parity is a legitimate justification under ADR-0002, so this isn't a blocker — but if the goal is only "migration path for configs pointing outside the cache dir", a narrower hatch (startup config only, not runtime `PUT /configs`) would give the same migration path without widening the API surface.

PR #462, opened 2026-08-20
fix(dns): pool plain tcp:// exchanges to avoid per-query reconnect

### LINKED ISSUE #460: DNS: per-query TCP reconnect on plain tcp:// upstreams costs an Android protect(fd) call per lookup and starves resolution
## Problem

Every plain `tcp://` DNS exchange in `meow-dns/src/client.rs` opens a fresh
TCP connection (`tcp_exchange` → `factory().connect_tcp` →
`meow_common::connect_tcp`), uses it once, and drops it. On Android each
`connect_tcp` additionally triggers a `VpnService.protect(fd)` JNI call
before the socket can be used. Under a page-load burst this means a full
TCP handshake + `protect(fd)` per DNS lookup, which:

- serialises many lookups on fresh handshakes and starves the proxy's own
  server-hostname resolution (the proxy can't dial nodes it hasn't
  resolved yet);
- surfaces as intermittent DNS timeouts on Android.

## Proposal

Reuse a small bounded pool of keep-alive plain `tcp://` connections per
`DnsClient` (RFC 7766 allows multiple queries per TCP connection). Scope is
plain `tcp://` only — DoT/DoH keep their per-query TLS handshake.

Concrete shape (small, self-contained, ~1 file):
- A `TcpPool` of up to 4 idle `TcpStream`s + a `Semaphore` bounding
  concurrency to 4, on `DnsClient`.
- `exchange()`'s `Transport::Tcp` arm checks out a pooled stream for the
  whole exchange and returns it only after a fully validated response.
- Cancellation safety: the checked-out stream is local to the exchange
  future, so a dropped/timed-out query discards the partially consumed
  stream instead of returning a framing-desynced stream to the pool.
- A reused stream runs under a short read deadline (timeout/2) so an
  idle-half-closed connection fails fast and retries on a fresh connect
  within the query budget; malformed short frames never re-enter the pool.

No public API change: `query` / `lookup_ip` / `exchange` keep their existing
signatures; resolver/cache/server/config are untouched.

## Scope note

Self-contained TCP-pooling slice split out of #387. The broader prefer-IPv4
/ per-family-cache / IPv6-policy rework in #387 is intentionally not
included here and is pursued in a separate issue/PR.

aimkiray:
Resolves #460.

## Problem
Each plain `tcp://` DNS exchange opens a fresh connection and (on Android)
pays a `VpnService.protect(fd)` JNI call per lookup; under a page-load
burst this starves the proxy's own server-hostname resolution and surfaces
as intermittent DNS timeouts. See #460 for details.

## Approach
A bounded keep-alive `TcpPool` (4 idle `TcpStream`s + a `Semaphore` of 4)
on `DnsClient`. `exchange()`'s `Transport::Tcp` arm checks out a pooled
stream for the whole exchange and returns it only after a fully validated
response. Concurrency is bounded to 4; the async mutex only guards the
idle-slot pop/push, so up to four exchanges run concurrently on separate
streams while a fifth caller waits for a permit.

Cancellation safety: the checked-out stream stays local to the exchange
future, so a dropped/timed-out query discards the partially consumed stream
instead of returning a framing-desynced stream to the pool. A reused stream
runs under a short read deadline (timeout/2) so an idle-half-closed
connection fails fast and the fresh-connect retry still fits within the
outer per-query timeout; malformed short frames and any error/timeout on a
reused stream never re-enter the pool.

Pooling is scoped to the direct `Transport::Tcp` arm of `exchange()` only.
DoT/DoH keep a full TLS handshake per query, and the UDP-truncation TCP
fallback keeps its existing one-shot `tcp_exchange`, so both are out of
scope.

No public API change: `query` / `lookup_ip` / `exchange` keep their existing
signatures; resolver/cache/server/config are untouched.

## Verification
- three-way `cargo clippy --all-targets -- -D warnings` (default /
  no-default-features / all-features)
- `cargo test -p meow-dns --lib` (incl. the three new pooling tests:
  `tcp_short_frames_never_panic_or_return_to_pool`,
  `cancelled_tcp_exchange_discards_partial_stream`,
  `tcp_pool_bounds_concurrency_and_reuses_connections`)
- `cargo test --test rules_test`, `--test trojan_integration`

Diff: 1 file, `crates/meow-dns/src/client.rs`, +251/-4.


madeye:
Reviewed and verified: merges clean with `main`, 147 meow-dns tests pass.

The core design is right, and I went looking specifically for the classic pooling hazard — a stream desynced by a cancelled exchange being handed to the next query. It's handled correctly: streams are returned to the pool only after a fully validated response, so both the cancellation path (future dropped, local `stream` dropped with it) and every error path drop the stream instead of pooling it. `decode_validated_response`'s ID check backstops it.

Two findings.

**1. Pooled streams are never reaped — this is the significant one.** There's no idle timer, so a stream can sit in `idle` indefinitely. An upstream or NAT can drop it without an RST or FIN ever arriving: the write still succeeds and the peer simply never answers, so the reuse attempt burns `pooled_budget` = `timeout / 2` = **2.5 s at the default timeout** before falling back to a fresh connect. That makes the first query after any quiet period *worse* than the per-query reconnect this PR replaces, and silent-drop is most likely on exactly the mobile/NAT networks the PR targets.

I've implemented idle reaping on top of this branch and will open it against this PR shortly: stamp each stream on return, close anything past a 30 s idle window on checkout, reuse the freshest survivor. The reaping rule is split into a pure generic helper so it's testable without a socket or a clock, plus an end-to-end test asserting reuse inside the window and a fresh connect past it. Your existing `tcp_pool_bounds_concurrency_and_reuses_connections` still passes unchanged.

**2. `TCP_POOL_CAPACITY` doubles as pool size and concurrency cap.** The semaphore turns previously unbounded TCP DNS concurrency into a 4-deep queue, so a 30-domain page load serializes into waves rather than running in parallel. I understand the coupling is what bounds `idle` — but a bounded push (`if idle.len() < CAP { push }`) with unbounded fresh connects would decouple the two knobs and keep burst latency flat. Raising rather than changing, since it's a design call on your PR: was the concurrency cap intended, or a side effect of bounding the pool?

madeye:
Idle reaping is up as aimkiray/meow-rs#1, targeting this branch so it lands inside this PR rather than racing it. One commit; full bar green locally, and `tcp_pool_bounds_concurrency_and_reuses_connections` passes unchanged.

aimkiray:
Thanks for the thorough review — both findings are spot-on.

**1. Idle reaping — merged.**

Your stacked PR is fast-forwarded into this branch as `5d4eb18`. Full bar green locally: `fmt`, three-way `clippy`, `doc`, and the full `--test` target list. All TCP-pool tests pass, including the new `take_fresh` unit tests and the end-to-end reaping test. The original `tcp_pool_bounds_concurrency_and_reuses_connections` still passes unchanged.

The layering is clean: `TCP_POOL_IDLE_TIMEOUT` (30 s, per RFC 7766 §6.2.3) is the first line of defence — a stream idle past the window is closed on checkout, so the common path after a quiet period is a plain connect rather than a reuse attempt that has to time out first. The existing `pooled_budget = timeout / 2` becomes the second line for the edge case where a stream goes silently dead *inside* the idle window. `take_fresh` as a pure helper is a nice separation.

**2. `TCP_POOL_CAPACITY` doubling as concurrency cap — you're right, it's a side effect.**

The semaphore was introduced to bound the pool size safely (you need a permit to push), and the concurrency cap came along for free. The doc comment describes the *mechanism* but the *why* was never a deliberate design call.

That said, I think 4 is a reasonable default for this specific path and I'd prefer to keep it for now:

- **Plain `tcp://` DNS is a low-frequency path.** Most DNS traffic uses UDP; TCP is for truncated responses, zone transfers, or explicitly configured `tcp://` upstreams. Even with `tcp://` as the only upstream, 4 concurrent exchanges cover the typical query rate — DNS queries are small and fast (~1–5 ms on a warm connection), so the semaphore rarely blocks in practice.
- **It protects the upstream.** Unbounded concurrency on a single `tcp://` upstream means a burst of 30 domain lookups opens 30 simultaneous TCP connections to the same resolver. On mobile/NAT networks (the exact scenario this PR targets) that can trigger connection-rate limits or NAT port exhaustion.
- **The cap is a constant, not a hard limit.** If real-world data shows 4 is too low for a specific workload, raising `TCP_POOL_CAPACITY` or splitting it into separate `POOL_SIZE` / `MAX_CONCURRENCY` constants is a one-line follow-up. I'd rather not over-engineer the knob before we have evidence.

If you'd prefer to decouple now rather than defer, I'm happy to do it — but I'd want to see the concurrency default set high enough (e.g. 32) that it doesn't regress burst latency, with the pool size staying at 4. That's effectively your `if idle.len() < CAP { push }` with a generous semaphore.

I've also added a doc comment on `TCP_POOL_CAPACITY` clarifying that the semaphore is a pool-size bound first and a concurrency cap second, so the coupling is explicit rather than implicit.

aimkiray:
Took a deeper look at the pooling + reaping implementation before this lands. No blockers found; summarizing the check below.

**Correctness**
- **Permit accounting is leak-free.** `_permit` is RAII; every return path (success, `?` early-return, inner-timeout, cancellation) releases it. No `forget()`/stranding path exists.
- **Pool size is provably ≤ 4.** Each exchange holds ≤1 permit and pushes ≤1 stream while the permit is live, so `idle.len() ≤ TCP_POOL_CAPACITY` regardless of reaping/concurrency. `take_fresh` only removes entries and runs inside the mutex critical section (no `.await`), so checkout + reap is atomic.
- **Cancellation safety holds.** A checked-out stream is a local; dropping the future closes the (possibly half-consumed) stream and returns the permit — it never re-enters the pool. `cancelled_tcp_exchange_discards_partial_stream` covers this.
- **No framing desync.** A stream is returned only after a fully validated response (`decode_validated_response`); ID/type/question checks backstop it. Short/malformed frames drop the stream (`tcp_short_frames_never_panic_or_return_to_pool`).
- **Lock ordering is safe.** Semaphore is acquired before the mutex, the critical section is sync, so no hold-and-wait cycle.

**Scope/regressions**
- Only direct `Transport::Tcp` is pooled; UDP, UDP-truncation TCP fallback, DoT/DoH, and proxy-routed `tcp://` keep their prior paths. No public API change.
- Two accepted edge behaviors (both bounded by the outer timeout, neither hangs):
  1. Under ≥5 concurrent queries on one client, a caller can spend budget in `acquire()` so the pooled attempt eats most of the window — the documented concurrency-cap tradeoff.
  2. A silently-dropped-but-<30 s-idle stream + an upstream needing `(timeout/2, timeout)` can flip a prior `Ok` to `Timeout`. Rare (reaper + pathological-latency combo); the warm-reuse common case is strictly faster.

**Verification**
- `cargo test -p meow-dns --lib`: 151 passed; `--test proxy_routed_dns`: 1 passed; `fmt` + `clippy -p meow-dns --all-targets -D warnings` clean. TCP-pool/cancellation/reaping tests stable across 10 repeats. CI on the PR is all green (lint/test/features/MSRV/macOS/windows/openwrt/tproxy).

Non-blocking follow-ups I noted: derive `pooled_budget` from the remaining outer deadline instead of `timeout/2`; lazily init `TcpPool` only for `Transport::Tcp` clients; split `POOL_SIZE`/`MAX_CONCURRENCY` if real-world load shows 4 is too low.

Ready to merge from my side whenever you're happy with it. Want me to hit merge, or would you prefer to land it yourself?

aimkiray:

Verified every finding in the review against this PR's code and pushed three fix-ups (`0777a1c`, `ac0e5a2`, `b7638d5`). All fast-forward pushes; no history rewritten.

### Round 1 (`0777a1c`)

- **H1 (confirmed, fixed):** the semaphore permit was held across the whole exchange, so concurrent exchanges collapsed onto `TCP_POOL_CAPACITY = 4` — a `tcp://` primary nameserver queued page-load bursts into the 5 s outer timeout. The semaphore is removed entirely; the idle bound is enforced in `TcpPool::push` (`idle.len() < CAPACITY` checked and the stream pushed under the same lock — no check-then-use race). Surplus streams are dropped (connection closed) on return.
- **H2 (confirmed, fixed):** the stale-reuse read share was `timeout / 2`. Now `min(timeout / 2, 250 ms)` so a fresh-dial fallback always has budget left.
- Low items: the dead `AcquireError` handling left with the semaphore is gone; `tcp_exchange` was folded into `tcp_message_exchange` (the UDP-truncation fallback now does one explicit `connect_tcp` + exchange, not pooled); the test was rewritten to prove both directions — a burst of `2 × TCP_POOL_CAPACITY` concurrent queries dials all of them immediately (`accepted == burst`), the idle pool stays ≤ 4, and the query right after the burst reuses a pooled stream.

### Round 2 (`ac0e5a2`, docs)

- **N1 (confirmed — design trade-off):** with concurrency uncapped, only pools of ≤ 4 idle streams are kept, so a *fully concurrent* burst larger than the capacity pays a fresh connect per query. The capacity doc already said this; the contradictory sentence in the `TCP_POOL_IDLE_TIMEOUT` comment ("a burst of queries still shares connections") was corrected, and `TCP_POOL_CAPACITY`'s doc gained a tuning note — reuse (and Android `protect(fd)`) costs return for bursts above the capacity; raising the idle cap to 16–32 would restore it. Left at 4 for the maintainer to tune.

### Benchmark quantifying the capacity trade-off (`b7638d5`)

Added `tcp_pool_burst_connection_budget` (`#[ignore]`; run with `cargo test -p meow-dns --lib tcp_pool_burst -- --ignored --nocapture`). Each phase uses a fresh client (empty pool), local answering DNS-over-TCP server, live connection counting:

| Arrival pattern | Queries | New connects | Reuse |
| --- | --- | --- | --- |
| concurrent burst | 4 / 16 / 32 / 64 | 4 / 16 / 32 / 64 | 0% |
| staggered @ 5 ms | 64 | **1** | **98%** |
| staggered @ 20 ms | 64 | **1** | **98%** |

Interpretation for the maintainer: the reuse benefit holds for any arrival that finds an idle pooled stream — which is the common case for millisecond-scale DNS queries — and vanishes only when dozens of queries fire inside the same scheduling instant. Capacity 4 is reasonable as shipped; the bench stays in-tree so an Android `protect(fd)` profiler can re-measure with a larger idle cap without new tooling. Each phase also asserts the pool stays bounded, so the bench doubles as a behavioral regression test.

### Validation

`cargo test -p meow-dns --lib`: 150 pass. The single failure (`udp_client_times_out_on_unroutable`) is environment-dependent (ICMP-unreachable yields `Io` instead of `Timeout` on some networks) and **pre-dates this PR** — it is exactly the test #476 makes hermetic. `cargo clippy` clean.

PR #463, opened 2026-08-20
fix(dns): prefer-IPv4 lookup, per-family cache, ipv6 policy, single-flight

### LINKED ISSUE #461: DNS: A+AAAA fanout wastes queries, min-TTL cache collapse evicts fresh answers, and top-level ipv6 doesn't drive every resolver
## Problem

Three related DNS-resolution issues, all present on current main.

### 1. Parallel A+AAAA fanout
`DnsClient::lookup_ip` does `tokio::join!(query A, query AAAA)` — both
families in parallel on every lookup. The resolver races two upstreams, so
a single host resolution is **4 TCP DNS queries** (2 upstreams × 2
families), all serialised on one connection per upstream. Under a
page-load burst this starves the proxy's own server-hostname resolution.

### 2. min-TTL cache collapse
`CacheEntry` has a single `expire_at = min(v4_ttl, v6_ttl)`. A 10 s AAAA
NODATA (clamped MIN TTL) collapses a 3600 s A answer to a 10 s entry — a
~360× re-query amplification for dual-stack domains. The short-TTL family
evicts the still-fresh long-TTL family.

### 3. Incomplete ipv6 propagation
The top-level `ipv6` setting drives the main DNS resolver, but not the
bootstrap / direct-adapter / proxy-provider resolvers. A `direct:` proxy
with a `dns:` override, or a proxy-provider health check, resolves AAAA
even when the user disabled IPv6 — wasting a round-trip and yielding v6
addresses the VPN has no route for.

## Proposal

- **Prefer-IPv4**: query A first; fall back to AAAA only when A yields no
  address; skip AAAA entirely when IPv6 is disabled. The common
  v4-capable case becomes a single query. v6-only domains still resolve
  via the AAAA fallback.
- **Per-family cache**: track `expire_v4` / `expire_v6` independently so a
  short-TTL family no longer evicts a still-fresh long-TTL one; a
  per-family merge preserves the other family's IPs and expiry.
- **ipv6 everywhere**: thread the top-level `ipv6` through every
  `Resolver::new` (bootstrap, direct adapter, proxy provider).
- **Single-flight keyed by host+family** to avoid thundering herds, and
  **first-definitive-negative** so a dead upstream's timeout doesn't stall
  a live upstream's NXDOMAIN.

## Scope note

Per-family / prefer-IPv4 / IPv6-policy slice split out of #387. TCP
exchange pooling is a separate, self-contained issue/PR (#460).

ADR-0011 note: `CacheEntry` grows by one `Instant` (per-family expiry);
`-Zprint-type-sizes` 56→72 B macOS / 48→56 B Linux, at/under the M2 72 B
per-slot cap, guarded by a size-regression test.

aimkiray:
Resolves #461.

## Problem
See #461: parallel A+AAAA fanout (4 queries/host under two racing
upstreams), min-TTL cache collapse (short-TTL AAAA evicts long-TTL A), and
the top-level `ipv6` not driving bootstrap/direct/provider resolvers.

## Approach
- **client.rs**: split `query` into `query` (timeout wrapper) +
  `query_inner` (no double timeout); `lookup_ip_with_ipv6` does A-first
  with AAAA fallback, skipping AAAA when IPv6 is off; `lookup_family` /
  `lookup_set` give the resolver a per-family query that preserves the
  NXDOMAIN/NODATA distinction; `IpLookupResult` / `FamilyAnswer` /
  `FamilySet` unify the pipeline shape.
- **cache.rs**: per-family `CacheEntry` (`expire_v4`/`expire_v6`) +
  `QueryFamilies` bitset; `get_lookup` returns a `CacheLookup` with v4/v6
  `FamilyCacheHit`s; `merge_family` preserves the other family's fresh IPs
  and its own expiry.
- **resolver.rs**: unified pipeline (policy → main → fallback) with
  symmetric first-wins negative bookkeeping; single-flight keyed by
  host+family (broadcast); first-definitive-negative; `resolve_ips` only
  re-queries the families the cache can't already answer; `ipv6` drives
  `required` and `filter_enabled_ips`. Reconciled with #410 host-alias
  chain and #420 `resolve_ips_local`.
- **server.rs**: A/AAAA answer path uses the per-family lookup results;
  AAAA under `ipv6:false` returns NODATA without an upstream round-trip;
  HTTPS/SVCB `ipv6hint` stripped when IPv6 is off.
- **config**: `ipv6` threaded through `dns_parser` / `proxy_parser` /
  `proxy_provider` / `rule_provider` / direct dial.

NXDOMAIN is not cached (re-query gets a fresh authoritative answer); NODATA
is cached per-family with its own TTL.

ADR-0011 footprint (CacheEntry touched): per-family expiry replaces one
`expire_at` with `expire_v4`+`expire_v6` (one extra `Instant`).
`-Zprint-type-sizes`: 56→72 B macOS (Instant=16 B), 48→56 B Linux
(Instant=8 B). macOS is at the M2 72 B per-`LruEntry`-slot cap (under on
Linux); guarded by `cache_entry_fits_m2_size_cap`. ReverseEntry unchanged.

## Verification
- three-way `cargo clippy --all-targets -- -D warnings`
- `cargo test -p meow-dns --lib` (157), `--test rules_test` (102),
  `--test trojan_integration` (5)
- meow-tunnel: fakeip/lazy_resolve/pre_resolve/raii_guard/relay_churn/
  reload_churn; meow-config `config_test` (67); meow-api (lib+api_test+
  api_logs_ws); `direct_dns_inherits_disabled_ipv6_policy`

Diff: 24 files, +1894/-348. TCP pooling is intentionally not included
(separate PR #462); the two touch disjoint regions of `client.rs`.


madeye:
Reviewed at design level and verified: merges clean with `main`, 157 meow-dns tests pass including `cache_entry_fits_m2_size_cap`.

Flagging up front that this isn't a follow-up nit like the rest of the current queue — 1,894 lines across 24 files, rewriting the resolver pipeline, cache layout, server answer path and config threading. It warrants its own focused review pass, and it's the strongest argument for soak time before the next release. Two specific points:

**1. `CacheEntry` budget — which reading is the M2 exit number?** This takes it 56 → 72 B on macOS, landing exactly on the cap with zero headroom. The guard test asserts `size_of::<CacheEntry>() <= 72`, but `CLAUDE.md` words the budget as "72 B per **LruEntry slot**". If that means the whole slot (key + entry + LRU links), this is over budget rather than on it. ADR-0011 itself defers the number ("No upfront target — write the goal in M2.dns-cache after baseline data"), so the M2-exit measurement is the authority — worth confirming which of the two it measured, since either reading leaves macOS with no room for the next field.

**2. NXDOMAIN is deliberately not cached.** The stated rationale (a re-query always gets a fresh authoritative answer) is reasonable in isolation, but it removes negative-cache damping: DGA/malware traffic, or any client retry-looping on a bogus name, re-queries upstream every time. RFC 2308 negative caching exists for that load. Worth confirming this matches mihomo before shipping, and if it's a deliberate divergence, recording it as one under ADR-0002.

The A-first-with-AAAA-fallback change is consistent with this project's stated dual-stack/v4-only scope, so no concern there.

aimkiray:
Thanks for the two design points — I traced the history on each.

**1. CacheEntry budget.** The 72 B cap is on `size_of::<CacheEntry>()` (the `.val` of each `LruEntry`), not the full slot. On macOS the full per-slot cost is ~104 B (key `Arc<str>` 16 + value 72 + LRU links 16); on Linux it is ~88 B (Instant = 8 B). The test and code comment are consistent on this — the wording in `CLAUDE.md` ("72 B per LruEntry slot") is imprecise and I will tighten it to "72 B per `CacheEntry` (the LRU `.val`)" to remove the ambiguity.

**2. NXDOMAIN not cached.** I dug into this and I think the current behavior is actually a mistake, not a deliberate divergence. Here is the trail:

- **PR #387 initial version** had a bug where the A-first path returned NXDOMAIN for AAAA queries on dual-stack domains (your blocking review point #1). The concern was that clients would *negative-cache* that NXDOMAIN against the bare name, poisoning subsequent A lookups.
- **Fix commit `35f30ae`** ("address resolver review feedback") introduced per-family caching to fix the poisoning, and its commit message says: *"Track queried address families so A-first resolution does not poison AAAA lookups, and preserve NODATA/NXDOMAIN/SERVFAIL semantics."*
- **PR #463** inherited the same code unchanged from #387.

The poisoning concern is about what the DNS *server* returns to clients (client-side negative caching) — which `server.rs:179` already handles by returning SERVFAIL instead of NXDOMAIN for generic-forward failures. But *our internal cache* not caching NXDOMAIN is a separate layer: it only affects whether *we* re-query upstream for repeat NXDOMAIN, not what we send to clients. The two were conflated in the defensive reaction to the #387 review, and the code comment ("a re-query always gets a fresh authoritative answer") is post-hoc — that logic would apply equally to NODATA, which we *do* cache.

I verified against mihomo source: `putMsgToCache` (`dns/util.go:60`) caches NXDOMAIN using `minimalTTL(Answer + Ns + Extra)` — i.e. the SOA TTL from the Authority section (RFC 2308 negative caching). So this is a real parity gap, not an intentional divergence.

**Proposed fix (plan C):** Cache NXDOMAIN as a per-family negative, same as NODATA:
- `client.rs`: extract the SOA TTL from `message.authorities` on the `ResponseCode::NXDomain` branch, carry it through `FamilyLookupResult::NxDomain(Duration)` → `FamilyAnswer::NxDomain(Duration)`.
- `resolver.rs` `merge_set_into_cache`: cache `NxDomain(ttl)` via `merge_family(host, family, &[], ttl, …)` — same path as NODATA, no `cache.rs` changes.
- `resolver.rs` `lookup_real_with_ttl`: a cached NXDOMAIN reads back as `FamilyCacheHit::NoData` (the cache entry has `queried = BOTH`, empty IPs), so the server path returns `NoData` → `build_nxdomain` on first upstream miss, and `build_noerror_empty` on subsequent cache hits. Most clients treat both identically; if exact rcode preservation on cache hits matters, that is a follow-up.
- Delete the misleading comment at `resolver.rs:1091`.

~30 lines changed + ~20 lines tests; `cache.rs` and `server.rs` untouched.

Before I implement: do you recall any other concern from the #387 review that motivated not caching NXDOMAIN, beyond the poisoning issue? If not I will go ahead with plan C.

aimkiray:
## Deep Review Fixes

Based on a thorough code review of PR #463, the following issues were fixed:

### 🔴 High

**H1: BOTH path NXDOMAIN downgraded to NODATA**
- Added `classify_family_message()` in `client.rs` to uniformly classify upstream RCODE; `lookup_ip_with_ipv6_inner()` now retains independent `FamilyAnswer` per A/AAAA family.
- When A returns NXDOMAIN, it is propagated to v6 (DNS-correct: NXDOMAIN applies to the name, not the RR type), and AAAA is not queried.
- `family_set_from_ip_lookup()` uses per-family answers directly instead of deriving from the aggregate IP list.

**H2: Cross-granularity cache overwrite**
- Added `preserve_existing` guard in `cache.rs` `merge_family()`: a fresh NXDOMAIN is not downgraded or shortened by a later NODATA merge; a fresh positive answer is not overwritten by an empty result.
- A later NXDOMAIN can still overwrite a stale NoData (rcode upgrade); a later positive Answer can still replace a negative (DNS change recovery).

### 🟠 Medium

**M1: NODATA not using RFC 2308 SOA TTL**
- Both `lookup_family()` and `classify_family_message()` now use `negative_ttl()` (`min(SOA.TTL, SOA.MINIMUM)`) for NoError + empty answer, consistent with the NXDOMAIN path.

### Tests added (3)
- `classify_family_message_branches`: all four branches of the BOTH-path classifier (Answer / NoData / NxDomain / Failed).
- `fresh_positive_is_not_overwritten_by_nodata_merge`: positive answer survives a concurrent NoData merge.
- `nxdomain_is_cached_and_preserves_rcode_on_second_lookup` extended: added v6 NXDOMAIN propagation assertion.

### Cleanup
- Removed `http.pid`, `rules/ShellClash-*.list`, and unrelated merge changes (branch rebuilt from `origin/main`).
- Consolidated redundant `merged_queried` assignment in `merge_family`.
- Fixed `negative_ttl` doc comment referencing the wrong field name.

aimkiray:
## Deep Review

Re-reviewed every change in this PR item by item, confirming all 4×P1, 2×P2, and the low-risk items from the report are correctly implemented and covered by regression tests.

### P1 blockers

**1. A-first caused DirectAdapter to lose IPv6 fallback** ✅
`lookup_ip_with_ipv6_inner()` (client.rs:477) now races A and AAAA concurrently via `tokio::join!` when IPv6 is enabled, collecting both families with IPv4 ordered first (prefer-IPv4) — no longer A-first serial short-return. `resolve_ips()` (resolver.rs:943) requires `BOTH` and returns both families; `DirectAdapter::dial_tcp()` (direct.rs:282) iterates the full address list. The old "A has DNS addresses but TCP fails → no AAAA candidate → fail" path is gone; IPv6 remains a connection fallback when IPv4 is unreachable. Test: `resolve_ips_returns_both_families_for_dual_stack`.

**2. `merge_family` revived an expired sibling family** ✅
cache.rs:629 now checks `other_fresh` (each family's expiry vs `now`) and only `union(other)` + copies its IPs/expiry when the sibling is still fresh; an expired sibling is not marked queried and stays `Miss`, so it is re-queried on demand. Test: `merge_family_does_not_revive_expired_sibling` covers the full "dual-stack → AAAA expires → new A merges → AAAA stays Miss" chain.

**3. Single-family cache blocked `resolve_ips()` from querying the other family** ✅
resolver.rs:957 sets `want = required` and only drops families that are already fresh; any required family still `Miss` keeps `want` non-empty and triggers `run_pipeline`. It no longer short-circuits on "any available IP"; it only returns `None` when all required families are fresh. Tests: `resolve_ips_queries_missing_family_after_a_only_cache`, `dual_stack_cache_queries_missing_aaaa_family`.

**4. API reported the wrong IPv6 default** ✅
routes.rs:632 now uses `raw.ipv6.unwrap_or(true)`, matching config lib.rs:1652 `unwrap_or(true)`; configuration.md:24 default is `true`. Runtime, API, and docs are aligned.

### P2

**5. Fast negative suppressed a slow positive in one tier** ✅
`query_pool_set()` (resolver.rs:416) introduces `NEGATIVE_GRACE_PERIOD = 300ms`: the first definitive negative is recorded but not returned immediately; remaining futures keep racing within the grace window; `has_positive()` always returns immediately; the negative is only committed after the grace period; `Err` (network failure) is never definitive. Tests: `fast_negative_does_not_suppress_slow_positive`, `both_query_keeps_racing_when_one_family_is_unknown`.

**6. Single-flight broadcast miss race** ✅
`lookup_real_with_ttl()` (resolver.rs:1167) is now `let Some(set) = result else { re-read cache }`: the publisher merges the result into the cache before broadcasting, so a subscriber that missed the broadcast re-reads the cache instead of judging `Failed`, avoiding occasional SERVFAIL.

### Low-risk items

**7. DoH response body unbounded** ✅
`doh_exchange` used unbounded `read_to_end`; now uses the extracted `read_bounded_http_response()` (`#[cfg(feature = "encrypted")]`), reading in 4 KiB chunks and returning `Protocol` error once `MAX_DOH_RESPONSE_BYTES = 65535 + 16K` (DNS message max + HTTP header headroom) is exceeded. Test: `doh_bounded_read_enforces_response_size_cap` (at-cap accepted, +1 byte rejected).

**8. `snapshot()` showed expired-family IPs** ✅
cache.rs:739 only shows IPs of fresh families; test `snapshot_hides_expired_family_ips`.

**9. `ipv6:false` ignored hosts AAAA overrides** ✅
server.rs:127 returns the hosts-table IPv6 address for AAAA queries; hosts is an explicit override that takes priority over the global `ipv6` toggle.

**10. `ipv6:false→true` silent change needs a release note** ✅
New root `CHANGELOG.md` (Keep a Changelog); `[Unreleased]` lists the `ipv6` default change as the first `Changed` entry and records every behaviour change/fix in this PR.

**11. v6-only serial A→AAAA added an RTT** ✅
Now concurrent via `tokio::join!`; no extra RTT.

### Fix-up

The earlier single-flight fix wrote `let set = match result { Some(s) => s, None => { … return … } }`, which tripped `clippy::manual_let_else` and broke `cargo clippy -- -D warnings`. Rewritten as `let Some(set) = result else { … };` — behaviour unchanged. Pre-existing `cargo fmt` violations in cache.rs / client.rs / resolver.rs were also cleaned up.

### Gates

```
cargo fmt --all -- --check                                              ✅ clean
cargo clippy -p meow-dns (default / no-default / all-features) -D warnings   ✅ all green
cargo clippy -p meow-api / -p meow-config --all-targets -D warnings       ✅ all green
cargo test -p meow-dns --features encrypted --lib                        ✅ 168 passed
cargo test -p meow-dns --features encrypted --test dns_cache_test        ✅ 18 passed
regression bar (common/dns_cache/config/rules/api/trojan/crate_invariants) ✅ 0 failed
```

All items from the report are implemented, test-covered, and the gates are green.

aimkiray:

Verified every finding in the review against this PR's head code and pushed two fix-ups (`85bb0ee`, `d522f4b`). All fast-forward pushes; no history rewritten. Note the branch head already contained several review fixes (5ce6d54) — everything below was re-verified against the actual head, not the original PR diff.

### Round 1 (`85bb0ee`)

- **B3 (confirmed, fixed):** this branch still carried the silent-loopback variant of the UDP timeout test that collides with #476. Restored to main's hunk, with a note attributing the hermetic version to #476.
- **H3 (confirmed, fixed):** `resolve_ips` discarded `run_pipeline`'s result. It now takes the returned `FamilySet` as the primary result — a definitive full-family set returns directly; a strict-subset query falls back to a cache re-read; and if that re-read was LRU-evicted, it still returns the pipeline's positives instead of failing. All three branches covered by the reasoning and tests.
- **H5 (confirmed, fixed as docs-only):** the CHANGELOG's hosts/`ipv6` wording claimed behavior the code doesn't have. Rewritten to describe the actual (gated) behavior. Per review decision, no code change for the hosts part.
- **M1 (confirmed, fixed):** tuple keys allocated on every single-flight probe. Replaced with three per-family `DashMap<Arc<str>, InflightTx>` slots; `Arc<str>: Borrow<str>` keeps the hot path allocation-free (`&str` lookups, `Arc` only on insert), and the guard removes by borrowed key.
- **M2 (confirmed, fixed):** a partial-negative placeholder no longer starves the 300 ms grace path — any first negative starts the negative deadline, and a later definitive outcome can still upgrade a kept `Partial`.
- **M3 (confirmed, documented):** v6-only bootstrap ignoring `ipv6: false` is deliberate (an encrypted upstream whose own hostname must resolve over IPv6 is the only path in). Documented as a divergence in code comments.
- **M4 (confirmed, documented):** `put_with_source(empty)` → NXDOMAIN is a tested contract; the doc now carries an explicit "caller beware" note pointing negative-cache writers at `merge_family`.
- **M5 (confirmed, fixed):** the size-cap guard only pinned `<= 72`, leaving Linux's 16 B headroom forever unused. Now asserts exact per-platform layouts (`== 56` on Linux, `== 72` on macOS) plus a universal `<= 72`, so any growth fails on *every* platform.
- **M6 (confirmed, fixed):** `merge_family` used `sort_unstable()` before dedup, discarding upstream ordering on every merge. Replaced with an order-preserving in-place dedup; the read side still yields v4-first (prefer-IPv4 contract intact).
- **ipv6 default:** reverted to `false` (mihomo/Clash parity) per review decision — an omitted `ipv6:` now resolves IPv4-only end to end. All eight scattered `unwrap_or(true/false)` sites collapsed into one `meow_config::effective_ipv6` helper; CHANGELOG, configuration.md, and config/API tests updated (including a new explicit-`ipv6: false` case). **User-visible change** — operators relying on the old always-dual-stack default must set `ipv6: true`; spelled out in the CHANGELOG.

### Round 2 (`d522f4b`)

- **N2 (confirmed, hardened):** `run_pipeline` relied on a `debug_assert` + `inflight_slot`'s `unreachable!()`. The early return now makes an unguarded empty-`want` call safe in release (`None` — same answer as a cache miss) while the `debug_assert` keeps the invariant loud in dev; `inflight_slot`'s `unreachable!` becomes a defensive backstop, documented as such.
- **N3 (confirmed, fixed):** `merge_family`'s release-build refusal of a non-single family was a fully silent no-op ("the cache never stores this host" with nothing to go on). It now emits a `tracing::warn!` before returning.

### Validation

- `cargo check --workspace --all-targets` green; `cargo test -p meow-dns --lib` 167 pass; `meow-config` / `meow-api` suites green; `cargo fmt` + `cargo clippy` clean.
- The only failure in local runs is `udp_client_times_out_on_unroutable`, environment-dependent (ICMP unreachable → `Io` instead of `Timeout`) and **pre-dating this PR** (verified by stashing all changes) — the hermetic version lands in #476.
- Known-untested note carried into the PR description scope: the DoH bounded-read hardening is review-requested scope creep kept unsplit to avoid rewriting published history.

PR #464, opened 2026-08-20
fix(transport): exempt #[cfg(test)] items from the ADR-0001 src scan

madeye:
## Problem

`main` is red at 5439f07. `crate_invariants_test::no_server_side_symbols_in_src` walks `meow-transport/src/**/*.rs` line by line with no notion of conditional compilation, so the in-src h2 mock server added with the write-cancellation contract tests (#428, #441) trips the `\baccept\b` pattern:

```
h2_common.rs:329: 'let _accepted = connection.accept().await;'
h2_common.rs:421: '.accept()'
h2_common.rs:424: '.expect("accept ok");'
h2_common.rs:446: 'while connection.accept().await.is_some() {}'
```

All four are inside `#[cfg(test)] mod tests`. The nightly `Coverage` job is red for the same reason.

## Why exempting is right, not a loosening

That code is compiled out of the published library, and ADR-0001 §1 (implementation plan, M1.A-3) explicitly prescribes exercising these client layers against *"a loopback `h2` server in-process"*. The guard was flagging the test strategy the ADR asks for. `tests/` is already exempt for exactly this reason — in-src test modules were simply never taught to the scanner.

## Change

- `test_only_lines()` returns the line span of every `#[cfg(test)]` item; both `src/` scans (F2 server symbols, F4 anyhow-at-boundary) skip those lines.
- Spans are tracked by brace depth over lines with string, raw-string and char literals stripped, so a `'{'` or an `r#"…}"#` cannot skew the count. Lifetimes (`'a`) and loop labels are not mistaken for char literals.
- Guard rails against the scanner silently switching itself off:
  - a block-less item (`#[cfg(test)] use …;`) terminates at its semicolon;
  - an unterminated span **panics** rather than swallowing the rest of the file;
  - `#[cfg(feature = "test-utils")]` does not match.
- Five self-tests pin that behaviour.

## Verification

The guard still bites — injecting `fn mutation_probe() -> Option<tokio::net::TcpListener>` into production code in `h2_common.rs` fails `no_server_side_symbols_in_src` exactly as before the change.

Full regression bar green locally: `fmt`, clippy ×3 (default / `--no-default-features` / `--all-features`), `RUSTDOCFLAGS="-D warnings" cargo doc`, and the full default-features test list (30 test binaries).

## Process fix

`CLAUDE.md`'s regression bar said `cargo test --lib`, which by construction runs **no** `--test` target and so could not have caught this before the push. It now mirrors the target list from the "Unit + integration tests (default features)" CI step, with a note to keep the two in sync.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -133,8 +133,22 @@ cargo clippy --all-targets -- -D warnings
 cargo clippy --all-targets --no-default-features -- -D warnings
 cargo clippy --all-targets --all-features -- -D warnings
 RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps
-cargo test --lib
+
+# Mirrors the "Unit + integration tests (default features)" CI step. `--lib`
+# alone is not enough: it skips every `--test` target, so a broken integration
+# test (e.g. the ADR-0001 guard in `crate_invariants_test`) passes locally and
+# lands `main` red.
+cargo test --lib --bin meow \
+  --test common_test --test dns_cache_test --test config_test \
+  --test statistics_test --test rules_test --test api_test \
+  --test config_persistence_test --test systemd_config_test \
+  --test trojan_integration --test vless_config_test --test vless_integration \
+  --test v2ray_plugin_integration --test pre_resolve_test \
+  --test tls_test --test ws_test --test crate_invariants_test
 ```
+
+Keep the target list in sync with `.github/workflows/test.yml`; a new `tests/`
+file that CI runs but this list omits is invisible to the local bar.
 
 The three-way clippy check (default / no-default-features / all-features) is enforced in CI via `.github/workflows/test.yml` (added in M1, per ADR-0010 §3).
 
--- a/crates/meow-transport/tests/crate_invariants_test.rs
+++ b/crates/meow-transport/tests/crate_invariants_test.rs
@@ -5,6 +5,7 @@
 //! is violated.
 
 use meow_transport::TransportError;
+use std::collections::HashSet;
 
 // ─── F1: no_proxy_dep ────────────────────────────────────────────────────────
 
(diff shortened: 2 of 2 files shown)


aimkiray (commented):
*(Duplicate of the issue comment below — please disregard.)*

aimkiray:
Nice approach — extending the `tests/` exemption to `#[cfg(test)]` items is semantically the right call, and the CLAUDE.md local-test-command update is a great addition that would have prevented this from landing on a red main.

One gap I noticed while reviewing `strip_literals_and_comments`:

**Block comments `/* ... */` are not handled.** The function strips `//` line comments, string literals, raw strings, and char literals, but not block comments. This means braces inside block comments are counted towards depth tracking:

```rust
#[cfg(test)]
mod tests {
    /* } */  // ← this } is counted, depth drops to 0 prematurely
    let _ = connection.accept().await;  // ← not in test_only span → false positive
}
```

Conversely, `/* { */` inflates depth and could extend the span past the real closing brace (false negative — code after the test module would be skipped).

This is an edge case, but since the whole point of the guardrail is to be a hard enforcement mechanism, a false positive on legitimate test code would be disruptive (CI red on a valid PR) and a false negative would defeat the guardrail's purpose.

Suggest adding block-comment stripping to `strip_literals_and_comments` (both single-line `/* } */` and multi-line `/* ... \n ... */` spanning lines), with a test case like:

```rust
#[test]
fn cfg_test_span_survives_braces_in_block_comments() {
    let src = "\
#[cfg(test)]
mod tests {
    /* } brace in block comment */
    fn f() {
        let _ = connection.accept().await;
    }
}
";
    let span = test_only_lines(std::path::Path::new("<memory>"), src);
    assert_eq!(span, (0..7).collect::<HashSet<usize>>());
}
```

PR #465, opened 2026-08-20
feat(dns): support "rule-set:" patterns in nameserver-policy

### LINKED ISSUE #299: feature request: add geosite and ruleset support in nameserver policy

aimkiray:
Closes #299

## What

Implements `rule-set:` pattern support in `dns.nameserver-policy`. This allows routing DNS queries for domains contained in a rule-provider (rule-set) to dedicated nameservers, matching upstream mihomo behavior.

## Why

Issue #299 requested both `geosite:` and `rule-set:` support in `nameserver-policy`. `geosite:` was already implemented; `rule-set:` was still skipped with a warn-once. This change completes the feature.

## Usage example

```yaml
dns:
  nameserver-policy:
    "rule-set:ShellClash-ChinaDomain": [223.5.5.5, 119.29.29.29]
    "rule-set:ShellClash-LocalAreaNetwork,ShellClash-GoogleCN": [127.0.0.1:53]
```

## How

- Added `RuleSet::matches_domain(&str) -> bool` in `crates/meow-rules`:
  - `DomainRuleSet`: direct trie search (hot path, no `Metadata` allocation).
  - `ClassicalRuleSet`: evaluates only domain rules against a host-only `Metadata` so IP/port/process rules cannot accidentally match.
  - `IpCidrRuleSet`: default `false`.
- Updated `crates/meow-config/src/dns_parser.rs::build_nameserver_policy` to:
  - Accept loaded rule-providers.
  - Compile `rule-set:` keys into `NameserverPolicyMatcher` closures that snapshot the provider on each lookup.
  - Validate provider behavior: `Domain` allowed, `Classical` allowed with a per-entry warning, `IpCidr` hard error, missing provider hard error.
  - Preserve original-case provider names (case-sensitive, like upstream).
- Reordered startup in `crates/meow-config/src/lib.rs` so rule-providers are loaded before DNS configuration is parsed.
- Updated tests, specs (`docs/specs/dns-nameserver-policy.md`), migration guide, and website guide.

## Breaking change

Configs that previously contained `rule-set:` keys in `nameserver-policy` were silently skipped with a warning. They now:

- hard-error if the referenced provider does not exist;
- hard-error if the provider behavior is `IpCidr`;
- warn per entry and match only domain rules if behavior is `Classical`.

This aligns with upstream mihomo and prevents silent DNS leakage.

## Changes

```
crates/meow-config/src/dns_parser.rs          | 202 ++++++++++++++++++++++++--
crates/meow-config/src/lib.rs                 |  37 +++--
crates/meow-rules/src/rule_set.rs             |  65 +++++++++
docs/migration-from-go-mihomo.md              |   3 +-
docs/specs/dns-nameserver-policy-test-plan.md |   6 +
docs/specs/dns-nameserver-policy.md           |   6 +-
website/guide/dns.md                          |  11 +-
```

## Testing

- `cargo fmt --all -- --check` ✅
- `cargo clippy --all-targets -- -D warnings` (default, no-default-features, all-features) ✅
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` ✅
- `cargo test -p meow-rules -p meow-config --lib` ✅ 356 passed, 0 failed
- `cargo test --test rules_test` ✅ 102 passed
- `cargo test --test trojan_integration` ✅ 5 passed
- End-to-end smoke test with an [endfield](https://github.com/aimkiray/endfield) subscription (commit `2a25d57`) and a local sing-box VMess node:
  - `taobao.com` (ShellClash-ChinaDomain) → policy nameserver ✅
  - `alt1-mtalk.google.com` (ShellClash-GoogleCN) → policy nameserver ✅
  - `foo.internal` (ShellClash-LocalAreaNetwork) → policy nameserver ✅
  - `example.com` / `github.com` → global nameserver ✅
  - `curl --proxy socks5h://127.0.0.1:7893 https://www.gstatic.com/generate_204` → `204` via LocalVMess → sing-box ✅

## Notes for reviewers

- `geosite:` in `nameserver-policy` already worked on `main` and is not changed by this PR.
- The CI `test` job failure (`no_server_side_symbols_in_src`) is a **pre-existing breakage on `main`** (PR #441 introduced `h2::server::Connection::accept()` in `src/h2_common.rs` inline tests). PR #464 is fixing it. This PR does not touch any `meow-transport` files.
- The `meow-dns::client::tests::udp_client_times_out_on_unroutable` test is a pre-existing environment-dependent flaky test unrelated to this change.

### DIFF
--- a/crates/meow-config/src/dns_parser.rs
+++ b/crates/meow-config/src/dns_parser.rs
@@ -1,4 +1,5 @@
 use crate::raw::{HostsValue, RawConfig};
+use crate::rule_provider::RuleProvider;
 use crate::DnsConfig;
 use meow_common::DnsMode;
 use meow_dns::fakeip::{FileStore, MemoryStore, Pool, Skipper, SkipperMode, Store};
@@ -7,6 +8,7 @@ use meow_dns::resolver::{
 };
 use meow_dns::upstream::{NameServerEntry, NameServerUrl};
 use meow_dns::{DnsClient, HostOrIp, Resolver};
+use meow_rules::RuleSetBehavior;
 use meow_trie::DomainTrie;
 use std::collections::HashMap;
 use std::net::{IpAddr, SocketAddr};
@@ -24,6 +26,7 @@ pub async fn parse_dns(
     cache_dir: Option<&std::path::Path>,
     proxy_registry: &HashMap<smol_str::SmolStr, Arc<dyn meow_common::Proxy>>,
     geosite: Option<Arc<meow_rules::geosite::GeositeDB>>,
+    rule_providers: &HashMap<String, Arc<RuleProvider>>,
 ) -> Result<DnsConfig, anyhow::Error> {
     let dns = match &raw.dns {
         Some(dns) if dns.enable.unwrap_or(false) => dns,
@@ -121,6 +124,7 @@ pub async fn parse_dns(
                     geosite.as_ref(),
                     &bootstrap_clients,
                     proxy_registry,
+                    rule_providers,
                 )
                 .await?,
             )
@@ -309,7 +313,10 @@ fn parse_nameserver_urls(servers: &[String]) -> Result<Vec<NameServerUrl>, anyho
 /// Build a `NameserverPolicy` from the raw YAML map.
 ///
 /// `geosite:` patterns are compiled into matchers when a geosite DB is loaded.
-/// Unsupported prefixes (currently `rule-set:`) warn once and skip.
+/// `rule-set:` patterns are compiled into matchers from the loaded
+/// rule-providers; the matcher snapshots the provider on each lookup so
+/// background refreshes take effect automatically.
+/// Other prefixed patterns (anything with `:`) warn once and skip.
 ///
 /// An entry with no valid nameservers after skipping → hard error.
 /// Class A per ADR-0002: DNS leakage risk for internal/corporate domains.
@@ -318,6 +325,7 @@ async fn build_nameserver_policy(
     geosite: Option<&Arc<meow_rules::geosite::GeositeDB>>,
     bootstrap_clients: &[Arc<DnsClient>],
     proxy_registry: &HashMap<smol_str::SmolStr, Arc<dyn meow_common::Proxy>>,
+    rule_providers: &HashMap<String, Arc<RuleProvider>>,
 ) -> Result<NameserverPolicy, anyhow::Error> {
     let mut policy = NameserverPolicy::new();
     let mut warned_unsupported_prefix = false;
@@ -350,10 +358,44 @@ async fn build_nameserver_policy(
                 continue;
             }
 
-            if key_lower.starts_with("rule-set:") || key_lower.contains(':') {
+            if key_lower.starts_with("rule-set:") {
+                // Provider names are case-sensitive — extract from the
+                // original-case `expanded_key`, not `key_lower`.
+                let name = expanded_key["rule-set:".len()..].trim();
+                if name.is_empty() {
+                    continue;
+                }
+                let provider = rule_providers.get(name).ok_or_else(|| {
+                    anyhow::anyhow!("nameserver-policy: not found rule-set: {name}")
+                })?;
+                match provider.behavior {
+                    RuleSetBehavior::IpCidr => {
+                        anyhow::bail!(
+                            "nameserver-policy: rule-set '{name}' behavior is IpCidr, \
+                             expected domain or classical"
+                        );
+                    }
+                    RuleSetBehavior::Classical => {
+                        // Warn once per entry; upstream only matches domain
+                        // rules inside a classical set.
+                        warn!(
+                            "nameserver-policy: rule-set '{name}' is classical; \
+                             only domain rules within it will be matched"
+                        );
+                    }
+                    RuleSetBehavior::Domain => {}
+                }
+                let provider = Arc::clone(provider);
+                patterns.push(PolicyPattern::Matcher(Arc::new(move |domain: &str| {
+                    provider.snapshot().matches_domain(domain)
+                })));
+                continue;
+            }
+
+            if key_lower.contains(':') {
                 if !warned_unsupported_prefix {
                     warn!(
-                        "nameserver-policy: unsupported prefixed patterns such as 'rule-set:' \
+                        "nameserver-policy: unsupported prefixed patterns containing ':' \
                         will be skipped"
                     );
                     warned_unsupported_prefix = true;
@@ -863,7 +905,7 @@ mod tests {
             "dns:\n  enable: true\n  nameserver:\n    - 1.1.1.1\n  proxy-server-nameserver:\n    - 223.5.5.5\n",
         )
         .unwrap();
-        let cfg = parse_dns(&raw, None, None, &HashMap::new(), None)
+        let cfg = parse_dns(&raw, None, None, &HashMap::new(), None, &HashMap::new())
             .await
             .unwrap();
         assert!(
@@ -876,7 +918,7 @@ mod tests {
     async fn no_proxy_server_nameserver_leaves_proxy_resolver_none() {
         let raw: RawConfig =
             serde_yaml::from_str("dns:\n  enable: true\n  nameserver:\n    - 1.1.1.1\n").unwrap();
-        let cfg = parse_dns(&raw, None, None, &HashMap::new(), None)
+        let cfg = parse_dns(&raw, None, None, &HashMap::new(), None, &HashMap::new())
             .await
             .unwrap();
         assert!(cfg.proxy_resolver.is_none());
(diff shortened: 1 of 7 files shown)


aimkiray:
## Deep Review — PR #465: `rule-set:` patterns in `nameserver-policy`

### Summary

The feature design is solid: `rule-set:` matching works, comma expansion works, live refresh via `PUT /providers/rules/:name` works (verified end-to-end with an HTTP provider that swaps its payload mid-run). The startup ordering change (rule-providers load before DNS) is safe — prefetched payloads are reused, HTTP fetch count is unchanged (1 per provider), and the step-1 proxy registry is functionally equivalent for provider downloads.

However, I found **three correctness issues** during testing — one is a real DNS leak, two are robustness/determinism problems. Fixes are pushed to this branch as commit `c2e90ea`.

---

### Issue 1 — Classical rule-set DNS leak (CRITICAL)

**Problem:** `ClassicalRuleSet::matches_domain` evaluates *all* parsed rules against a default `Metadata` (network = Tcp, conn_type = Http, ports = 0). Rules that match these defaults — `NETWORK,tcp`, `IN-TYPE,HTTP`, `DST-PORT,0`, `SRC-PORT,0` — return `true` for *every* domain, hijacking the entire DNS resolution to the policy nameserver.

**Reproduction (before fix):**
```yaml
rule-providers:
  cls:
    type: inline
    behavior: classical
    payload: [DOMAIN-SUFFIX,probe.example, NETWORK,tcp]
```
```
dig @127.0.0.1 -p 18157 totally-unrelated.example → 8.8.4.4 (policy)
                                                     ^ should be 9.9.9.9 (global)
```

**Fix:** Replace the "host-only Metadata probe" with a **rule-type whitelist**. Only domain-family rules (`Domain | DomainSuffix | DomainKeyword | DomainRegex | DomainWildcard | GeoSite`) are evaluated. Everything else — `NETWORK`, `IN-TYPE`, `DST-PORT`, `MATCH`, logic rules — is excluded. This matches upstream mihomo's documented "only domain rules within classical sets" semantics.

**Tests added:** `classical_rule_set_matches_domain_ignores_default_matching_rules` (5 rule types), `classical_rule_set_matches_domain_matches_geosite`, `parse_nameserver_policy_rule_set_classical_no_default_match_leak` (end-to-end at dns_parser level).

---

### Issue 2 — Startup availability regression (HIGH)

**Problem:** `load_providers_prefetched` warns and skips providers that fail to load (e.g. HTTP timeout, no cache). The `build_nameserver_policy` function then calls `.ok_or_else(|| ... "not found rule-set")` — treating a load failure identically to a typo. Result: a transient network failure during startup (CDN blip, boot-before-network) crashes the entire daemon, whereas the `rules:` path just drops the rule and continues.

**Reproduction (before fix):**
```
rule-providers:
  cn:
    type: http
    url: "http://127.0.0.1:9/unreachable.yaml"
    ...
nameserver-policy:
  "rule-set:cn": [223.5.5.5]
```
```
ERROR meow-rs stopped with an error error=nameserver-policy: not found rule-set: cn
```

**Fix:** Distinguish two cases:
- Name **not declared** in `rule-providers:` → hard error `"unknown rule-set 'NAME'"` (genuine typo, ADR-0002 Class A)
- Name **declared but absent** from loaded providers (its load already warned) → `warn!` and skip that matcher. Queries for its domains fall back to global nameservers — graceful degradation, consistent with the `rules:` path.

**Tests added:** `parse_nameserver_policy_rule_set_declared_but_unloaded_skips`, updated D14 assertion to `"unknown rule-set"`.

---

### Issue 3 — Non-deterministic matcher ordering (MEDIUM)

**Problem:** `build_nameserver_policy` iterates a `HashMap<String, RawNspValue>`. When two `rule-set:` (or `geosite:`) entries both match a domain, the winning nameserver depends on HashMap iteration order, which varies across runs. Verified with 6 consecutive starts of an overlapping config:
```
3.3.3.3 / 2.2.2.2 / 2.2.2.2 / 3.3.3.3 / 3.3.3.3 / 2.2.2.2
```

**Fix:** Sort keys before iteration so matcher insertion order is deterministic. With the fix, 5 consecutive runs all return the same winner (the alphabetically-first key wins):
```
8.8.4.4 / 8.8.4.4 / 8.8.4.4 / 8.8.4.4 / 8.8.4.4
```

A full `IndexMap` (preserve YAML order) would be ideal but is a larger change to `RawConfig`; sorted-key is the minimal fix. Documented in the spec.

**Tests added:** `parse_nameserver_policy_rule_set_overlapping_is_deterministic`.

---

### Verification

| Check | Result |
|-------|--------|
| `cargo fmt --all -- --check` | ✅ |
| `cargo clippy --all-targets` (default / no-default / all-features) | ✅ ✅ ✅ |
| `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` | ✅ |
| Full regression test suite (lib + 16 integration targets, ~500 tests) | ✅ |
| Runtime smoke: `NETWORK,tcp` no longer hijacks unrelated domains | ✅ |
| Runtime smoke: declared-but-failed provider → startup ok | ✅ |
| Runtime smoke: undeclared provider → hard error | ✅ |
| Runtime smoke: overlapping entries → 5/5 deterministic | ✅ |

### Files changed (fix commit)

```
crates/meow-config/src/dns_parser.rs          | +279 -38
crates/meow-rules/src/rule_set.rs             |  +65 -10
docs/migration-from-go-mihomo.md              |   +1  -1
docs/specs/dns-nameserver-policy-test-plan.md |   +5  -2
docs/specs/dns-nameserver-policy.md           |   +4  -2
website/guide/dns.md                          |   +9  -4
```

aimkiray:

Verified every finding in the review against this PR's head code and pushed three fix-ups (`7035a8e`, `a0b1f84`, `2db0280`). All fast-forward pushes; no history rewritten. The branch head already contained the reentrancy whitelist + hijack tests (c2e90ea); everything below was re-verified against the actual head.

### Round 1 (`7035a8e`)

- **H4 (confirmed, fixed):** `matches_domain` type-tested every rule on every DNS query. The domain-family subset is now pre-filtered once at construction (`domain_rule_indices`, single construction point verified by grep), the DNS hot path scans only those indices with an empty-subset fast path that skips even the `Metadata` construction, and the O(n)-per-query cost of large classical sets is documented ("prefer `behavior: domain`"). Two new tests pin the semantics: an `IP-CIDR` without `no-resolve` stays resolver-requiring on the full path but never fires on domain dispatch; an IP-only classical set never matches. `RuleMatchHelper` re-entry was verified impossible (bare marker, no resolver).

### Round 2 (`a0b1f84`)

- **N4 (confirmed, fixed — and it went deeper than reported):** logic rules (`AND`/`OR`/`NOT`) were excluded from the DNS dispatch whitelist, so e.g. `NOT,((DOMAIN-SUFFIX,foo.com))` silently missed and leaked the query to the global nameserver — exactly the leak this PR set out to prevent. Verifying this exposed a **pre-existing upstream bug**: the classical wrapper's placeholder splice split logic payloads on their inner commas, so *every* logic entry in a classical set was corrupted into a parse error and dropped (on main too). Fixed both:
  - the splice is now logic-aware (adapter appended at the end, mirroring the parser's `splice_inner_adapter`), so classical logic rules parse at all;
  - DNS dispatch now includes logic rules recursively — a rule whose leaves are *all* domain-family is decidable from the query hostname alone, exactly like a plain domain leaf (mihomo parity). One IP-family leaf anywhere under the rule still disqualifies it (no resolver re-entry), and `MATCH` stays excluded. Children are reached via the existing `as_any` downcast at construction time only. Two new tests: pure-domain `NOT`/`AND` rules join dispatch; `IP-CIDR`-leaf `OR` (and a nested `NOT(OR(...))`) stay out while still matching on the full metadata path.
- **N5 (partially confirmed):** the M11 regression test used a `type: file` provider, pinning load ordering but not the network path — annotated as known-untested in the round-2 commit, then actually covered by a new end-to-end test (`2db0280`):
  - a minimal no-auth **SOCKS5 relay** and a **local HTTP payload server**, each on its own OS thread (necessary detail: `load_config` blocks its worker synchronously while fetching, so test-runtime tasks would deadlock);
  - `proxy: remote` with `type: socks5, server: localhost` — a hostname resolved only at dial time;
  - full `load_config_from_str` run asserting the payload **traverses the relay** (SOCKS5 dial counter advances), the provider registers, and the `nameserver-policy` `rule-set:` entry builds — the ordering regression makes the whole load hard-error.
- **M11 (covered):** startup-order test from round 1 pins provider-before-DNS; with the network test the real fetch path is now exercised too.

### Validation

- `cargo test -p meow-rules`: 151 lib tests + integration suites green; `cargo test -p meow-config`: all 68 integration tests incl. the two new ones; `cargo test -p meow-dns` green; `cargo fmt` + `clippy` clean on all touched crates.

PR #470, opened 2026-08-24
fix(anytls): write auth record in a single TLS record

### LINKED ISSUE #469: All AnyTLS nodes handshake-fail against reference servers (auth split across multiple TLS records)
## Summary

Every AnyTLS node fails the protocol handshake against reference AnyTLS servers (anytls-go / sing-box / sing-anytls). The TCP connection and TLS handshake succeed, but the server closes the session immediately during the post-TLS authentication phase — before SYNACK — so the client sees:

```
[Client] SYNACK error: Protocol error: Session closed
```

This is **not** a network problem (TCP direct-connect to the port works) and **not** the nodes restricting clients: the same server accepts other clients fine. It is a wire-level incompatibility in meow-rs's AnyTLS client.

## Root cause

The reference AnyTLS server authenticates with a **single read** off the TLS connection, then parses the auth record from that one buffer:

```go
// sing-anytls service.go (anytls-go / sing-box use the same path)
n, err := b.ReadOnceFrom(conn)        // ONE read
by, err := b.ReadBytes(32)            // SHA256(password)  — from that buffer
by, err = b.ReadBytes(2)              // padding0 length    — from that buffer
```

On a rustls TLS stream, **each `write_all` produces its own TLS record**. meow-rs's `send_authentication` (`crates/meow-anytls/src/util/auth.rs`) writes the auth in **three separate `write_all` calls** — password hash (32 B), padding length (2 B), padding — so it becomes **three TLS records**. The server's single read returns only the first 32-byte record, then `ReadBytes(2)` hits EOF:

```
ERROR inbound/anytls: process connection: EOF: read padding length: fallback disabled
```

(`fallback disabled` because the server has no fallback configured.) The server tears the session down; the client's subsequent SYN/Push writes get `Broken pipe`, surfaced as `Protocol error: Session closed` / SYNACK failure.

The Go reference client (`sing-anytls client.go`) builds the whole auth header in one buffer and `WriteTo`s it once (one TLS record), which is why other clients work.

## Reproduce

1. Run a reference server locally (sing-box anytls inbound, self-signed cert, no fallback).
2. meow config with one anytls outbound pointing at it, `skip-cert-verify: true`.
3. `curl --proxy socks5h://127.0.0.1:18080 https://www.gstatic.com/generate_204`.

Before the fix: server log shows `EOF: read padding length: fallback disabled`; curl fails (`Connection reset by peer` / `Broken pipe`). After the fix: `http_code=204` for both HTTP/2 and HTTP/1.1, server shows a normal proxied connection.

## Fix

Coalesce the entire auth record (`SHA256(password) (32) + padding0 length (2) + padding0`) into a single buffer and write it with one `write_all`, matching `sing-anytls client.go`'s single `WriteTo`. This emits one TLS record so the server's single-read auth succeeds.

## Notes

- The vendored source (`crates/meow-anytls`, from the `madeye/anytls-rs` fork of `jxo-me/anytls-rs`) carries the same 3-write `send_authentication`, so re-vendoring would reintroduce the bug. The independent `ssrlive/anytls-rs` implementation already does the single-write form (`src/client_runner.rs`, builds `auth_data` then one `write_all`) — confirming the fix approach.
- The server-side `authenticate_client` in the same file uses `read_exact` (incremental) and is unaffected; only the client send path needs the change.
- A regression test is added that asserts the whole auth record leaves in a single `poll_write` (a real-TLS single-read test would false-pass because rustls coalesces already-arrived records into one `poll_read`).

aimkiray:
## What

Fixes the all-AnyTLS-nodes-handshake-fail bug reported in #469.

The reference AnyTLS server (anytls-go / sing-box / sing-anytls) authenticates with a **single read** off the TLS connection, then parses `SHA256(password) (32) + padding0 length (2) + padding0` from that one buffer. On a rustls TLS stream each `write_all` is its own TLS record, so meow-rs's three separate `write_all` calls in `send_authentication` produced three TLS records: the server's single read returned only the 32-byte password-hash record and EOF'd reading the 2-byte padding length (`EOF: read padding length: fallback disabled`), tearing the session down before SYNACK. Every AnyTLS node failed against such servers (not a network issue, not node-side client restriction).

## Fix

Coalesce the whole auth record into one buffer and write it with a single `write_all`, matching `sing-anytls client.go`'s single `WriteTo`. This emits one TLS record so the server's single-read auth succeeds. Server-side `authenticate_client` (uses `read_exact`, incremental) is untouched; only the client send path changes.

## Test

Added `send_authentication_writes_whole_record_in_one_call` — a `WriteRecorder` asserts the whole auth record leaves in a single `poll_write` (length, password-hash prefix, padding-length field, zero-filled padding). Verified the test **fails** under the old 3-write code (`got 3 writes`) and **passes** with the fix.

A real-TLS single-read test is deliberately **not** used as the guard: rustls coalesces multiple already-arrived records into one `poll_read`, so it would false-pass under the old multi-write code. The `WriteRecorder` pins the precise cause instead.

## Verification

- `cargo test -p meow-anytls --lib` → 4/4 (incl. new test)
- `cargo test -p meow-proxy --features anytls --test anytls_integration` → 7/7 (no regression; note these use meow's own server whose incremental `read_exact` auth never caught this)
- three-way clippy (`--all-targets`, `--no-default-features`, `--all-features`) → clean
- `cargo fmt --all -- --check` → clean
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` → clean
- End-to-end against a local sing-box anytls inbound (reference server): HTTP/2 → `204`, HTTP/1.1 → `204`, server log no longer shows `fallback disabled`; 10/10 repeated requests stable.

## Notes

- The vendored source (`crates/meow-anytls`, from the `madeye/anytls-rs` fork of `jxo-me/anytls-rs`) carries the same 3-write `send_authentication`, so a future re-vendor would reintroduce this. The independent `ssrlive/anytls-rs` already does the single-write form (`src/client_runner.rs`), confirming the approach.

Closes #469

### DIFF
--- a/crates/meow-anytls/src/util/auth.rs
+++ b/crates/meow-anytls/src/util/auth.rs
@@ -59,9 +59,6 @@ pub async fn send_authentication<W: AsyncWriteExt + Unpin>(
     password_hash: &[u8; 32],
     padding_factory: &Arc<PaddingFactory>,
 ) -> Result<()> {
-    // Write SHA256(password)
-    writer.write_all(password_hash).await?;
-
     // Get padding0 length from padding scheme
     let padding_sizes = padding_factory.generate_record_payload_sizes(0);
     let padding_len = padding_sizes.first().copied().unwrap_or(0);
@@ -73,15 +70,26 @@ pub async fn send_authentication<W: AsyncWriteExt + Unpin>(
         padding_len as u16
     };
 
-    // Write padding0_length
-    writer.write_all(&padding_len.to_be_bytes()).await?;
-
-    // Write padding0
+    // Coalesce the entire auth record — SHA256(password) (32) + padding0
+    // length (2) + padding0 — into a single buffer and write it in one shot.
+    //
+    // The reference anytls server (anytls-go / sing-anytls) authenticates with
+    // a *single* read off the TLS connection (`ReadOnceFrom`), then parses 32
+    // + 2 bytes from that buffer. Each `write_all` on a rustls TLS stream
+    // emits its own TLS record, so issuing three separate writes produces
+    // three TLS records: the server's single read returns only the first
+    // 32-byte record and then EOFs reading the 2-byte padding length,
+    // yielding `EOF: read padding length: fallback disabled` and tearing the
+    // session down before SYNACK. The Go client builds the whole header in
+    // one buffer and `WriteTo`s it once (sing-anytls client.go); mirror that.
+    let mut buf = Vec::with_capacity(32 + 2 + padding_len as usize);
+    buf.extend_from_slice(password_hash);
+    buf.extend_from_slice(&padding_len.to_be_bytes());
     if padding_len > 0 {
-        let padding = vec![0u8; padding_len as usize];
-        writer.write_all(&padding).await?;
+        buf.resize(buf.len() + padding_len as usize, 0);
     }
 
+    writer.write_all(&buf).await?;
     writer.flush().await?;
 
     Ok(())
@@ -91,7 +99,52 @@ pub async fn send_authentication<W: AsyncWriteExt + Unpin>(
 mod tests {
     use super::*;
     use crate::padding::PaddingFactory;
-    use tokio::io::duplex;
+    use std::io;
+    use std::pin::Pin;
+    use std::task::{Context, Poll};
+    use tokio::io::{AsyncWrite, duplex};
+
+    /// A writer that records every `poll_write` call verbatim.
+    ///
+    /// Used to assert that [`send_authentication`] emits the entire auth
+    /// record in a *single* write — the property the reference anytls server
+    /// (anytls-go / sing-anytls) depends on. That server authenticates with
+    /// one `ReadOnceFrom(conn)` then parses `SHA256(password) (32) + padding0
+    /// length (2) + padding0` from that one buffer. On a rustls TLS stream one
+    /// `write_all` is one TLS record, so splitting the auth across multiple
+    /// writes makes the server's single read return only the first record and
+    /// EOF on the padding length (`EOF: read padding length: fallback
+    /// disabled`), tearing the session down before SYNACK.
+    ///
+    /// A real-TLS single-read test is *not* a reliable guard here: rustls
+    /// coalesces multiple already-arrived records into one `poll_read`, so it
+    /// would false-pass under the old multi-write code. This recorder pins
+    /// the precise cause instead — the whole record is written in a single
+    /// `poll_write` — which is what yields one TLS record on the wire.
+    #[derive(Default)]
+    struct WriteRecorder {
+        writes: Vec<Vec<u8>>,
+    }
+
+    impl AsyncWrite for WriteRecorder {
+        fn poll_write(
+            mut self: Pin<&mut Self>,
+            _cx: &mut Context<'_>,
+            buf: &[u8],
+        ) -> Poll<io::Result<usize>> {
+            // Always accept the full buffer so `write_all` resolves in a
+            // single `poll_write` — matching a real TLS stream that takes the
+            // whole (small) record at once.
+            self.writes.push(buf.to_vec());
+            Poll::Ready(Ok(buf.len()))
+        }
+        fn poll_flush(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+            Poll::Ready(Ok(()))
+        }
+        fn poll_shutdown(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+            Poll::Ready(Ok(()))
+        }
+    }
 
     #[tokio::test]
     async fn test_authentication_success() {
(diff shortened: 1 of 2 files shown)


madeye:
can we add a regression test for this fix?

aimkiray:
> can we add a regression test for this fix?

Will do!

aimkiray:
Regression test added in 9c47bdf, addressing @madeye's request.

`anytls_auth_record_arrives_in_single_tls_record` lives in `crates/meow-proxy/tests/anytls_integration.rs` (a consumer of the vendored crate, so it survives a future re-vendor of `meow-anytls`) and exercises the real `AnytlsAdapter` end to end through a real rustls TLS stack.

It drives a record-granular TLS server — parse the 5-byte TLS record header off the raw TCP stream, feed exactly one record at a time to a manually-driven `rustls::server::ServerConnection`, read back that record's plaintext — and asserts the first application-data record carries the *entire* auth header (32-byte password hash + 2-byte padding0 length + zero-filled padding0). This pins the true invariant ("the auth is one TLS record") with no timing/coalescing dependency; a naive single-`poll_read` test would false-pass because rustls coalesces already-arrived records.

Verified:
- **Fails** under the old 3-write code: `first TLS record must carry the entire auth header (>= 34 bytes ...); got only 32 bytes — the auth was split across multiple TLS records (regression of #469)`.
- **Passes** with the fix.
- `cargo test -p meow-proxy --features anytls --test anytls_integration` → 8/8; `cargo test -p meow-anytls --lib` → 54/54.
- three-way clippy (`meow-proxy` default / no-default-features / all-features, `--all-targets -D warnings`) clean; `cargo fmt --all -- --check` clean.

The existing `anytls_rs::server::Server`-based integration tests use `read_exact` (incremental) auth, which is why they never caught this; the new test mimics the reference anytls-go/sing-anytls server's single-read auth instead.

crates/meow-anytls/src/util/auth.rs
```diff
@@ -73,15 +70,26 @@ pub async fn send_authentication<W: AsyncWriteExt + Unpin>(
         padding_len as u16
     };
 
-    // Write padding0_length
-    writer.write_all(&padding_len.to_be_bytes()).await?;
-
-    // Write padding0
+    // Coalesce the entire auth record — SHA256(password) (32) + padding0
```
madeye, line 73: **[suggestion]** The 12-line block above the coalesced write restates the bug history, the Go client, rustls record mapping, and the exact server error string. The same essay is then copied into `WriteRecorder`, the unit-test rustdoc, the meow-proxy module banner, `invalid_data`'s rustdoc, and `anytls_auth_record_arrives_in_single_tls_record`. Comments here should say why one write is required, not narrate the incident.

**Suggestion:** Keep one short why-comment at the write site (e.g. anytls-go authenticates with one `ReadOnceFrom`; rustls emits one TLS record per `write_all`) and drop the duplicated design history from the other sites.

crates/meow-proxy/tests/anytls_integration.rs
```diff
@@ -434,3 +434,237 @@ async fn anytls_udp_is_refused_when_not_enabled() {
...
+/// coalesces already-arrived records into one `poll_read`, so it would
+/// false-pass under the old multi-write code. Driving the server
+/// record-by-record — parse the 5-byte TLS record header, feed exactly one
+/// record to `read_tls`, decrypt, read its plaintext — pins the true
+/// invariant ("the auth is one TLS record") with no timing or coalescing
+/// dependency. It works for both TLS 1.2 and 1.3: in 1.3 the client's
+/// Finished travels in a type-23 record on the wire but yields *no*
+/// application plaintext, so `reader().read()` returns 0 for it and the
+/// loop skips it; the first non-empty plaintext is the auth.
+fn invalid_data<E: std::fmt::Display>(e: E) -> std::io::Error {
```
madeye, line 474: **[suggestion]** The long `///` comment describing record-granular capture (`Drive a raw rustls server…`) is attached to `invalid_data`, a one-line `InvalidData` constructor, not to `capture_first_appdata_plaintext`. That helper's rustdoc will show up in the wrong place, and the comment also claims TLS 1.3 Finished makes `reader().read()` return 0; the code correctly treats `WouldBlock` as empty, while `Ok(0)` is EOF.

**Suggestion:** Move the docs onto `capture_first_appdata_plaintext` (or delete them; the `//` banner above already explains the test) and, if kept, say `WouldBlock` rather than `returns 0`.

crates/meow-proxy/tests/anytls_integration.rs
```diff
@@ -434,3 +434,237 @@ async fn anytls_udp_is_refused_when_not_enabled() {
...
+/// off the TLS connection. The `anytls_rs::server::Server`-based tests above
+/// use `read_exact` (incremental) auth and so never caught the multi-write
+/// bug; this test drives a record-granular TLS server (see
+/// [`capture_first_appdata_plaintext`]) and asserts the first application
+/// record carries the *entire* auth header. It lives in `meow-proxy`, not
+/// the vendored `meow-anytls` crate, so a future re-vendor cannot silently
+/// drop it.
+#[tokio::test]
+async fn anytls_auth_record_arrives_in_single_tls_record() {
+    install_crypto_provider();
```
madeye, line 576: **[suggestion]** The new test is justified as the re-vendor-proof guard (`#![cfg(feature = "anytls")]` in `meow-proxy`). CI never runs `cargo test -p meow-proxy --features anytls --test anytls_integration` (unlike the explicit hysteria2/snell feature-gated steps). `cargo test --lib` will run the meow-anytls unit test, but a future re-vendor that drops that unit test would still land green.

**Suggestion:** Add a CI / local-bar step that runs `cargo test -p meow-proxy --features anytls --test anytls_integration`, or state that the meow-anytls unit test is the CI gate and keep the integration test as opt-in.

PR #473, opened 2026-08-25
feat: pluggable TcpDialer for dialer-proxy chaining (Phase 1)

### LINKED ISSUE #474: dialer-proxy: pluggable TcpDialer for transparent proxy chaining
# dialer-proxy: pluggable TcpDialer for transparent proxy chaining

## Background

Issue #210 implemented `dialer-proxy` support via `DialerProxyAdapter`, which wraps an outbound and dials through a front proxy using the existing `relay_tcp` algorithm. While this works for protocols that implement `connect_over` (HTTP, SOCKS5, Snell), it cannot work for the majority of TLS-based protocols.

## Problem

`DialerProxyAdapter` uses `relay_tcp` → `connect_over` on the final hop. However, `connect_over` is only implemented by HTTP, SOCKS5, and Snell — all other adapters (VLESS, Trojan, VMess, Shadowsocks) use the trait default which returns `Err(NotSupported)`.

This means `dialer-proxy` **does not work at all** for VLESS, Trojan, VMess, and Shadowsocks. The connection fails with `NotSupported` before any protocol handshake occurs.

Even for Snell, `connect_over` only writes the Snell header without performing TLS — so Snell with TLS is also broken in the relay path.

Additionally:

1. **UDP is hard-blocked.** `dial_udp` returns `MeowError::UdpNotSupported` for all dialer-proxy outbounds.

2. **Protocol-specific optimizations are lost.** `set_nodelay`, ALPN negotiation, uTLS fingerprinting, and other per-adapter connection setup all happen in `dial_stream`/`dial_tcp` — none of them run through the `connect_over` path.

## Proposed Solution

Introduce a pluggable `TcpDialer` trait mirroring mihomo's `C.Dialer` / `proxyDialer` architecture:

```rust
pub trait TcpDialer: Send + Sync {
    async fn dial(&self, host: &str, port: u16) -> Result<Stream>;
    fn is_proxy(&self) -> bool { false }
}
```

Each TCP-based adapter holds an `Arc<dyn TcpDialer>` and calls `dialer.dial()` instead of `connect_tcp_host()` directly. When `dialer-proxy` is configured, a `ProxyDialer` is injected that tunnels through the front proxy — making chaining **transparent to the adapter's TLS + protocol handshake**.

### Key difference

| Protocol | Existing (relay-based) | Proposed (pluggable dialer) |
|----------|------------------------|----------------------------|
| VLESS+Reality | ❌ `NotSupported` (no `connect_over`) | ✅ Works |
| Trojan+TLS | ❌ `NotSupported` (no `connect_over`) | ✅ Works |
| VMess+TLS | ❌ `NotSupported` (no `connect_over`) | ✅ Works |
| Shadowsocks | ❌ `NotSupported` (no `connect_over`) | ✅ Works |
| HTTP | ✅ Works (`connect_over` implemented) | ✅ Works |
| SOCKS5 | ✅ Works (`connect_over` implemented) | ✅ Works |
| Snell (no TLS) | ✅ Works (`connect_over` writes header) | ✅ Works |
| Snell (with TLS) | ❌ `connect_over` skips TLS | ✅ Works |
| UDP | ❌ Hard-blocked | ⚠️ TCP only (`is_proxy()` guard on SS `support_udp()`) |
| `set_nodelay` | ❌ Lost in relay path | ✅ `DirectDialer` sets it; `ProxyDialer` does not |
| Nested chains | ✅ Works | ✅ Works (topological resolve order unchanged) |

### Backward compatibility

- `parse_proxy()` delegates to `parse_proxy_with_dialer(raw, &DirectDialer)` — zero behavior change for existing callers.
- `DialerProxyAdapter` is retained as a fallback when re-parsing fails (e.g., test configs without `type` field).
- SS `PluginKind::External` (SIP003) bypasses the dialer (plugins listen on localhost; tunnelling would fail).

## Scope (Phase 1)

- `TcpDialer` trait + `DirectDialer` + `ProxyDialer` + `ConnStream` (`meow-proxy/src/dialer.rs`)
- All TCP-based adapters updated: VLESS, Trojan, SS, VMess, HTTP, SOCKS5, Snell
- `parse_proxy_with_dialer` threading in `meow-config`
- `apply_dialer_proxies` re-parse model (mihomo-compatible)
- `is_proxy()` guard on SS `support_udp()` to prevent UDP traffic leak

## Testing

- 364 meow-proxy unit tests pass
- 198 meow-config unit tests pass (including 2 new dialer_proxy tests)
- 37-scenario integration matrix tested against local sing-box (all protocols × all dialer-proxy combinations, nested 3-level chains, error handling)
- Real-world test: replaced sing-box `detour` chain (SS → VLESS+Reality) with meow-rs `dialer-proxy`, downloaded 848 MB file from GitHub at 3.3 MB/s (server bandwidth capped at 30 Mbps), zero errors

aimkiray:
# feat: pluggable TcpDialer for dialer-proxy chaining (Phase 1)

Closes #474

## Summary

Introduce a pluggable dialer model mirroring mihomo's `C.Dialer` / `proxyDialer` architecture. Each TCP-based adapter now holds an `Arc<dyn TcpDialer>` and calls `dialer.dial()` instead of `connect_tcp_host()` directly. When `dialer-proxy` is configured, a `ProxyDialer` is injected that tunnels through another proxy — making chaining transparent to the adapter's TLS + protocol handshake.

This fixes the fundamental limitation of the existing `DialerProxyAdapter` (relay-based approach): VLESS, Trojan, VMess, and Shadowsocks now work with `dialer-proxy` because the full `dial_stream` (including TLS handshake) runs on the transparently tunneled stream, instead of relying on `connect_over` which these protocols do not implement.

### New file

- **`crates/meow-proxy/src/dialer.rs`** — `TcpDialer` trait, `DirectDialer` (default, uses `connect_tcp_host`, sets `TCP_NODELAY`), `ProxyDialer` (tunnels through another proxy via `ProxyConn`), `ConnStream` (bridges `Box<dyn ProxyConn>` → `Stream` for blanket impl).

### Adapter updates (meow-proxy)

| Adapter | Changes |
|---------|---------|
| VLESS | `dial_stream` + mux `DialFn` use `dialer.dial()` |
| Trojan | `open_tls_with_header` + mux `DialFn` use `dialer.dial()` |
| Shadowsocks | All `dial_tcp_stream` branches (obfs, v2ray-plugin, ech-tls-tunnel, none) + `SsCore` mux `DialFn`; `PluginKind::External` bypasses dialer (SIP003 plugins are local) |
| VMess | `dial_vmess` + mux `DialFn` use `dialer.dial()` |
| HTTP | `dial_stream` uses `dialer.dial()` |
| SOCKS5 | `dial_stream` uses `dialer.dial()` |
| Snell | `dial_fresh` uses `dialer.dial()` |
| `v2ray_plugin` | `dial()` accepts `&dyn TcpDialer` |
| `ech_tls_tunnel` | `dial()` accepts `&dyn TcpDialer` |

### Config injection (meow-config)

- `parse_proxy_with_dialer(config, dialer)` threads the dialer into all adapter constructors.
- `parse_proxy()` delegates to `parse_proxy_with_dialer` with `DirectDialer` (zero behavior change).
- `apply_dialer_proxies` re-parses with `ProxyDialer`; falls back to `DialerProxyAdapter` wrapper if re-parsing fails.
- New `is_proxy()` method on `TcpDialer` trait: SS `support_udp()` checks it to prevent UDP traffic leak through dialer-proxy (plain UDP bypasses the TCP tunnel).

### Test updates

- All adapter integration tests updated with `DirectDialer` parameter.
- 2 new `dialer_proxy_tests` in `meow-config`: `re_parse_injects_proxy_dialer_for_real_protocol`, `fallback_wraps_when_re_parse_fails`.

### Why not keep using `relay_tcp`?

`relay_tcp` calls `connect_over` on the final hop, but `connect_over` is only implemented by HTTP, SOCKS5, and Snell. VLESS, Trojan, VMess, and Shadowsocks use the trait default which returns `Err(NotSupported)`. The relay approach literally cannot work for these protocols — the connection fails before any handshake occurs.

Even for Snell, `connect_over` only writes the protocol header without performing TLS, so Snell with TLS is also broken in the relay path.

The pluggable dialer approach fixes this by injecting `ProxyDialer` at the TCP dial layer, so the adapter's full `dial_stream` (including TLS handshake, Reality exchange, uTLS fingerprinting, ALPN negotiation) runs unchanged on the transparently tunneled stream.

### Why `ConnStream` instead of using `connect_over`?

The `Stream` type is a sized newtype (`Box<dyn Stream>` wrapper) that implements the blanket `Stream` trait. `ProxyConn` is `?Sized`. `ConnStream` bridges `Box<dyn ProxyConn>` → `Stream` so the tunneled connection can be passed to transport layers (TLS, WS, gRPC) that expect `Stream`.

### Why `is_proxy()` on the trait?

SS plain UDP uses a raw socket (`UdpSocket::bind`) that bypasses the TCP dialer entirely. If `dialer-proxy` is configured, UDP packets would leak from the real source IP instead of going through the front proxy. `is_proxy()` returns `true` for `ProxyDialer`, and SS `support_udp()` checks it to disable plain UDP. Mux UDP (which goes over TCP through `dialer.dial()`) is unaffected. VLESS/Trojan/Snell UDP goes through `dialer.dial()` (TCP), so they're unaffected. VMess UDP is not implemented.

### `set_nodelay` placement

`DirectDialer::dial()` calls `set_nodelay(true)` once on the raw TCP stream. `ProxyDialer` correctly does **not** set it — the underlying connection is managed by the front proxy, which should control Nagle's algorithm.

## Backward compatibility

- `parse_proxy()` → `parse_proxy_with_dialer(raw, &DirectDialer)` — existing callers see zero change.
- `DialerProxyAdapter` retained as fallback for configs that can't be re-parsed.
- Existing `relay` group functionality unchanged.

## Test results

```
cargo fmt --all -- --check                         → 0 diffs
cargo clippy --all-targets -- -D warnings          → 0 errors
cargo clippy --all-targets --no-default-features   → 0 errors
cargo clippy --all-targets --all-features          → 0 errors
RUSTDOCFLAGS="-D warnings" cargo doc               → 0 errors
cargo test -p meow-proxy --lib                     → 364 passed, 0 failed
cargo test -p meow-config --lib                    → 198 passed, 0 failed
```

### Integration matrix (37 scenarios, local sing-box)

Tested all protocol × dialer-proxy combinations through a local sing-box server:

- **7 baseline** (direct, no dialer-proxy): SS, SS-chacha20, Trojan, VLESS, VMess, HTTP, SOCKS5
- **22 chained**: each protocol through SOCKS5/SS/HTTP/Trojan/VLESS/VMess dialer-proxy
- **3 nested** (3-level: A → B → C)
- **2 HTTP/1.1** comparison
- **2 error handling** (missing dialer-proxy → fallback to direct; self-reference → fallback)
- **1 UDP safety** (SS `udp: true` + dialer-proxy → TCP works, UDP blocked)

**37/37 passed, 0 failed.**

### Real-world test

Tested by replacing a sing-box `detour` chain (SS → VLESS+Reality) with meow-rs `dialer-proxy`:

- All curl tests pass (gstatic, Google, YouTube, ip.sb)
- 10/10 sequential stability, 5/5 concurrent
- Downloaded 848 MB from GitHub at 3.3 MB/s (server bandwidth capped at 30 Mbps), zip integrity verified, zero errors

## Diffstat

```
23 files changed, 662 insertions(+), 91 deletions(-)
```

1 new file (`dialer.rs`), 22 modified files across `meow-proxy` and `meow-config`.

### DIFF
--- a/crates/meow-config/src/lib.rs
+++ b/crates/meow-config/src/lib.rs
@@ -536,6 +536,7 @@ pub fn rebuild_from_raw_with_cache_dir(
 fn apply_dialer_proxies(
     proxies: &mut HashMap<SmolStr, Arc<dyn Proxy>>,
     raw_proxies: &[HashMap<String, serde_yaml::Value>],
+    ipv6: bool,
 ) {
     // Collect proxy -> dialer edges from the raw config.
     let mut pending: Vec<(SmolStr, SmolStr)> = Vec::new();
@@ -598,7 +599,7 @@ fn apply_dialer_proxies(
                 let proxy_dialer: Arc<dyn meow_proxy::dialer::TcpDialer> = Arc::new(
                     meow_proxy::dialer::ProxyDialer::new(Arc::clone(&dialer_proxy)),
                 );
-                match proxy_parser::parse_proxy_with_dialer(raw, &proxy_dialer) {
+                match proxy_parser::parse_proxy_with_dialer(raw, &proxy_dialer, ipv6) {
                     Ok(rebuilt) => {
                         proxies.insert(name.clone(), rebuilt);
                     }
@@ -784,7 +785,7 @@ fn rebuild_from_raw_impl(
 
     // Apply per-outbound `dialer-proxy` wrappers (issue #210). Runs after both
     // leaf proxies and groups are built so a dialer may reference either.
-    apply_dialer_proxies(&mut proxies, raw.proxies.as_deref().unwrap_or(&[]));
+    apply_dialer_proxies(&mut proxies, raw.proxies.as_deref().unwrap_or(&[]), ipv6);
 
     let download_proxy = internal_http::first_named_proxy(raw.proxies.as_deref(), &proxies);
     // Per-provider `proxy:` overrides resolve against the full registry —
@@ -1999,7 +2000,7 @@ mod dialer_proxy_tests {
     fn wraps_proxy_with_dialer() {
         let mut proxies = registry(&["A", "fast"]);
         let before = proxies.clone();
-        apply_dialer_proxies(&mut proxies, &[raw_proxy("A", Some("fast"))]);
+        apply_dialer_proxies(&mut proxies, &[raw_proxy("A", Some("fast"))], true);
         assert!(was_wrapped(&before, &proxies, "A"));
         assert!(!was_wrapped(&before, &proxies, "fast"));
     }
@@ -2008,7 +2009,7 @@ mod dialer_proxy_tests {
     fn self_reference_is_skipped() {
         let mut proxies = registry(&["A"]);
         let before = proxies.clone();
-        apply_dialer_proxies(&mut proxies, &[raw_proxy("A", Some("A"))]);
+        apply_dialer_proxies(&mut proxies, &[raw_proxy("A", Some("A"))], true);
         assert!(!was_wrapped(&before, &proxies, "A"));
     }
 
@@ -2016,7 +2017,7 @@ mod dialer_proxy_tests {
     fn missing_dialer_is_skipped() {
         let mut proxies = registry(&["A"]);
         let before = proxies.clone();
-        apply_dialer_proxies(&mut proxies, &[raw_proxy("A", Some("ghost"))]);
+        apply_dialer_proxies(&mut proxies, &[raw_proxy("A", Some("ghost"))], true);
         assert!(!was_wrapped(&before, &proxies, "A"));
     }
 
@@ -2027,6 +2028,7 @@ mod dialer_proxy_tests {
         apply_dialer_proxies(
             &mut proxies,
             &[raw_proxy("A", Some("B")), raw_proxy("B", Some("A"))],
+            true,
         );
         assert!(!was_wrapped(&before, &proxies, "A"));
         assert!(!was_wrapped(&before, &proxies, "B"));
@@ -2040,6 +2042,7 @@ mod dialer_proxy_tests {
         apply_dialer_proxies(
             &mut proxies,
             &[raw_proxy("A", Some("B")), raw_proxy("B", Some("C"))],
+            true,
         );
         assert!(was_wrapped(&before, &proxies, "A"));
         assert!(was_wrapped(&before, &proxies, "B"));
@@ -2092,11 +2095,11 @@ mod dialer_proxy_tests {
         // below is what actually distinguishes the two paths.
         let mut proxies = registry(&["A", "front"]);
         let parsed_a =
-            proxy_parser::parse_proxy(&raw_socks5_proxy("A", None, true)).expect("parse socks5");
+            proxy_parser::parse_proxy(&raw_socks5_proxy("A", None, true), true).expect("parse socks5");
         proxies.insert(SmolStr::from("A"), parsed_a);
         let before = proxies.clone();
 
-        apply_dialer_proxies(&mut proxies, &[raw_socks5_proxy("A", Some("front"), true)]);
+        apply_dialer_proxies(&mut proxies, &[raw_socks5_proxy("A", Some("front"), true)], true);
 
         assert!(was_wrapped(&before, &proxies, "A"), "A should be re-parsed");
         assert!(
@@ -2136,7 +2139,7 @@ mod dialer_proxy_tests {
             let mut proxies = registry(&["A", "front"]);
             let before = proxies.clone();
 
-            apply_dialer_proxies(&mut proxies, &[raw_typed_proxy("A", ty, Some("front"))]);
+            apply_dialer_proxies(&mut proxies, &[raw_typed_proxy("A", ty, Some("front"))], true);
 
             let after = proxies.get("A").expect("entry survives");
             assert!(
@@ -2174,7 +2177,7 @@ mod dialer_proxy_tests {
 
         // Must not panic and must not spawn: the parse is rejected before
         // `ShadowsocksAdapter::new` runs, so the fallback wrapper is used.
-        apply_dialer_proxies(&mut proxies, &[raw]);
+        apply_dialer_proxies(&mut proxies, &[raw], true);
 
         assert!(
             proxies.contains_key("A"),
@@ -2200,7 +2203,7 @@ mod dialer_proxy_tests {
             serde_yaml::Value::Number(serde_yaml::Number::from(2222)),
         );
 
-        apply_dialer_proxies(&mut proxies, &[first, last]);
+        apply_dialer_proxies(&mut proxies, &[first, last], true);
 
         let rebuilt = proxies.get("A").expect("present after");
         assert_eq!(
@@ -2219,12 +2222,12 @@ mod dialer_proxy_tests {
         let mut proxies = registry(&["A", "front"]);
         // Give "A" a real socks5 adapter so the fallback has something to wrap.
         let parsed_a =
-            proxy_parser::parse_proxy(&raw_socks5_proxy("A", None, true)).expect("parse socks5");
+            proxy_parser::parse_proxy(&raw_socks5_proxy("A", None, true), true).expect("parse socks5");
         proxies.insert(SmolStr::from("A"), parsed_a);
         let before = proxies.clone();
 
         // raw_proxy has no `type` → parse_proxy_with_dialer fails → fallback.
-        apply_dialer_proxies(&mut proxies, &[raw_proxy("A", Some("front"))]);
+        apply_dialer_proxies(&mut proxies, &[raw_proxy("A", Some("front"))], true);
 
         assert!(
             was_wrapped(&before, &proxies, "A"),
@@ -2350,14 +2353,14 @@ mod dialer_proxy_tests {
         let mut proxies: HashMap<SmolStr, Arc<dyn Proxy>> = HashMap::new();
         proxies.insert(
             SmolStr::from("front"),
-            proxy_parser::parse_proxy(&raw_front).expect("parse front"),
+            proxy_parser::parse_proxy(&raw_front, true).expect("parse front"),
         );
         proxies.insert(
             SmolStr::from("inner"),
-            proxy_parser::parse_proxy(&raw_inner).expect("parse inner"),
+            proxy_parser::parse_proxy(&raw_inner, true).expect("parse inner"),
         );
 
-        apply_dialer_proxies(&mut proxies, &[raw_front, raw_inner]);
+        apply_dialer_proxies(&mut proxies, &[raw_front, raw_inner], true);
 
         // Dial a final target through `inner`; `inner` must reach its own
         // server (127.0.0.1:inner_port) *via* `front`.
--- a/crates/meow-config/src/lib.rs
+++ b/crates/meow-config/src/lib.rs
@@ -509,9 +509,20 @@ pub fn rebuild_from_raw_with_cache_dir(
 /// For every proxy that declares `dialer-proxy: <name>`, its registry entry is
 /// re-parsed from the raw config with a [`meow_proxy::dialer::ProxyDialer`]
 /// injected, so the adapter dials its server through `<name>` transparently
-/// (mihomo `proxyDialer` model). If re-parsing fails (e.g. incomplete test
-/// configs or unsupported protocol types), it falls back to wrapping the
-/// existing entry with a [`meow_proxy::DialerProxyAdapter`] (relay chain).
+/// (mihomo `proxyDialer` model).
+///
+/// Adapter types that do not establish their underlying connection through the
+/// pluggable dialer — `anytls`, `hysteria2` (QUIC), and `ss` with an external
+/// SIP003 plugin — reject the injected dialer at parse time. Those fall back to
+/// wrapping the existing entry with a [`meow_proxy::DialerProxyAdapter`] (relay
+/// chain), which works where `connect_over` is implemented (HTTP/SOCKS5/Snell)
+/// and fails loudly at dial time otherwise. The fallback never degrades to a
+/// direct dial, so a configured chain cannot be silently bypassed.
+///
+/// UDP: paths that use a raw datagram socket bypass the TCP dialer entirely
+/// (Shadowsocks plain relay, SOCKS5 UDP ASSOCIATE). Those refuse the
+/// association rather than leaking the real source path; mux-based UDP rides
+/// the dialer over TCP and is unaffected.
 ///
 /// Nested dialer-proxies are resolved deepest-first so each layer sees its
 /// dialer's final (already-rebuilt) form, and dialer-proxy cycles are detected
@@ -571,12 +582,17 @@ fn apply_dialer_proxies(
                 resolved.insert(name);
                 continue;
             };
-            // The inner outbound may have failed to parse earlier; skip if so.
-            // Re-parse with ProxyDialer injected (mihomo model).
-            // Falls back to DialerProxyAdapter wrapper if re-parsing fails
-            // (e.g. incomplete test configs or unsupported protocol types).
+            // Re-parse the raw block with a `ProxyDialer` injected (mihomo
+            // model), so the adapter's own dial + handshake runs on the
+            // tunneled stream.
+            //
+            // `rev()` matters: the registry-building loop above uses `insert`,
+            // so for a config with duplicate `name:` entries the *last* block
+            // wins. A forward `find` here would resurrect the *first* block and
+            // silently swap the running definition out from under the user.
             let raw_proxy = raw_proxies
                 .iter()
+                .rev()
                 .find(|rp| rp.get("name").and_then(|v| v.as_str()) == Some(&*name));
             if let Some(raw) = raw_proxy {
                 let proxy_dialer: Arc<dyn meow_proxy::dialer::TcpDialer> = Arc::new(
@@ -586,15 +602,39 @@ fn apply_dialer_proxies(
                     Ok(rebuilt) => {
                         proxies.insert(name.clone(), rebuilt);
                     }
-                    Err(_) => {
-                        // Fallback: wrap with DialerProxyAdapter (relay_tcp path).
+                    Err(e) => {
+                        // The adapter type cannot carry an injected dialer
+                        // (anytls, hysteria2, SS-with-external-SIP003-plugin) or
+                        // the block is otherwise unparseable. Fall back to the
+                        // relay-based `DialerProxyAdapter`, which preserves the
+                        // pre-dialer behaviour: it works for the protocols that
+                        // implement `connect_over` (HTTP/SOCKS5/Snell) and fails
+                        // loudly at dial time for the rest — never silently
+                        // dialing direct and leaking past the chain.
+                        //
+                        // The inner outbound may itself have failed to parse
+                        // earlier, in which case there is nothing to wrap.
                         if let Some(inner) = proxies.get(&name).cloned() {
+                            warn!(
+                                "proxy '{name}': cannot inject dialer-proxy '{dialer}' \
+                                 ({e}); falling back to the relay-based wrapper"
+                            );
                             let wrapped: Arc<dyn Proxy> =
                                 Arc::new(meow_proxy::DialerProxyAdapter::new(inner, dialer_proxy));
                             proxies.insert(name.clone(), wrapped);
+                        } else {
+                            warn!(
+                                "proxy '{name}': dialer-proxy '{dialer}' not applied \
+                                 ({e}); the outbound itself failed to parse"
+                            );
                         }
                     }
                 }
+            } else {
+                warn!(
+                    "proxy '{name}': dialer-proxy '{dialer}' not applied; no raw \
+                     config block found for this name"
+                );
             }
             resolved.insert(name);
         }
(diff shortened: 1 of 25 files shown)


aimkiray:
# Review: pluggable TcpDialer for dialer-proxy chaining (Phase 1)

## Overall

The direction is right and this should land — but three blocking items need to be
addressed first, one of which inverts the safety property the PR itself claims.

Replacing the relay/`connect_over` model with an injected `TcpDialer` is what
mihomo does, and it is the only way to get the TLS and Reality handshakes of
VLESS/Trojan/VMess/SS to actually run *on top of* the tunnel. I fully agree with
the PR description's analysis of the `relay_tcp` limitation: `connect_over` is
implemented only by HTTP/SOCKS5/Snell, everything else takes the trait default
and returns `Err(NotSupported)`, so the relay path is *structurally* incapable of
working for those protocols. The `ConnStream` bridge from
`Box<dyn ProxyConn>` → `Stream` is a clean solution too.

The problems cluster in two places: the degradation behaviour when injection
fails or is impossible is unsafe, and the UDP-side guard is installed where it
never fires.

Review baseline: `origin/main` (`bc38424`, 0.21.1) plus this PR's three commits
cherry-picked, with no other in-flight PRs present.

---

## Blocking

### 1. The `is_proxy()` UDP leak guard never fires, and it is a regression

The PR description says SS's `support_udp()` checks `dialer.is_proxy()` to stop
plain UDP from bypassing the chain. But **`support_udp()` is not on the UDP
dispatch path**.

`crates/meow-tunnel/src/udp.rs:223` calls straight through after
`resolve_proxy()`:

```rust
match proxy.dial_udp(&metadata).await {
```

Across the whole crate, `meow-tunnel` references `support_udp()` in exactly one
place — a test mock. So when a rule directly references an SS outbound with
`dialer-proxy` + `udp: true`, `ShadowsocksAdapter::dial_udp` still reaches
`bind_udp` + `ProxySocket` and **the real source IP goes straight to the SS
server** while TCP rides the front proxy. That is precisely the outcome a
`dialer-proxy` user cannot tolerate.

`support_udp()` actually affects only two things: LoadBalance member filtering
and the `udp` field in `GET /proxies`.

Worse, this is a **regression**. Before this change the outbound was wrapped in
`DialerProxyAdapter`, whose `dial_udp` hard-returns `MeowError::UdpNotSupported`
— a loud failure. After the change it becomes a silent leak. Under the ADR-0002
taxonomy this is Class A.

**Fix**: move the enforcement into `dial_udp` itself, after the mux branch and
before the plain relay. Keep the `support_udp()` check for display consistency,
but document that it is *not* the enforcement point and that the two must stay in
sync — otherwise the next person walks into the same trap.

### 2. The same leak class exists in SOCKS5 UDP ASSOCIATE

This one was not in my original pass; I found it while implementing the fix. In
`Socks5Adapter::dial_udp` only the TCP control connection goes through
`self.dialer`:

```rust
let mut control = self.dial_stream().await?;      // via dialer
let relay = self.run_udp_associate(&mut control).await?;
...
let socket = meow_common::bind_udp(bind).await?;  // raw socket
socket.connect(relay).await?;                     // straight to the relay
```

The datagrams bypass the chain entirely. Same treatment as issue 1: refuse in
`dial_udp`, keep `support_udp()` in sync.

### 3. `anytls` / `hysteria2` silently discard `dialer-proxy`

These two arms in `proxy_parser.rs` never receive the `dialer` argument:

```rust
"anytls"    => { let adapter = parse_anytls(name, config)?;    ... }
"hysteria2" => { let adapter = parse_hysteria2(name, config)?; ... }
```

`anytls` still dials via `meow_common::connect_tcp_host`
(`anytls_adapter.rs:746`), and `hysteria2` binds its own QUIC socket. So for
these types `parse_proxy_with_dialer` **succeeds while throwing the dialer away**,
and the `Err` fallback never triggers.

Also a regression: previously they were wrapped in `DialerProxyAdapter` →
`relay_tcp` → the `connect_over` default returning `NotSupported`, giving the
user a clear error. Now it silently dials direct, bypassing an explicitly
configured front proxy.

**Fix**: add a guard that returns `Err` on these arms when `dialer.is_proxy()`,
letting the fallback or a hard failure take over. Same for `direct` (a raw egress
by definition). The governing principle: **never silently degrade to a direct
dial when injection is impossible.**

---

## Important

### 4. The fallback swallows the error and is dead code in production

```rust
Err(_) => {
    if let Some(inner) = proxies.get(&name).cloned() { … }
}
```

No `warn!`, so the reason is lost entirely. It is also a no-op in practice:
re-parsing uses the *same* raw block, so anything that parsed successfully the
first time (`lib.rs:671`) will parse again; anything that failed is absent from
`proxies`, making `proxies.get(&name)` return `None` and the branch do nothing.
Today it is reached only by the unit test that deliberately omits `type`.

Once issues 3 and 6 are fixed it becomes a real path, so it should stay — but it
needs a `warn!(…)` with the reason, plus a comment explaining why it is safe
(it never degrades to a direct dial).

### 5. Duplicate names resurrect the wrong definition

```rust
raw_proxies.iter()
    .find(|rp| rp.get("name").and_then(|v| v.as_str()) == Some(&*name));
```

`find` takes the *first* block with that name, while the registry loop at
`lib.rs:671` uses `insert` and therefore keeps the *last*. With duplicate proxy
names, the re-parse overwrites the registry's last definition with the first one
— the user's running config is quietly swapped out. `.rev().find()` fixes it.

### 6. SS SIP003 external plugins get a second subprocess

`ShadowsocksAdapter::new` calls `Plugin::start(...)` in the
`PluginKind::External` arm (`shadowsocks_adapter.rs:143`), spawning a child
process. The re-parse in `apply_dialer_proxies` spawns **another one**. The old
Arc only reclaims its plugin when dropped, but groups were built *before* this
pass and still hold that Arc — so two plugin processes stay alive.

Beyond that, the SS external arm deliberately bypasses the dialer anyway (the
right call — the plugin is a local process reached over loopback), so re-parsing
it is pure waste. The interception has to happen *before*
`ShadowsocksAdapter::new`, because the side effect lives in the constructor.

### 7. Group members hold the pre-injection adapter

The doc comment already acknowledges this ("group members are resolved eagerly
before this pass") and the old implementation behaved the same way, so it is not
newly introduced here. But it is inherent to the "build first, then swap registry
entries" shape — mihomo installs the dialer at adapter construction time and
does not have the problem. See the architecture note below.

---

## Architecture note (Phase 2)

`ProxyDialer` currently owns an `Arc<dyn Proxy>`, and that ownership direction is
what forces the three-stage "build with `DirectDialer` → build groups → re-parse
in topological order" dance. Issues 4, 5, and 6 are all downstream of that
re-parse step.

If `ProxyDialer` instead resolved its target **lazily** by name (holding a `Weak`,
or a by-name lookup handle into the registry), every adapter could get its final
dialer during a **single** parse pass:

- no re-parse → issues 4, 5, 6 all disappear
- group members naturally receive the dialer-equipped adapter → issue 7 disappears
- nested `dialer-proxy` needs no topological loop; cycle detection collapses into
  a one-shot static check at construction, or a depth cap at dial time

The cost is one extra map lookup on the dial path, negligible against a TCP
handshake.

If Phase 2 intends to do this, it is worth stating in the PR description or in #474 — otherwise this re-parse machinery tends to calcify into the permanent
shape.

---

## Minor

**The `ConnStream` doc comment is inaccurate** (`dialer.rs:104-109`). It claims
that going through `ConnStream` makes `as_any_mut()` fail to downcast to
`RealityTlsStream`, so raw-passthrough does not fire, framed as a Phase-1
limitation. But `RealityTlsStream` holds `inner: Box<dyn Stream>`
(`reality_tls.rs:271`), and `spawn_reality_stream` always returns
`RealityTlsStream` as the outer type regardless of whether a `TcpStream` or a
`ConnStream` sits underneath — the downcast still hits and passthrough is
unaffected. As written, the comment will send someone off to "fix" a performance
problem that does not exist.

**The constructor signature churn is broad.** `Socks5Adapter::new`,
`HttpAdapter::new`, `SnellAdapter::new`, `ShadowsocksAdapter::new` and others all
gain a required parameter, dragging 10 test files into signature updates. The
repo already has a builder idiom (`with_udp`, `with_mux`, `with_resolver`);
`with_dialer(dialer)` defaulting to `DirectDialer` would be more consistent and
would shrink the diff considerably.

**A `SocketAddr` → `String` → `IpAddr` round-trip** (around
`shadowsocks_adapter.rs:290`):

```rust
ServerAddr::SocketAddr(sa) => self.dialer.dial(&sa.ip().to_string(), sa.port()).await,
```

`sa` is already a literal address; it is formatted into a `String` only for
`connect_tcp_host` to parse it straight back into an `IpAddr`. Not on the relay
hot path, but the repo carries ADR-0008 allocation constraints and lints like
`format_push_string`. A `dial_addr(&self, addr: SocketAddr)` default method on
the trait handles it.

There is a semantic win alongside: when `ProxyDialer` overrides `dial_addr`, it
should carry the address in `Metadata::dst_ip` rather than rendering it into
`host`, so the front proxy encodes an IP-typed address instead of a domain name
that merely looks like an IP — matching the SOCKS5/Trojan/VLESS address encoding
convention and mihomo's behaviour. Worth adding IP-literal detection in `dial()`
too.

**Docs disagree with actual behaviour:**

- `docs/migration-from-go-mihomo.md:640` states that UDP is not routed through
  the dialer and such associations are refused rather than leaking the real
  source path. Given issues 1/2/3, that holds for neither SS (via a direct rule
  reference), SOCKS5, nor hysteria2/anytls.
- `docs/gap-analysis.md:244` claims support on every outbound type; per issue 3
  this needs to become a per-type enumeration.

**Tests are thin and their discriminator is fragile.** The two new tests
distinguish the code paths indirectly, via whether `support_udp()` returns true
or false — which incidentally demonstrates how brittle that is: after fixing
issue 2, `Socks5Adapter::support_udp()` returns `false` when a `ProxyDialer` is
injected, and the test's discriminator collapses on the spot.

A higher-signal approach is to stand up a local mock SOCKS5 front proxy and
assert that the CONNECT it receives targets the **inner server's address**, not
the final destination — that is the only assertion that genuinely separates the
injection path from the relay wrapper. Three cases are also missing: multi-level
nesting with real types, the dialer-discard behaviour of hysteria2/anytls, and
`dial_udp` refusal for SS + `dialer-proxy` + `udp: true`.

---

## Fixes already prepared

I implemented all of the above and ran it through the regression bar — 7 files,
`+620/-53`. Take or adapt as useful:

- **Issues 1/2** — enforcement moved into `dial_udp` (SS and SOCKS5); mux UDP is
  unaffected since it rides `dialer.dial()`; `support_udp()` retained and
  documented as display-only
- **Issue 3** — new `reject_unthreaded_dialer()`; `anytls`/`hysteria2`/`direct`
  refuse when `is_proxy()`, falling through to a loud failure
- **Issue 4** — fallback gains a `warn!` with the reason, plus a note on why it
  never degrades to a direct dial
- **Issue 5** — `.rev().find()`, aligned with the registry's last-wins semantics
- **Issue 6** — new `is_external_sip003_plugin()`, intercepting before the
  constructor runs
- **Minor** — corrected the `ConnStream` comment; added `dial_addr` to
  `TcpDialer` and overrode it in both dialers; synced both docs
- **Tests** — replaced the fragile discriminator and added 6: the end-to-end
  mock-front-proxy chain assertion, two UDP-refusal regressions, two
  "direct path unaffected" guards, duplicate-name last-wins, and the fallback
  for non-injectable types

---

## Verification status

Run on a clean branch of `origin/main` (`bc38424`) plus this PR's three commits,
with no other in-flight PRs:

| Check | Result |
|---|---|
| `cargo fmt --all -- --check` | pass |
| clippy default / `--no-default-features` / `--all-features` (all `-D warnings`) | pass |
| `RUSTDOCFLAGS="-D warnings" cargo doc --workspace` | pass |
| CI-mirroring test target list | pass |
| `crate_invariants_test` | 9 passed |
| `meow-proxy --lib` | 368 passed |
| `meow-config --lib` | 204 passed |
| `dialer_proxy_tests` | 11 passed |
| socks5 / http / snell integration | 26 passed |

The new end-to-end chain test was mutation-checked: swapping
`ProxyDialer::new(...)` for `DirectDialer` makes it fail as expected (the front
proxy receives no CONNECT within 5s) while the other 10 still pass, confirming the
assertion pins the injection behaviour rather than passing vacuously.

**Not verified**: the 37-scenario sing-box integration matrix and the real-node
smoke tests from the PR description — I cannot reproduce that environment. The
conclusions for issues 1/2/3 come from reading the dispatch path and the parser
arms; the regression tests assert the post-fix refusal, so **I have not captured
packets confirming the pre-fix leak**. If you have that sing-box setup handy, a
capture against "SS + `dialer-proxy` + `udp: true`, with a rule referencing the
node directly" would be the most direct evidence for issue 1.

**One environment-dependent failure, unrelated to this PR**: `meow-dns --lib`'s
`client::tests::udp_client_times_out_on_unroutable` (`client.rs:779`) fails on my
machine — it reproduces identically on pristine `origin/main`, a non-hermetic
test (the kernel returns an immediate error for an unroutable address instead of
timing out). #476 appears to target exactly this. Skipping it, `meow-dns --lib`
is 137 passed.

aimkiray:
## Fixes for the review items

All items from my review above are implemented. Five commits on top of this
branch's `1f862f6`, each building standalone:

```
1fd6e5c docs: correct dialer-proxy coverage and UDP claims
8bd838c fix(config): report re-parse failures, honour last-wins on duplicate names
6cde5c0 fix(config): reject an injected dialer instead of dropping it silently
81474a6 fix(proxy): refuse raw-socket UDP under a dialer-proxy (Class A)
35fc0de feat(dialer): add dial_addr and correct the ConnStream downcast note
```

7 files, +620/-53.

### Blocking items

**Issues 1 and 2 — raw-socket UDP leak** (`81474a6`). Enforcement moved out of
`support_udp()` and into `dial_udp`, which is what `meow-tunnel` actually calls.
For Shadowsocks the refusal sits after the mux branch and before the plain relay,
so mux UDP keeps working over the chain and the error message points at
smux/yamux as the way to carry UDP. SOCKS5 UDP ASSOCIATE gets the same treatment.

`support_udp()` keeps its check, now documented as the *advertised* capability
(LoadBalance filtering, the `udp` field in `GET /proxies`) with a pointer to the
real enforcement point and a note that the two must stay in sync.

**Issue 3 — silently discarded dialer** (`6cde5c0`). `reject_unthreaded_dialer`
returns `Err` for `anytls`, `hysteria2`, and `direct` when `dialer.is_proxy()`.
A `DirectDialer` still parses everything unchanged, so the non-chained path is
untouched. These configs now reach the relay-based fallback, which fails loudly
where `connect_over` is unimplemented rather than dialing direct.

### Important items

**Issue 4 — swallowed error** (`8bd838c`). The fallback warns with the parse
error, distinguishing "the outbound itself failed to parse, nothing to wrap" from
a genuine fallback, and warns when no raw block matches the name. With the parser
guards in place this branch is now a live path rather than dead code.

**Issue 5 — duplicate names** (`8bd838c`). `.rev().find()`, aligned with the
registry's `insert`/last-wins semantics.

**Issue 6 — SIP003 double-spawn** (`6cde5c0`). `is_external_sip003_plugin`
intercepts before `ShadowsocksAdapter::new`, since `Plugin::start` is called from
the constructor. It mirrors the constructor's own built-in dispatch, so
`obfs`/`simple-obfs`/`v2ray-plugin`/`ech-tls-tunnel` are unaffected.

**Issue 7 — group members** left as-is. It predates this PR and is inherent to the
build-then-swap shape; the lazy-resolution note in my review is the real fix and
belongs in Phase 2.

### Minor items

`35fc0de` adds `dial_addr(SocketAddr)` with a default impl; `DirectDialer` hands
the address to `connect_tcp` untouched, and `ProxyDialer` carries it in
`Metadata::dst_ip` instead of rendering it into `host` — so the front proxy
encodes an IP-typed address rather than a domain that looks like an IP, matching
mihomo and the SOCKS5/Trojan/VLESS encoding. `dial()` gained the same IP-literal
detection. The same commit rewrites the `ConnStream` doc note, which described a
downcast breakage that does not occur.

`1fd6e5c` fixes the migration guide and gap analysis.

I did **not** touch the constructor signatures. Converting to
`with_dialer(dialer)` would shrink the diff and match the existing builder idiom,
but it is a mechanical change across 10 test files and is better as its own commit
if you want it — say the word.

### Tests

`re_parse_injects_proxy_dialer_for_real_protocol` is replaced. It discriminated
via `support_udp()`, which the SOCKS5 fix invalidates — a good illustration of why
that discriminator was too weak to rely on.

The replacement, `chain_dials_inner_server_through_front_proxy`, stands up a mock
SOCKS5 front proxy and asserts the CONNECT it receives targets the **inner
server's address**, not the final destination. That is the assertion that
separates the injection path from the relay wrapper. Mutation-checked: swapping
`ProxyDialer::new(...)` for `DirectDialer` makes it fail (no CONNECT within 5s)
while the other 10 still pass, so it is not passing vacuously.

Five more: two UDP-refusal regressions, two direct-dialer guards against
regressing the unchained path, plus coverage for non-injectable types, the
external-SIP003 rejection, and duplicate-name last-wins.

### Verification

Run on this PR's branch as pushed (`1fd6e5c`) — its own base plus the five commits
above, no other in-flight PRs present. I also ran the same bar on a branch rebased
onto current `origin/main` (`bc38424`); both are clean, so the fixes do not depend
on the 0.21.0 → 0.21.1 delta:

| Check | Result |
|---|---|
| `cargo fmt --all -- --check` | pass |
| clippy default / `--no-default-features` / `--all-features` (`-D warnings`) | pass |
| `RUSTDOCFLAGS="-D warnings" cargo doc --workspace` | pass |
| CI-mirroring test target list | pass |
| `crate_invariants_test` | 9 passed |
| `meow-proxy --lib` | 368 passed |
| `meow-config --lib` | 204 passed |
| `dialer_proxy_tests` | 11 passed |
| socks5 / http / snell integration | 26 passed |

Each of the five commits was also built in isolation to confirm the history is
bisectable.

**Still unverified**: the 37-scenario sing-box matrix and the real-node smoke
tests — I can't reproduce that environment. The new tests assert the post-fix
refusals, so the pre-fix leak is still inferred from the dispatch path rather than
confirmed by capture. A capture against "SS + `dialer-proxy` + `udp: true`, rule
referencing the node directly" would close that gap if you have the setup handy.

**Unrelated failure**: `meow-dns --lib`'s
`client::tests::udp_client_times_out_on_unroutable` fails on my machine and
reproduces identically on pristine `origin/main` — non-hermetic, and #476 looks
aimed at it. Skipping it leaves `meow-dns --lib` at 137 passed.

aimkiray:

Verified every finding in the review against this PR's code and pushed one fix-up (`c8c02d1`). Fast-forward push; no history rewritten.

### Findings verified & fixed

- **B2 (confirmed — real bug, fixed):** all three `Metadata` constructions in `TcpDialer::dial`/`dial_addr` inherited from `..Default::default()`, which sets `ConnType::Http` — so every dialer-proxy hop presented itself as a user HTTP connection instead of an internal one. All three sites now explicitly set `Network::Tcp` + `ConnType::Inner`, aligned with `proxy_tcp_exchange` and mihomo's `C.INNER` semantics.
- **M10 (confirmed, documented only per review decision):** the `TcpDialer` trait doc now records the per-connection double-heap-allocation overhead, the RPITIT refactoring candidate, and the ADR-0008 benchmark method for validating it. Code untouched.

### New test

`proxy_dialer_dials_are_inner_tcp_connections` uses a `CapturingProxy` mock and asserts `conn_type == ConnType::Inner && network == Network::Tcp` (plus host/`dst_ip`/`dst_port`) for all three dial shapes: hostname, IP-literal target, and `SocketAddr`.

### Validation

- `cargo test -p meow-proxy`: 369 lib tests + all integration binaries green; `cargo fmt` + `clippy` clean.

PR #476, opened 2026-08-25
test(dns): make UDP timeout test hermetic via a silent loopback peer

### LINKED ISSUE #475: DNS UDP timeout test is non-hermetic (fails on networks that hijack/answer UDP/53)
## Problem

The unit test `meow-dns::client::tests::udp_client_times_out_on_unroutable` is non-hermetic and fails on real networks that don't behave like a "silent black hole".

The test queries the TEST-NET-1 address `192.0.2.1:53` with a 200 ms timeout and asserts `Err(ClientError::Timeout(_))`:

```rust
#[tokio::test]
async fn udp_client_times_out_on_unroutable() {
    // 192.0.2.1/24 is TEST-NET-1, guaranteed not to respond.
    let client = DnsClient::udp("192.0.2.1:53".parse().unwrap())
        .with_timeout(Duration::from_millis(200));
    let r = client.query("example.test", RecordType::A).await;
    assert!(matches!(r, Err(ClientError::Timeout(_))));
}
```

## Reproduction

Reproduced on a clean tree (`git stash` on an unmodified commit) on a host whose default route goes through a home gateway (`192.168.31.1`). The failure is **not** an ICMP-unreachable / non-Timeout error as one might assume — it is worse:

- A direct UDP probe to `192.0.2.1:53` returns a real DNS response in ~1 ms, with the source address spoofed as `192.0.2.1:53`.
- The response is a valid NXDOMAIN with `recursion_available = true` and a root-server SOA in the authority section.
- The test therefore receives `Ok(NXDOMAIN)` and the `matches!(r, Err(ClientError::Timeout(_)))` assertion panics.

Root cause: the upstream router / ISP **hijacks outbound UDP/53** and answers with its own resolver, spoofing the source IP to the original destination. This is common on CN ISP / home-router networks. Other hosts may instead get ICMP port-unreachable (producing an `Io` error, also non-Timeout). Either way the test's assumption — *"192.0.2.1 is guaranteed not to respond"* — does not hold, so the test is environment-dependent.

```
$ cargo test -p meow-dns --lib client::tests::udp_client_times_out_on_unroutable -- --nocapture

thread 'client::tests::udp_client_times_out_on_unroutable' panicked at client.rs:779:9:
assertion failed: matches!(r, Err(ClientError::Timeout(_)))
test client::tests::udp_client_times_out_on_unroutable ... FAILED
```

## Impact

A legitimately green tree fails `cargo test --lib` purely due to the host network, with no code defect. Anyone developing behind a 53-hijacking router hits this.

## Proposed fix

Make the timeout path deterministic and network-independent by using a bound-but-silent `127.0.0.1` UDP socket that never replies, instead of a "guaranteed unroutable" address:

```rust
let sink = tokio::net::UdpSocket::bind("127.0.0.1:0").await.unwrap();
let addr = sink.local_addr().unwrap();
let client = DnsClient::udp(addr).with_timeout(Duration::from_millis(200));
let r = client.query("example.test", RecordType::A).await;
assert!(matches!(r, Err(ClientError::Timeout(_))));
drop(sink);
```

The sink only needs to stay bound for the duration of the query (a closed port would yield `ECONNREFUSED`, not a timeout). A single query sends one datagram, so the kernel recv buffer never fills and no drain task is needed — no leaked background task.

aimkiray:
## Summary

Make the `udp_client_times_out_on_unroutable` unit test hermetic so it no longer depends on the host network silently dropping packets to `192.0.2.1`.

Closes #475.

## Problem

`udp_client_times_out_on_unroutable` queries TEST-NET-1 (`192.0.2.1:53`) with a 200 ms timeout and asserts `Err(ClientError::Timeout(_))`. The test assumes `192.0.2.1` is a silent black hole. On many real networks that assumption is wrong:

- The upstream router/ISP hijacks outbound UDP/53 and answers with its own resolver, spoofing the source IP to the original destination. A direct probe to `192.0.2.1:53` returns a valid NXDOMAIN in ~1 ms, so the test receives `Ok(NXDOMAIN)` and the assertion panics.
- Other hosts get ICMP port-unreachable, producing an `Io` error (also non-Timeout).

Either way a green tree fails `cargo test --lib` purely because of the host network — no code defect. Reproduced on a clean tree via `git stash`.

## Fix

Replace the unroutable address with a bound-but-silent `127.0.0.1` UDP socket that never replies:

```rust
#[tokio::test]
async fn udp_client_times_out_when_no_response() {
    // A silent loopback peer makes the timeout path deterministic
    // regardless of the host network. Using a "guaranteed unroutable"
    // address like 192.0.2.1 is non-hermetic: many ISPs/routers hijack
    // outbound UDP/53 and answer with a spoofed source (Ok/NXDOMAIN) or
    // return ICMP unreachable (an Io error), so the client never reaches
    // Timeout.
    //
    // The sink is never read and never writes a reply. It only has to stay
    // bound for the duration of the query: a closed port would yield
    // ECONNREFUSED on the client's connected socket instead of a timeout.
    // The client sends a single datagram, so the kernel recv buffer never
    // fills and no drain task is needed.
    let sink = tokio::net::UdpSocket::bind("127.0.0.1:0").await.unwrap();
    let addr = sink.local_addr().unwrap();
    let client = DnsClient::udp(addr).with_timeout(Duration::from_millis(200));
    let r = client.query("example.test", RecordType::A).await;
    assert!(matches!(r, Err(ClientError::Timeout(_))));
    drop(sink);
}
```

Why this shape and not the alternatives:

- **Not `set_socket_factory` injection.** That is a process-global `OnceLock` that can be set only once and never reset; using it in a `#[tokio::test]` (which runs in parallel with the rest of the suite in one process) would be racy and pollute other tests. A local loopback socket is per-test, no global state, safe under parallel execution.
- **No drain task.** The client sends exactly one datagram per query, so the sink's kernel recv buffer never fills. An earlier draft spawned an infinite drain loop that leaked the task + socket until process exit; it was unnecessary and has been removed. The sink's only job is to keep the port bound so the client's connected socket has a live peer (a closed port yields `ECONNREFUSED`, not a timeout).
- **No production-code change.** Downstream (`resolver.rs`) already maps every `ClientError` variant to a unified `LookupFailed`, so `Timeout` vs `Io` make no difference to callers. Mapping ICMP-unreachable to `Timeout` in `udp_exchange` would change product semantics and is out of scope for a test fix.

## Verification

- Target test: 10/10 consecutive runs pass, each ~200 ms (the timeout boundary), confirming determinism with no network jitter.
- `cargo test -p meow-dns --lib`: 138 passed / 0 failed — no regression.
- `cargo clippy -p meow-dns --lib --tests`: clean.
- `cargo fmt -p meow-dns --check`: clean.

### DIFF
--- a/crates/meow-dns/src/client.rs
+++ b/crates/meow-dns/src/client.rs
@@ -771,12 +771,25 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn udp_client_times_out_on_unroutable() {
-        // 192.0.2.1/24 is TEST-NET-1, guaranteed not to respond.
-        let client = DnsClient::udp("192.0.2.1:53".parse().unwrap())
-            .with_timeout(Duration::from_millis(200));
+    async fn udp_client_times_out_when_no_response() {
+        // A silent loopback peer makes the timeout path deterministic
+        // regardless of the host network. Using a "guaranteed unroutable"
+        // address like 192.0.2.1 is non-hermetic: many ISPs/routers hijack
+        // outbound UDP/53 and answer with a spoofed source (Ok/NXDOMAIN) or
+        // return ICMP unreachable (an Io error), so the client never reaches
+        // Timeout.
+        //
+        // The sink is never read and never writes a reply. It only has to stay
+        // bound for the duration of the query: a closed port would yield
+        // ECONNREFUSED on the client's connected socket instead of a timeout.
+        // The client sends a single datagram, so the kernel recv buffer never
+        // fills and no drain task is needed.
+        let sink = tokio::net::UdpSocket::bind("127.0.0.1:0").await.unwrap();
+        let addr = sink.local_addr().unwrap();
+        let client = DnsClient::udp(addr).with_timeout(Duration::from_millis(200));
         let r = client.query("example.test", RecordType::A).await;
         assert!(matches!(r, Err(ClientError::Timeout(_))));
+        drop(sink);
     }
 
     #[tokio::test]
--- a/crates/meow-dns/src/client.rs
+++ b/crates/meow-dns/src/client.rs
@@ -783,13 +783,14 @@ mod tests {
         // bound for the duration of the query: a closed port would yield
         // ECONNREFUSED on the client's connected socket instead of a timeout.
         // The client sends a single datagram, so the kernel recv buffer never
-        // fills and no drain task is needed.
+        // fills and no drain task is needed. Binding for the whole test scope
+        // covers that; the sink needs no explicit drop at the end — the query
+        // has already completed by then (review low item).
         let sink = tokio::net::UdpSocket::bind("127.0.0.1:0").await.unwrap();
         let addr = sink.local_addr().unwrap();
         let client = DnsClient::udp(addr).with_timeout(Duration::from_millis(200));
         let r = client.query("example.test", RecordType::A).await;
         assert!(matches!(r, Err(ClientError::Timeout(_))));
-        drop(sink);
     }
 
     #[tokio::test]


aimkiray:

Verified every finding in the review against this PR's code and pushed one fix-up (`34a44b4`). Fast-forward push; no history rewritten.

### Findings verified & fixed

- **Low (confirmed, fixed):** the trailing `drop(sink);` was a no-op — the `Sink`'s value ends at the same point regardless. Removed, with the surrounding comment extended to state the actual pinning mechanism: the silent loopback peer's *bind* (and the held socket) covers the entire test scope, which is what makes the timeout deterministic. No behavior change.

### Why this PR matters (independently confirmed)

The pre-hermetic test (`udp_client_times_out_on_unroutable` in its main form) was observed failing on a dev machine where the unroutable address produces an ICMP-unreachable error (`ClientError::Io`) instead of a timeout — i.e., the exact flakiness this PR eliminates, reproduced outside CI. With this PR's loopback-peer approach the same test is green there. This is also why #463 restored main's hunk for this test (its fix-up notes point here).

### Validation

- `cargo test -p meow-dns --lib` green including this suite; `cargo fmt` + `clippy` clean.

PR #483, opened 2026-08-27
perf: suspend idle control-plane work

KawaroX:
## Summary

- skip log event formatting and timestamp creation when `/logs` has no subscribers
- start one shared `/traffic` sampler only while clients are connected
- make lazy fallback/url-test health checks run only after new group traffic

## Behavior

- non-lazy health checks are unchanged
- continuously used lazy groups retain interval-based probing
- unused lazy groups perform no network probes
- `/traffic` clients share one sample and one serialized frame per second
- the first traffic frame establishes a fresh baseline, so historical totals are not reported as the current rate

## Verification

- `cargo test --lib` — 1152 passed, 3 ignored
- targeted log stream, traffic stream, health-check, proxy-group, and idle-sampler tests passed
- `cargo clippy -p meow-app -p meow-api -p meow-proxy -p meow-tunnel --all-targets -- -D warnings`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2361,6 +2361,7 @@ name = "meow-app"
 version = "0.21.2"
 dependencies = [
  "anyhow",
+ "async-trait",
  "base64",
  "clap",
  "dashmap",
@@ -2374,10 +2375,12 @@ dependencies = [
  "meow-listener",
  "meow-proxy",
  "meow-rules",
+ "meow-trie",
  "meow-tunnel",
  "parking_lot",
  "rustls",
  "serde_yaml",
+ "smol_str",
  "tempfile",
  "tokio",
  "tracing",
--- a/crates/meow-api/src/lib.rs
+++ b/crates/meow-api/src/lib.rs
@@ -113,6 +113,7 @@ impl ApiServer {
             listeners: self.listeners.clone(),
             external_ui: self.resolve_external_ui(),
             config_mutation_lock: tokio::sync::Mutex::new(()),
+            traffic_feed: Default::default(),
         });
 
         let app = routes::create_router(state);
--- a/crates/meow-api/src/log_stream.rs
+++ b/crates/meow-api/src/log_stream.rs
@@ -72,6 +72,13 @@ impl<S: tracing::Subscriber> Layer<S> for LogBroadcastLayer {
         event: &tracing::Event<'_>,
         _ctx: tracing_subscriber::layer::Context<'_, S>,
     ) {
+        // The broadcast sender exists for the process lifetime, but most
+        // deployments have no active /logs consumer. Avoid formatting fields,
+        // reading the clock, and constructing a heap-backed message when there
+        // is nobody to receive it.
+        if self.tx.receiver_count() == 0 {
+            return;
+        }
         let level = match *event.metadata().level() {
             tracing::Level::TRACE | tracing::Level::DEBUG => LogLevel::Debug,
             tracing::Level::INFO => LogLevel::Info,
@@ -227,17 +234,20 @@ mod tests {
     }
 
     #[test]
-    fn layer_send_with_no_subscribers_does_not_panic() {
-        // Documented contract: a send Err (no subscribers / channel full) is
-        // acceptable — verify we don't regress to a panicking `.unwrap()`.
+    fn layer_skips_event_recording_with_no_subscribers() {
+        struct PanicOnFormat;
+        impl std::fmt::Debug for PanicOnFormat {
+            fn fmt(&self, _f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                panic!("event fields must not be formatted without log subscribers");
+            }
+        }
+
         let (tx, rx) = broadcast::channel(1);
         drop(rx);
         let layer = LogBroadcastLayer { tx };
         let registry = tracing_subscriber::registry().with(layer);
         subscriber::with_default(registry, || {
-            tracing::info!("payload");
-            tracing::error!("oh no");
+            tracing::info!(message = ?PanicOnFormat);
         });
-        // No panic = pass.
     }
 }
--- a/crates/meow-api/src/routes.rs
+++ b/crates/meow-api/src/routes.rs
@@ -56,6 +56,10 @@ where
         ))
     }
 }
+#[derive(Default)]
+pub struct TrafficFeed {
+    sender: std::sync::Mutex<Option<broadcast::Sender<Arc<str>>>>,
+}
 
 pub struct AppState {
     pub tunnel: Tunnel,
@@ -80,6 +84,9 @@ pub struct AppState {
     /// set_tun_handle has completed, leaving a running TUN device behind an
     /// `enable=false` config.
     pub config_mutation_lock: tokio::sync::Mutex<()>,
+    /// Shared on-demand traffic sampler. No timer runs until a client
+    /// subscribes to `/traffic`.
+    pub traffic_feed: TrafficFeed,
 }
 
 /// The API server owns one raw/runtime configuration, so all mutation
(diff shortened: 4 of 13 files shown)


aimkiray (dismissed):
Thanks for the PR — the overall direction is good. Moving traffic sampling behind an active `/traffic` subscriber and avoiding log-event formatting when `/logs` is unused should reduce idle control-plane overhead without changing the public API shape.

I found a couple of issues that need to be addressed or explicitly documented before merge.

## 1. `MissedTickBehavior::Delay` changes non-lazy health-check behavior

`crates/meow-app/src/health_check.rs:45-48`

The PR description says that non-lazy health checks are unchanged, but this line changes the behavior for every health-check loop:

```rust
ticker.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
```

Tokio's default behavior is `Burst`. With the previous implementation, if a probe took longer than `interval_secs`, the next missed tick was delivered immediately after the probe completed. With `Delay`, the next tick is scheduled after another full interval from the current time.

For example, with `interval: 1` and a probe that takes five seconds:

- before this PR: the loop catches up immediately after the probe completes;
- after this PR: the loop waits for another one-second interval.

This changes the effective health-check cadence and can make a dead member remain marked alive longer than before, especially when the configured interval is shorter than the five-second probe timeout.

Please either keep the existing `Burst` behavior for non-lazy groups and use `Delay` only where the new lazy behavior requires it, or keep `Delay` but update the PR description/configuration documentation and add a regression test describing the intentional schedule change.

At minimum, the current “non-lazy health checks are unchanged” statement should be corrected.

## 2. `load-balance` accepts `lazy`, but no health-check task is ever created

Locations:

- `crates/meow-app/src/health_check.rs:16-20`
- `crates/meow-proxy/src/group/load_balance.rs:17-33`
- `crates/meow-proxy/src/group/load_balance.rs:157-182`

`RawProxyGroup.lazy` is a generic field, but `extract_specs()` still only creates health-check specs for:

```rust
.filter(|g| matches!(g.group_type.as_str(), "fallback" | "url-test"))
```

`LoadBalanceGroup` also has no `UsageTracker` and does not override `usage_generation()`.

This means a configuration such as:

```yaml
proxy-groups:
  - name: LB
    type: load-balance
    proxies:
      - proxy-a
      - proxy-b
    url: https://www.gstatic.com/generate_204
    interval: 300
    lazy: true
```

is accepted, but:

- no health-check loop is spawned for `LB`;
- `lazy` has no effect;
- `LoadBalanceGroup::select()` continues to rely on member `alive()` state that is never updated by the group-level health-check loop.

This also conflicts with the existing load-balance specification, which documents periodic health checks and the `lazy` field in `docs/specs/group-load-balance.md:39-41` and `docs/specs/group-load-balance.md:69-83`.

Please choose one of these directions:

1. If `load-balance` is intended to support the same health-check contract:
   - include `load-balance` in `extract_specs()`;
   - add a `UsageTracker` to `LoadBalanceGroup`;
   - call `touch()` from `dial_tcp`, `dial_udp`, and `unwrap_proxy`;
   - expose `usage_generation()`;
   - add a lazy health-check integration test.

2. If lazy health checks are intentionally limited to `fallback` and `url-test`:
   - reject or warn on `lazy` for `load-balance`; and
   - update `docs/specs/group-load-balance.md` so the accepted configuration surface is not misleading.

The current behavior silently accepts a setting that does nothing.

## 3. The new lazy behavior needs one end-to-end scheduler test

Locations:

- `crates/meow-app/src/health_check.rs:41-111`
- `crates/meow-app/src/health_check.rs:130-137`
- `crates/meow-proxy/src/group/fallback.rs:339-352`
- `crates/meow-proxy/src/group/urltest.rs:451-464`

The new tests cover the `should_probe()` predicate and verify that a dial increments the generation, but they do not exercise the actual asynchronous loop.

The important behavior is:

```text
unused lazy group
→ no probe at the first interval
→ group is used
→ one probe at the next interval
→ no new use
→ later intervals do not probe again
→ group is used again
→ probing resumes
```

Please add at least one deterministic test using paused Tokio time and an injectable/mock probe function. It should also cover the immediate first tick handling in:

```rust
if spec.lazy {
    ticker.tick().await;
}
```

Without an end-to-end test, regressions in `last_probed_generation` handling or tick scheduling can pass while the predicate tests remain green.

## Same-class adjacencies

These are not necessarily blockers for this PR, but should be made explicit:

- `SelectorGroup` does not increment a usage generation. This is consistent with the current `extract_specs()` filter because selector groups do not have an automatic health-check loop.
- `RelayGroup` also does not increment a usage generation. The relay specification explicitly describes relay as a fixed chain without group-level health checks.
- `UsageTracker` using `Ordering::Relaxed` is sufficient for a monotonic event counter. The counter does not publish associated configuration state, so an Acquire/Release pair is not required for correctness here.
- The traffic-feed shutdown double-check is generally sound: subscription creation and sender removal use the same mutex, and a new subscriber arriving before the second check prevents the old sampler from being removed.
- Calling `sample_traffic()` when the first subscriber arrives correctly establishes the baseline. The first emitted frame is sampled one interval later, so traffic accumulated while there were no subscribers is not reported as the current one-second rate.

## Testing performed

On PR commit `9e1ca70`:

- `cargo test --lib`
- `cargo test -p meow-api --lib --tests`
- `cargo test -p meow-proxy --lib group`
- `cargo test -p meow-app --lib health_check`
- `cargo test -p meow-tunnel --lib tun_handle_tests`
- `cargo test --test statistics_test`
- `cargo clippy -p meow-api -p meow-app -p meow-proxy -p meow-tunnel --all-targets -- -D warnings`

All completed successfully.

The implementation itself is clean and the idle `/logs` and `/traffic` optimizations are directionally correct. Before merge, I would request clarification/fix for the changed non-lazy ticker semantics and the currently inert `load-balance.lazy` configuration.

aimkiray:
Follow-ups for the review points — I've pushed two commits (`437060b`, `7cb57d5`) to the PR branch. Leaving the PR description untouched and recording the corrections here instead.

**1. `MissedTickBehavior::Delay` vs the "non-lazy health checks are unchanged" claim.** The claim was inaccurate — `Delay` does change the tick semantics for non-lazy loops, and I kept it deliberately; `437060b` adds the code comment documenting why:

- For `interval >= probe timeout` (default 300 s vs 5 s) no tick is ever missed, and the schedule is identical to before.
- For `interval < probe timeout`, the old default `Burst` degenerates into a back-to-back probe storm: every missed tick is delivered immediately, and a non-lazy loop probes on every tick. `Delay` schedules the next tick a full interval from now. It also prevents a missed-tick probe storm right after system suspend.

So the change is a fix rather than a regression; the only cost is that catch-up after a long stall waits one extra interval, which matters only in the storm case above.

**2. `load-balance` accepting an inert `lazy`.** Agreed the config surface is misleading — but this predates the PR: `extract_specs()`'s filter and the `lazy` field are byte-identical in the base commit (`bc38424`), and no group-level health-check loop has ever existed for `load-balance`. The gap is between `docs/specs/group-load-balance.md` and the implementation, not something this PR introduces. Filed #485 to resolve it (reject-or-warn on `lazy` for `load-balance` + spec correction) rather than expanding this perf PR.

**3. End-to-end scheduler test.** `7cb57d5` adds paused-time tests that drive the real `run_health_check_loop` against a `Tunnel` whose fallback-group members answer probes from a canned `204` response — no network I/O, no real timers, deterministic under `start_paused`. Covered sequence: an unused lazy group probes nothing at the first interval; a use triggers a probe at the next tick; further idle ticks skip; a use again resumes probing; non-lazy groups probe on the immediate first tick. The `if spec.lazy { ticker.tick().await; }` first-tick consumption is exercised by the no-probe-before-first-interval assertion.

On the same-class adjacencies — your reading matches mine (Relaxed is fine for the monotonic counter, selector/relay need no generation, the traffic-feed double-check shutdown is sound, and the baseline sample timing is correct); nothing to change there.

One more finding from auditing the lazy semantics: the first use of a long-idle lazy group can hit a member that died while the group was unprobed, and since nothing marks members dead on dial failure, the misses repeat until the next probe tick (up to one `interval`). mihomo covers this with `GroupBase.onDialFailed` escalation. I have that fix staged on the `perf/dial-failure-escalation` branch — aligned with upstream's exemption/refused/threshold policy — and will propose it as a separate PR after this one merges, to keep this PR's scope on the idle-control-plane work.

PR #484, opened 2026-08-28
fix(release): prevent minimal build from overwriting full artifact

ltsj:
## Summary

The release workflow builds the full binary first and then builds the `minimal`
feature set in the same Cargo target directory. The minimal build overwrites the
full binary, so the tarball and OpenWrt IPK published afterward contain the
minimal feature set instead of the advertised full release.

This is especially visible for `aarch64-unknown-linux-musl`, which runs the
minimal size gate and is used to package the official OpenWrt AArch64 IPKs.

## Changes

- Build the minimal size-check binary with a separate
  `CARGO_TARGET_DIR=target/minimal`.
- Measure the minimal binary from that isolated directory.
- Keep the full binary in the normal target directory for archive and IPK
  packaging.

## Impact

No runtime source code or feature definitions are changed. This only fixes
release artifact selection: published archives and OpenWrt IPKs will contain
the full feature set as intended.

## Verification

- `git diff --check`
- Confirmed the full packaging paths remain unchanged.
- GitHub Actions should verify the AArch64 minimal size gate and package the
  full binary independently.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -117,22 +117,21 @@ jobs:
 
       - name: Build minimal release binary (stripped, for size check)
         if: matrix.minimal_budget_bytes != 0 && runner.os == 'Linux'
+        env:
+          # Keep the size-check build separate from the full release artifact.
+          # Cargo otherwise reuses target/<triple>/release/meow and the
+          # minimal build replaces the binary that is packaged below.
+          CARGO_TARGET_DIR: target/minimal
         run: |
           cargo zigbuild --release --locked \
             --no-default-features --features minimal \
             --target ${{ matrix.target }} \
             --bin meow
-          # Copy minimal binary aside before default binary overwrites artifact.
-          # Binary is already stripped via [profile.release] strip = true in
-          # the workspace Cargo.toml — no post-build strip needed (and the
-          # host `strip` cannot read aarch64 ELF anyway).
-          cp "target/${{ matrix.target }}/release/meow" \
-             "target/${{ matrix.target }}/release/meow-minimal"
 
       - name: Minimal binary size check
         if: matrix.minimal_budget_bytes != 0 && runner.os == 'Linux'
         run: |
-          SIZE=$(stat -c%s "target/${{ matrix.target }}/release/meow-minimal")
+          SIZE=$(stat -c%s "target/minimal/${{ matrix.target }}/release/meow")
           BUDGET=${{ matrix.minimal_budget_bytes }}
           echo "Minimal binary size: ${SIZE} bytes (budget: ${BUDGET} bytes)"
           if [ "${SIZE}" -gt "${BUDGET}" ]; then


aimkiray (changes requested):
Thanks for the PR — the diagnosis is correct and the fix addresses the right layer. I verified the reported behavior end to end: the v0.21.2 `aarch64-unknown-linux-musl` tarball ships a 6,115,720 B binary containing only the `minimal` feature set (`ss/trojan/dns-server/listener-mixed`), while the same-tag `aarch64-unknown-linux-gnu` artifact is 12,125,528 B with the full set. Isolating `CARGO_TARGET_DIR` for the size-check build is the correct structural fix — it removes the shared output path entirely rather than shuffling copies around, so this class of bug cannot silently regress.

I verified the change locally on an aarch64 host:

- full build → `target/release/meow` contains VLESS/VMess/Hysteria2/Snell/AnyTLS/mux markers;
- minimal build with `CARGO_TARGET_DIR=target/minimal` → the full binary's hash is unchanged and the markers are absent from the minimal tree;
- reproduced the original bug in the shared-directory layout to confirm the overwrite direction.

Before merge, there are a few items that need to be addressed or explicitly acknowledged.

## 1. The scope of the damage is larger than the PR describes

The PR says the minimal build "overwrites the full binary" as a current-state issue. In fact the ordering `Build full → Build minimal → Package from target/<triple>/release/meow` has been in place since `f249a56` (v0.8.0), with only the strip invocation removed later in `0b100bc`. The comment preserved at the `cp` site — *"Copy minimal binary aside before default binary overwrites artifact"* — describes the overwrite running in the opposite direction of what actually happens, which is consistent with this being an unintentional inversion of the protected artifact rather than a deliberate choice to ship minimal.

Concretely, every aarch64-unknown-linux-musl artifact from v0.8.0 through v0.21.2 is a minimal build, and every OpenWrt AArch64 ipk since v0.17.0 is as well. That is the platform where the documented TUN inbound (`listener-tun`), VLESS, VMess, Hysteria2, Snell, AnyTLS, and sing-mux are all absent from the shipped binary while being documented as release features.

Please update the PR description to state the affected release range. This matters for deciding whether to cut a v0.21.3 with the fix rather than waiting for the next feature release, and for the release notes of whichever release carries the fix.

## 2. The release workflow itself is not exercised before merge

The 8 checks on this PR all come from `test.yml`; `release.yml` only runs on tag pushes and manual dispatch. The changed steps — the `env:` block placement, the `target/minimal/<triple>/release/meow` path in the size-check step, zigbuild honoring a step-level `CARGO_TARGET_DIR` — have not been executed anywhere yet.

The workflow already supports a dry-run path: `workflow_dispatch` with a `tag` input builds and packages without uploading. Please run one dispatch against this branch (or immediately after merge, before the next tag) and confirm from the run log that:

- the aarch64-musl packaged binary is back at the ~11 MiB full size, not ~6 MiB;
- the minimal size check still reads a real file and reports a passing size against the 8 MiB budget.

A path typo in the size-check step would otherwise fail loudly, but a subtly wrong packaged size would not fail anything — which is exactly how the original bug survived.

## 3. Consider one assertion that the packaged binary is the full build

The root cause was that two builds shared one output path, and the only gate (`meow-minimal` size check) validated the artifact that is not shipped. The PR fixes the overwrite, but nothing in the workflow now would catch a future regression of the same shape — for example someone reordering steps, or a new artifact path being added.

A cheap guard in the "Package archive" step would close that class of bug permanently, e.g. assert that the binary being packaged exceeds the maximal minimal-build size, or contains a feature marker unique to the full build. This is not a blocker for this PR, but given the bug's longevity I would like it either added here or tracked in a follow-up issue.

## 4. rust-cache will now cache `target/minimal/`

`Swatinem/rust-cache` caches the whole `target/` tree under the `release-${{ matrix.target }}` key, so the isolated minimal tree (~0.5 GB before save-phase cleanup) will be included for the aarch64-musl job. This is not a correctness problem, and it buys near-incremental minimal builds on subsequent releases since the cache key is `Cargo.lock`-based and feature-set-independent. The tradeoff is cache pressure under the 10 GB per-repo quota across the 11 matrix jobs.

No change requested. Flagging it so the cache-size growth after this lands is recognized as expected rather than investigated as a regression.

## 5. Minor

- The removed comment correctly disappears along with the `cp`; nothing else in the tree references the `meow-minimal` filename, so no dangling documentation.
- `minimal_soft_gate` currently has no effect for any target with `minimal_budget_bytes != 0` (aarch64-musl is the only one, and it is hard-gated). Pre-existing, not introduced here — noting it only because the soft-gate branch referenced ADR-0007 §1 for mipsel, which was dropped from the matrix.

## Testing performed

On PR commit `a232f17`:

- YAML structure of the modified workflow validated (`env:` on the correct step, path in the size-check step consistent with `CARGO_TARGET_DIR` + `--target`);
- local aarch64 reproduction: full build, minimal build into a separate target dir, hash comparison confirming the full binary is untouched, feature-marker comparison (`VLESS: request header sent`, `mux.cool: UDP stream closed`) present in full and absent in minimal;
- downloaded and inspected the v0.21.2 aarch64-musl and aarch64-gnu release artifacts to confirm the shipped-feature discrepancy the PR fixes;
- `git diff --check` clean.

The fix is correct and minimal. Requesting changes only for the PR-description scope update (#1) and the dispatch dry-run verification (#2); #3 and #4 can be follow-ups if you prefer.

aimkiray (approved):
Merging with the fix as-is. The two requested items are tracked as follow-ups rather than pre-merge blockers: the PR-description scope update (affected range v0.8.0–v0.21.2 aarch64-musl / OpenWrt artifacts) will be reflected in the next release notes, and a release dry-run via workflow_dispatch will be performed before the next tag to confirm the packaged binary is back to the full build (~11 MiB). The fix itself is verified correct (see the request-changes review above for the local reproduction and artifact-level evidence).

Issue #425, opened 2026-08-19
smux: port muxcool's deliver_now/Parked flow control (session-wide head-of-line blocking) + NOP keepalive


madeye:
Deferred from the #414 review (finding #1). `STREAM_QUEUE` was raised 2 → 32,
which moves the threshold but leaves the mechanism unchanged: the smux reader
task still awaits the bounded per-stream channel in
`crates/meow-proxy/src/mux/smux.rs`:

```rust
result = tx.send(chunk) => { let _ = result; }
```

A consumer that stops reading blocks the reader task, which halts frame dispatch
for **every** stream on that physical connection — one paused download stalls
every other tab sharing the session. The 4 MiB `receive_budget` semaphore already
bounds memory, so the bounded channel buys no safety here, only stalls.

`mux/muxcool.rs` (#418) implements the target shape and is the reference:
`deliver_now` / `Parked` / the `space` `Notify` park at most one frame and pause
the connection read (session-wide flow control, TCP window semantics) without the
driver ever blocking on a stream queue — so a consumer stalled mid-write cannot
deadlock writers sharing the session.

**Action:** port that pattern to smux. The `STREAM_QUEUE = 32` comment in
smux.rs already names this as the intended follow-up; this issue tracks it so it
does not live only in a code comment.

**Also deferred from the same review (#7):** smux has no `CMD_NOP` keepalive.
xtaci and sing-mux both send them; without one, a half-dead session is only
detected on an IO failure or after the 60 s idle eviction, so a stream opened
onto a silently-dead session pays the full `open_stream_flags` retry.
`mux/muxcool.rs`'s `keepalive_loop` is the in-tree precedent.

madeye:
@aimkiray — owner, please. Deferred by agreement in the #414 review (findings #1 and #7). Your `deliver_now` / `Parked` / `space` design in `mux/muxcool.rs` is the reference implementation; this is the port to smux, plus the `CMD_NOP` keepalive.

aimkiray:
Taking this — the smux scaffold in #414 was mine, so the flow-control follow-up lands naturally with me.

Plan: port muxcool's `deliver_now` / `Parked` / `space` Notify delivery to `smux.rs`, so a full per-stream queue parks at most one frame and pauses the connection read (session-wide flow control) instead of blocking the shared reader task on the bounded channel.

[...]

PR #491, opened 2026-09-02
fix(mux): isolate stalled smux streams

aimkiray:
## Summary

Fixes the smux side of #425 by giving each stream a bounded inbox and isolating a stalled stream instead of pausing the whole session:

- Each stream has a 2048-frame inbox. The reader never awaits it; when an inbox is full the chunk is parked in a per-stream spill (FIFO preserved: the consumer drains the inbox first, then the spill, woken via a notify permit), so a busy-but-draining consumer is never aborted for queue depth alone. Spill bytes stay bounded by the session receive budget because every parked chunk holds its permits.
- The per-stream receive share (`MAX_STREAM_BUFFER`, 128 KiB, i.e. receive budget / 32) is enforced by a per-stream stall watchdog, not at frame arrival: a stream that crosses its share gets one full grace window (`STREAM_STALL_GRACE`, 500 ms), after which the watchdog re-checks the live watermark and retires the stream only if it is still holding more than its share. An ordinary buffered burst — even one that pushes `unread` far past the share while the consumer is momentarily unscheduled — drains inside the window and survives. A consumer still over its share after the window is structurally slower than the peer (smux v1 has no window update, so there is nothing else to hold it back) and is cut instead of pinning budget permits until the session-wide deadline kills the session wholesale.
- The only synchronous retirement left is the adversarial bound: more than inbox+spill (4096) frames parked against a consumer that drained none of them implies sub-32 B average frames — a tiny-frame flood — which ordinary traffic cannot produce before the grace-gated byte cap does.
- Payload for unknown or retired stream IDs is drained with a fixed 8 KiB scratch buffer before it can consume shared receive capacity.
- Stream-map cleanup is synchronous (`parking_lot::Mutex`), so drops outside a Tokio runtime no longer skip cleanup, and no per-drop task is spawned when the outbound queue is backpressured — a single deferred-FIN forwarder preserves FIFO instead.
- A retired stream surfaces `ConnectionAborted` after draining already-buffered data (inbox and spill). Every other stream on the session is untouched.
- If the session-wide 4 MiB receive budget stays saturated for 30 s (`BUDGET_STALL_TIMEOUT`), the reader abandons the whole session. The deadline is armed the first time an acquire blocks and cleared only by an *uncontested* acquire, so a trickle of permits cannot postpone abandonment forever and leave the session wedged while the pool still considers it healthy.
- Protocol errors and budget saturation log at `warn!`, stream retirement at `info!`, so these events survive the default filter. A dropped FIN distinguishes a dead session (channels `Closed`) from a wedged writer (queues `Full`), so the backpressure diagnostic no longer fires on a session that is simply gone.

### Why the per-stream cap is budget/32

`MuxClient::offer` computes `reuse = sessions.len() >= max_connections || load < min_streams` when `max_connections > 0`, and never consults `max_streams` on that path — so a single session can carry an unbounded number of streams. At the default 4/4/4 config, roughly seventeen concurrent connections already place five streams on one session. With an earlier `/4` cap, four stalled streams exhausted the entire budget and wedged the reader for every other stream on the session; at `/32` that takes thirty-two. A `const` assertion pins the divisor so the invariant breaks the build if the constants ever move.

### Failure-model change

This changes the failure model relative to the previous draft (and muxcool): instead of session-wide TCP-window-style backpressure, a stalled consumer is explicitly retired. That is the observable behavior change required to fix the "one stalled stream stalls every other stream" symptom in #425.

smux v1 has no window-update frame, so per-stream flow control is not expressible at all: the reader either parks the session behind a slow consumer (muxcool's choice, and the head-of-line blocking #425 reports) or cuts that consumer. This PR takes the second branch, grace-gated: the cut lands only after the consumer has been over its share for a full 500 ms window — recovery-based (watermark), not progress-based, so a slow-but-draining consumer that crosses the cap for under the window always survives, while one that holds the share for the full window is cut.

### Review revision: healthy bursts must not be aborted

The first revision retired a stream the moment `unread` exceeded its share, at frame arrival. That mis-fired on healthy, actively reading streams: the reader drains a buffered burst back-to-back without yielding to the consumer, so on a current-thread runtime sixteen 32 KiB frames in one burst pushed `unread` past the share while a consumer continuously calling `read_to_end` was merely unscheduled — it got `ConnectionAborted` after exactly 131072 bytes (the review's P1 repro). Retirement is no longer decided at arrival time; the grace-gated watchdog above fixes it deterministically, because the consumer being alive is exactly what clears the watermark inside the window.

## Test plan

- [x] 24 smux tests, including: the reviewer's 16×32 KiB buffered-burst repro (current-thread runtime, 512 KiB delivered with clean EOF); over-share recovery inside the grace window; a stalled hog retired only after the grace window while its peer and the session survive; a never-draining inbox overflow retired after the grace window with the 2049th byte preserved through the spill; the tiny-frame-flood last resort retiring immediately; draining a stream frees its receive cap so it parks rather than aborts; malformed control frames do not kill the session; a FIN dropped by full outbound queues does not kill the session; a FIN dropped by an already-dead session is not reported as backpressure; a wedged receive budget abandons the session
- [x] Mutation-tested: restoring arrival-time retirement fails the burst repro on its intended assertion; short-circuiting the watchdog hangs the stalled-hog test; dropping the spill loses the 2049th byte
- [x] Real-server E2E (`crates/meow-app/tests/smux_singbox_integration.rs`): the full stack — config parsing → mixed listener → tunnel routing → VLESS adapter + smux → a real sing-box 1.14.0 VLESS inbound with multiplex enabled → direct outbound → local echo server, and back:
  - a 4 MiB bulk transfer (32× the per-stream receive share) with a continuously reading consumer, byte-identical round trip — the E2E twin of the buffered-burst regression;
  - a slow reader sharing one physical session (`max-connections: 1`) with fast readers: the slow stream is isolated by the stall watchdog while fast readers complete before and after on the very same session.
  The suite **fails** (never silently skips) when sing-box is missing; CI installs the pinned binary in the test job.
- [x] `cargo fmt --all -- --check`; three-way `cargo clippy --all-targets` (default / no-default-features / all-features) — clean
- [x] `cargo test -p meow-proxy --features mux --lib` — 409 passed, 0 failed, 3 ignored
- [x] `cargo test --workspace --lib` — 1281 passed across 13 binaries
- [x] Flakiness: 10× serial iterations of the smux suite and 5× of the real-server E2E suite, 0 failures; the buffered-burst repro is stable 10×

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -43,6 +43,20 @@ jobs:
           sudo apt-get update
           sudo DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends simple-obfs
 
+      # The real-server smux E2E (smux_singbox_integration) REFUSES to
+      # silently skip without this binary. Keep the version in sync with
+      # SINGBOX_VERSION in crates/meow-app/tests/smux_singbox_integration.rs
+      # and with the test.yml install step.
+      - name: Install sing-box (real-server smux E2E)
+        run: |
+          set -euo pipefail
+          SINGBOX_VERSION=1.14.0
+          curl -fsSL -o /tmp/sing-box.tar.gz \
+            "https://github.com/SagerNet/sing-box/releases/download/v${SINGBOX_VERSION}/sing-box-${SINGBOX_VERSION}-linux-amd64.tar.gz"
+          tar -xzf /tmp/sing-box.tar.gz -C /tmp
+          sudo install -m 755 "/tmp/sing-box-${SINGBOX_VERSION}-linux-amd64/sing-box" /usr/local/bin/sing-box
+          command -v sing-box
+
       - name: Collect coverage
         env:
           MIHOMO_REQUIRE_INTEGRATION_BINS: "1"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -85,6 +85,21 @@ jobs:
           command -v obfs-local
           command -v obfs-server
 
+      # Real-server smux E2E (smux_singbox_integration) runs against this
+      # pinned sing-box and REFUSES to silently skip when it is missing,
+      # so a green run always means real-server smux was exercised. Keep
+      # the version in sync with SINGBOX_VERSION in
+      # crates/meow-app/tests/smux_singbox_integration.rs.
+      - name: Install sing-box (real-server smux E2E)
+        run: |
+          set -euo pipefail
+          SINGBOX_VERSION=1.14.0
+          curl -fsSL -o /tmp/sing-box.tar.gz \
+            "https://github.com/SagerNet/sing-box/releases/download/v${SINGBOX_VERSION}/sing-box-${SINGBOX_VERSION}-linux-amd64.tar.gz"
+          tar -xzf /tmp/sing-box.tar.gz -C /tmp
+          sudo install -m 755 "/tmp/sing-box-${SINGBOX_VERSION}-linux-amd64/sing-box" /usr/local/bin/sing-box
+          command -v sing-box
+
       - name: Build
         run: cargo build --tests
 
@@ -117,6 +132,7 @@ jobs:
             --test ws_test \
             --test crate_invariants_test \
             --test crate_publish_metadata_test \
+            --test smux_singbox_integration \
             -- --nocapture
 
       - name: UDP routing regression (SOCKS5, Shadowsocks, TUN)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -148,8 +148,16 @@ cargo test --lib --bin meow \
   --test trojan_integration --test vless_config_test --test vless_integration \
   --test v2ray_plugin_integration --test pre_resolve_test \
   --test tls_test --test ws_test --test crate_invariants_test \
-  --test crate_publish_metadata_test
+  --test crate_publish_metadata_test \
+  --test smux_singbox_integration
 ```
+
+`smux_singbox_integration` runs the full stack (config → mixed listener →
+tunnel → VLESS adapter + smux) against a real sing-box server and **fails**
+(never silently skips) when the binary is missing: install the pinned
+version from https://github.com/SagerNet/sing-box/releases or point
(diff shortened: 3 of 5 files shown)


madeye:
Please address the following before approval. Reviewed head: `a246a6307a34e856ebb5425d3e27197290fbb0c1`.

**[P1] The receive cap aborts healthy, actively reading streams.**

At [smux.rs:348–364](https://github.com/meow-rs/meow-rs/blob/a246a6307a34e856ebb5425d3e27197290fbb0c1/crates/meow-proxy/src/mux/smux.rs#L348-L364), the reader retires a stream immediately when the next frame would exceed 128 KiB of unread data. It can process five 32 KiB frames before the consumer task is scheduled again, so the cap does not distinguish a stalled consumer from an actively reading task.

I reproduced this with a current-thread Tokio test:

1. Create a smux session over a 1 MiB duplex pipe and open a stream.
2. Spawn a task continuously calling `read_to_end`, then yield to let it run.
3. Send sixteen 32 KiB PSH frames (512 KiB total), followed by FIN, in one buffered burst.

The PR implementation returns `ConnectionAborted` after exactly 131,072 bytes. The same probe passes against the pre-PR implementation. All 21 existing smux unit tests pass on the PR head, so they miss this regression. This is a synthetic scheduling/buffering reproduction, not a real-server interoperability test.

Please revise the retirement policy so an ordinary buffered burst does not abort a healthy reader, and add this regression test while retaining coverage for stalled-stream isolation.

**Real-server smux E2E coverage is missing.**

The smux echo, pool, and UDP tests use in-memory/mock peers. The real v2ray-plugin integration test explicitly runs the server with [`mux=0`](https://github.com/meow-rs/meow-rs/blob/a246a6307a34e856ebb5425d3e27197290fbb0c1/crates/meow-proxy/tests/v2ray_plugin_integration.rs#L190-L201), so it does not validate smux interoperability.

Please add an automated local-server test through config parsing → listener → routing → proxy adapter → real sing-box/mihomo server, explicitly selecting `protocol: smux`. Cover large transfers and concurrent fast/slow readers sharing a physical session, and ensure the CI test fails rather than silently skips when its required server is unavailable. A passing unit suite and green CI currently do not establish this behavior.

PR #497, opened 2026-09-03
feat(transport): add XHTTP (splithttp) transport protocol support (#344)

### LINKED ISSUE #344: Support for the XHTTP transport.
Are there any plans to add support for the XHTTP transport? 
At the moment, H2 transport is implemented, but it is only compatible with Sing-box servers, while xray-core servers are much more widespread due to the popularity of web panels such as 3x-ui and others. Xray-core does not support the H2 transport; instead, it uses XHTTP.

ltsj:
feat(transport): add XHTTP (splithttp) transport protocol support (#344)

Implements the XHTTP (splithttp) transport protocol for `meow-rs` (tracking issue #344).

XHTTP is an HTTP/2-based transport protocol widely used in Xray-core, sing-box, and mihomo. Following ADR-0001 layered transport architecture and addressing review feedback on #376:
- `XhttpLayer` is strictly an inner `Transport` layer inside `TransportChain` (applied after `TlsLayer`). Target destination parsing is handled by the proxy protocol (VLESS / VMess) inside the payload, keeping transport concerns decoupled.
- Implements `stream-one` mode (single bidirectional HTTP/2 POST tunnel stream).
- Reuses and extends `H2Stream` from `meow-transport/src/h2_common.rs` with `AbortHandle` lifecycle management to guarantee background HTTP/2 driver tasks are aborted on stream drop/error, eliminating connection leaks.

### 1. `meow-transport` (`crates/meow-transport`)
- **`src/xhttp.rs`**:
  - Implements `XhttpConfig` and `XhttpLayer: Transport`.
  - Supports `:path`, `:authority` selection (uniform randomized selection across multiple configured hosts).
  - Supports `Content-Type: application/grpc` default and `no-grpc-header` toggle.
  - Supports custom HTTP headers (`extra_headers`).
  - Supports `x-padding-bytes` generation formatted as `?x_padding=XXX` in the `Referer` header matching the splithttp specification.
  - Non-blocking lazy response handling via `RecvState::with_timeout(..., 15s, Success, "xhttp")`.
- **`src/h2_common.rs` & `src/h2.rs`**:
  - Extended `H2Stream` with `conn_abort: Option<tokio::task::AbortHandle>` and `with_conn_abort(self, handle)`.
  - Aborts the connection driver task on drop and on handshake/connect failure.
- **`src/error.rs` & `src/lib.rs`**:
  - Added `Xhttp` error variant to `TransportError`.
  - Added `xhttp` feature and exported `pub mod xhttp;`.
- **`tests/xhttp_test.rs`**:
  - 6 integration tests:
    - `xhttp_round_trip_1mib`: 1MiB bidirectional data round-trip over loopback h2.
    - `xhttp_host_selection_is_uniform`: Uniform distribution across multiple hosts.
    - `xhttp_headers_and_no_grpc_header`: Custom headers and `no-grpc-header` behavior.
    - `xhttp_round_trip_with_deferred_response`: Server sending 200 OK after receiving client data.
    - `xhttp_drop_aborts_driver`: Driver background task aborted immediately on stream drop.
    - `xhttp_config_validation`: Path, host, mode, headers, and padding parameter validation.

### 2. `meow-config` (`crates/meow-config`)
- **`Cargo.toml`**: enabled `xhttp` feature in `meow-transport`.
- **`src/proxy_parser.rs`**:
  - `default_transport_alpn`: automatically sets ALPN to `["h2"]` for `xhttp` network.
  - `parse_vless`: parses `network: "xhttp"` and `xhttp-opts`:
    - `path`, `host` (single or sequence), `mode` (`stream-one` and `auto` supported; unsupported modes rejected), `headers`, `no-grpc-header`, and `x-padding-bytes` (range string or array).
- **`tests/vless_config_test.rs`**:
  - 4 config tests verifying minimal, full options, unsupported mode rejection, and invalid padding range rejection.

### 3. `meow-proxy` (`crates/meow-proxy`)
- **`Cargo.toml`**: enabled `xhttp` feature in `meow-transport` dev-dependencies.
- **`tests/vless_integration.rs`**:
  - Added `vless_over_xhttp_roundtrip` end-to-end integration test verifying full-stack VLESS proxy over XHTTP tunnel to echo server.

## Verification
- `cargo fmt --all -- --check`: passed.
- `cargo clippy -p meow-transport -p meow-config --all-targets --all-features -- -D warnings`: passed (0 warnings).
- `cargo test -p meow-transport --features xhttp --test xhttp_test`: passed (6/6).
- `cargo test -p meow-transport --features h2 --test h2_test`: passed (8/8).
- `cargo test -p meow-config --test vless_config_test -- xhttp`: passed (4/4).
- `cargo test -p meow-proxy --test vless_integration -- vless_over_xhttp_roundtrip`: passed (1/1).

Closes #344.

### DIFF
--- a/crates/meow-config/src/proxy_parser.rs
+++ b/crates/meow-config/src/proxy_parser.rs
@@ -1568,9 +1568,123 @@ fn parse_vless(
             };
             chain.push(Box::new(HttpUpgradeLayer::new(hu_cfg)));
         }
+        "xhttp" => {
+            use meow_transport::xhttp::{XhttpConfig, XhttpLayer};
+            let xhttp_opts = config.get("xhttp-opts");
+            let path = xhttp_opts
+                .and_then(|o| o.get("path"))
+                .and_then(|v| v.as_str())
+                .unwrap_or("/")
+                .to_string();
+            let hosts: Vec<String> = xhttp_opts.and_then(|o| o.get("host")).map_or_else(
+                || vec![server.to_string()],
+                |v| {
+                    if let Some(s) = v.as_str() {
+                        vec![s.to_string()]
+                    } else if let Some(seq) = v.as_sequence() {
+                        seq.iter()
+                            .filter_map(|item| item.as_str().map(std::string::ToString::to_string))
+                            .collect()
+                    } else {
+                        vec![server.to_string()]
+                    }
+                },
+            );
+            if hosts.is_empty() {
+                return Err(format!(
+                    "vless: xhttp-opts.host must not be empty for proxy '{name}'"
+                ));
+            }
+            let mode = xhttp_opts
+                .and_then(|o| o.get("mode"))
+                .and_then(|v| v.as_str())
+                .unwrap_or("stream-one")
+                .to_string();
+            if !mode.is_empty()
+                && !mode.eq_ignore_ascii_case("auto")
+                && !mode.eq_ignore_ascii_case("stream-one")
+            {
+                return Err(format!(
+                    "vless: unsupported xhttp mode '{mode}'; only 'stream-one' and 'auto' are supported"
+                ));
+            }
+            let extra_headers: Vec<(String, String)> = xhttp_opts
+                .and_then(|o| o.get("headers"))
+                .and_then(|h| h.as_mapping())
+                .map(|m| {
+                    m.iter()
+                        .filter_map(|(k, v)| {
+                            let key = k.as_str()?.to_string();
+                            let val = v.as_str()?.to_string();
+                            Some((key, val))
+                        })
+                        .collect()
+                })
+                .unwrap_or_default();
+            let no_grpc_header = xhttp_opts
+                .and_then(|o| o.get("no-grpc-header"))
+                .and_then(serde_yaml::Value::as_bool)
+                .unwrap_or(false);
+            let x_padding_bytes =
+                if let Some(padding_val) = xhttp_opts.and_then(|o| o.get("x-padding-bytes")) {
+                    if let Some(s) = padding_val.as_str() {
+                        let parts: Vec<&str> = s.split('-').collect();
+                        if parts.len() == 2 {
+                            let min = parts[0].trim().parse::<usize>().map_err(|e| {
+                                format!("vless: invalid min in x-padding-bytes '{s}': {e}")
+                            })?;
+                            let max = parts[1].trim().parse::<usize>().map_err(|e| {
+                                format!("vless: invalid max in x-padding-bytes '{s}': {e}")
+                            })?;
+                            if min > max {
+                                return Err(format!(
+                                    "vless: x-padding-bytes min ({min}) exceeds max ({max})"
+                                ));
+                            }
+                            Some((min, max))
+                        } else {
+                            return Err(format!(
+                                "vless: invalid x-padding-bytes range '{s}', expected 'min-max'"
+                            ));
+                        }
+                    } else if let Some(seq) = padding_val.as_sequence() {
+                        if seq.len() == 2 {
+                            let min = seq[0].as_u64().ok_or_else(|| {
+                                "vless: invalid min in x-padding-bytes".to_string()
+                            })? as usize;
+                            let max = seq[1].as_u64().ok_or_else(|| {
+                                "vless: invalid max in x-padding-bytes".to_string()
+                            })? as usize;
+                            if min > max {
+                                return Err(format!(
+                                    "vless: x-padding-bytes min ({min}) exceeds max ({max})"
+                                ));
+                            }
+                            Some((min, max))
+                        } else {
+                            return Err(
+                                "vless: x-padding-bytes array must have 2 elements".to_string()
+                            );
+                        }
+                    } else {
+                        None
+                    }
+                } else {
+                    Some((100, 1000))
+                };
+            let xhttp_cfg = XhttpConfig {
+                path,
+                hosts,
+                extra_headers,
+                mode,
+                no_grpc_header,
+                x_padding_bytes,
+            };
+            chain.push(Box::new(XhttpLayer::new(xhttp_cfg)));
+        }
         other => {
             return Err(format!(
-                "vless: unsupported network '{other}'; valid values: tcp, ws, grpc, h2, httpupgrade"
+                "vless: unsupported network '{other}'; valid values: tcp, ws, grpc, h2, httpupgrade, xhttp"
             ));
         }
     }
@@ -1885,7 +1999,7 @@ fn default_transport_alpn(network: &str, alpn: Vec<String>) -> Vec<String> {
     }
     match network {
         "ws" => vec!["http/1.1".to_string()],
-        "grpc" | "h2" => vec!["h2".to_string()],
+        "grpc" | "h2" | "xhttp" => vec!["h2".to_string()],
         _ => alpn,
     }
 }
@@ -2500,6 +2614,10 @@ mod tests {
         );
         assert_eq!(
             default_transport_alpn("h2", Vec::new()),
+            vec!["h2".to_string()]
+        );
+        assert_eq!(
+            default_transport_alpn("xhttp", Vec::new()),
             vec!["h2".to_string()]
         );
         // Plain TCP keeps ALPN absent.
--- a/crates/meow-config/src/proxy_parser.rs
+++ b/crates/meow-config/src/proxy_parser.rs
@@ -1402,9 +1402,9 @@ fn parse_vless(
     if tls {
         use meow_transport::tls::{TlsConfig, TlsLayer};
         let sni = if servername.is_empty() {
-            server.to_string()
+            server
         } else {
-            servername
+            servername.as_str()
         };
         let mut tls_cfg = TlsConfig::new(sni);
         tls_cfg.skip_cert_verify = skip_cert_verify;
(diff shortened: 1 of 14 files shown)


crates/meow-transport/src/h2_common.rs
```diff
@@ -186,6 +193,9 @@ impl H2Stream {
 impl Drop for H2Stream {
     fn drop(&mut self) {
         self.best_effort_eos();
+        if let Some(handle) = self.conn_abort.take() {
+            handle.abort();
```
madeye: **[bug]** `H2Stream::drop` now calls `best_effort_eos()` and then immediately `abort()`s the h2 connection driver. `send_data(..., true)` only queues the EOS DATA frame; the background `Connection` future is what writes it (and any still-buffered payload) to the socket. Aborting that task drops the codec and the TCP/TLS stream without another poll, so the peer can see a connection reset instead of a clean end-of-stream. This is a regression for `network: h2` as well — `H2Layer` now attaches the abort handle via `with_conn_abort` — and it contradicts the existing `drop_sends_end_of_stream_not_reset` contract (that test still passes only because it constructs `H2Stream::new` without an abort handle). h2mux is unaffected because it does not use `with_conn_abort` on per-stream values. Aborting on the `connect()` error paths is correct; abort-on-Drop of a live stream is not.

**Suggestion:** Keep the error-path `abort()` calls. Do not abort from `H2Stream::drop`. Either let the driver run until the connection future completes after EOS (peer close / GOAWAY), or own the `JoinHandle` on a wrapper that waits for a short flush/shutdown before aborting. Add a loopback test that writes, drops an `H2Stream` built with `with_conn_abort`, and asserts the server still observes a clean body+EOS rather than a reset.

crates/meow-config/src/proxy_parser.rs
```diff
@@ -1568,9 +1568,123 @@ fn parse_vless(
...
+                    "vless: xhttp-opts.host must not be empty for proxy '{name}'"
+                ));
+            }
+            let mode = xhttp_opts
+                .and_then(|o| o.get("mode"))
+                .and_then(|v| v.as_str())
+                .unwrap_or("stream-one")
+                .to_string();
+            if !mode.is_empty()
+                && !mode.eq_ignore_ascii_case("auto")
```
madeye: **[bug]** `mode: auto` is accepted and then implemented as stream-one (the layer never branches on `mode` except to validate it). Current Xray (`transport/internet/splithttp/dialer.go`) and mihomo (`Config.EffectiveMode`) resolve `auto`/omitted mode to `packet-up`, unless REALITY is set (then `stream-one`, or `stream-up` if `downloadSettings` is present). Copied Clash/Xray configs almost always say `mode: auto` or omit it. Against a REALITY inbound that is fine; against a TLS/CDN inbound whose server is `packet-up` (or whose CDN will not stream a bidirectional POST) the client will 400/hang while claiming `auto` is supported. The omitted-mode default at line 1601 (`unwrap_or("stream-one")`) has the same mismatch versus upstream `auto` → `packet-up`.

**Suggestion:** Until packet-up/stream-up exist, reject `auto` (and document that omitted mode is an explicit stream-one default, not Xray `auto`). Alternatively implement `EffectiveMode(has_reality)` and error on the unresolved modes instead of silently substituting stream-one.

crates/meow-config/src/proxy_parser.rs
```diff
@@ -1568,9 +1568,123 @@ fn parse_vless(
...
         }
+        "xhttp" => {
+            use meow_transport::xhttp::{XhttpConfig, XhttpLayer};
+            let xhttp_opts = config.get("xhttp-opts");
+            let path = xhttp_opts
+                .and_then(|o| o.get("path"))
+                .and_then(|v| v.as_str())
+                .unwrap_or("/")
+                .to_string();
+            let hosts: Vec<String> = xhttp_opts.and_then(|o| o.get("host")).map_or_else(
```
madeye: **[bug]** When `xhttp-opts.host` is absent, `:authority` falls back to `server` only. Mihomo (`adapter/outbound/vless.go` `NewVless` xhttp branch) and Xray (`splithttp/dialer.go`) fall back to TLS `servername` / REALITY server name, then the dial address, and bracket IPv6. A typical node (`server: <ip>`, `servername: cdn.example.com`, no `xhttp-opts.host`) will therefore send `:authority` as the IP. CDNs and Host-based Xray inbounds that expect the domain will not match. IPv6 `server` values also produce an invalid `http://2001:db8::1/...` URI because they are not bracketed.

**Suggestion:** Mirror mihomo: `xhttp-opts.host` if set, else non-empty `servername`, else `server`, with IPv6 bracketed. Cover this in `vless_config_test` by asserting the built layer’s hosts, not just `proxies.contains_key`.

crates/meow-config/src/proxy_parser.rs
```diff
@@ -1568,9 +1568,123 @@ fn parse_vless(
...
+                                ));
+                            }
+                            Some((min, max))
+                        } else {
+                            return Err(
+                                "vless: x-padding-bytes array must have 2 elements".to_string()
+                            );
+                        }
+                    } else {
+                        None
```
madeye: **[bug]** If `x-padding-bytes` is present but neither a `min-max` string nor a 2-element sequence (for example YAML `x-padding-bytes: 100`, a bool, or a mapping), the parser takes the `else` arm and stores `None`. `XhttpLayer` then omits the Referer padding header. Xray servers default-validate padding in `100-1000` and answer `400 Bad Request` when it is missing (this is exactly why mihomo #2647 defaulted empty padding to `100-1000`). Absent `x-padding-bytes` correctly defaults to `(100, 1000)`; a present-but-unparseable value should not disable it.

**Suggestion:** Return a parse error for an unrecognised `x-padding-bytes` shape, same as the inverted-range and malformed-string paths. Keep `None` only as an explicit layer-level “do not pad” escape for tests.

crates/meow-transport/src/xhttp.rs
```diff
@@ -0,0 +1,401 @@
...
+        let host = self
+            .config
+            .hosts
+            .choose(&mut rand::rng())
+            .cloned()
+            .unwrap_or_else(|| "localhost".to_string());
+
+        let mut request_builder = http::Request::builder()
+            .method(http::Method::POST)
+            .uri(format!("http://{}{}", host, self.config.path));
```
madeye: **[suggestion]** The request URI is always `http://{host}{path}`, so the h2 `:scheme` pseudo-header is `http` even when the inner stream is TLS/REALITY. Mihomo `DialStreamOne` builds `url.URL{Scheme: "https", ...}` and Xray uses `https` whenever TLS or REALITY is configured. Referer is independently formatted as `https://...`, so scheme on the request and on Referer also disagree. Most raw Xray inbounds ignore `:scheme`, but CDN frontends that expect HTTPS can treat this as a fingerprint or routing miss.

**Suggestion:** Build the request URI with `https://` whenever the outer chain is TLS/REALITY (or always `https` for stream-one, matching mihomo). Keep `http://` only for a true h2c path.

crates/meow-transport/src/xhttp.rs
```diff
@@ -0,0 +1,401 @@
...
+            if let Some((min, max)) = self.config.x_padding_bytes {
+                if max > 0 {
+                    let pad_len = if min >= max {
+                        min
+                    } else {
+                        rand::rng().random_range(min..=max)
+                    };
+                    if pad_len > 0 {
+                        let padding = "X".repeat(pad_len);
+                        let separator = if self.config.path.contains('?') {
```
madeye: **[suggestion]** Padding Referer is built by appending `?x_padding=` or `&x_padding=` onto `path`. Xray `ApplyXPaddingToHeader` for `queryInHeader` does `u.RawQuery = key + "=" + paddingValue`, replacing any existing query rather than appending. For a path such as `/api?ed=1` the Referer becomes `https://host/api?ed=1&x_padding=XXX` here versus `https://host/api?x_padding=XXX` upstream. Padding extraction still finds `x_padding`, so this is compatibility-tolerant, but it is a wire-format divergence.

**Suggestion:** Parse the request URL and set the Referer query to exactly `x_padding=<padding>`, as Xray does.

crates/meow-transport/tests/xhttp_test.rs
```diff
@@ -0,0 +1,265 @@
...
+
+    let tcp = tokio::net::TcpStream::connect(addr)
+        .await
+        .expect("tcp connect");
+
+    let layer = XhttpLayer::new(XhttpConfig::default());
+    let stream = layer.connect(Box::new(tcp)).await.expect("xhttp connect");
+
+    // Drop the stream
+    drop(stream);
```
madeye, line 232: **[suggestion]** `xhttp_drop_aborts_driver` drops the stream, sleeps 50ms, and asserts nothing — it cannot fail if the driver is leaked, and it cannot catch Issue 1’s lost-EOS race. `parse_vless_xhttp_full_opts_ok` likewise only checks that the proxy name exists, so host fallback, padding defaulting, and header wiring are untested at the config layer. The VLESS round-trip mock is a raw h2 echo; it never checks `content-type`, Referer padding, or `:scheme`.

**Suggestion:** Assert the loopback server observes a clean EOS (and that the driver task has finished) after drop. Assert parsed `XhttpConfig` fields for the full-opts YAML. Have the VLESS mock (or the D1 loopback test) check Referer `x_padding` length and `:scheme`.

crates/meow-transport/src/xhttp.rs
```diff
@@ -0,0 +1,401 @@
...
+        let has_content_type = self
+            .config
+            .extra_headers
+            .iter()
+            .any(|(k, _)| k.eq_ignore_ascii_case("content-type"));
+        if !self.config.no_grpc_header && !has_content_type {
+            request_builder = request_builder.header("content-type", "application/grpc");
+        }
+
+        // Add extra custom headers.
```
madeye: **[suggestion]** Several comments restate the next statement (`// Add extra custom headers.`, `// Add Content-Type: application/grpc if not disabled and not already present.` at 108, `// Wait for send readiness bounded by OPEN_TIMEOUT.`). They do not explain a non-obvious constraint. The deferred-response comment below is the useful kind.

**Suggestion:** Drop the restating comments; keep only the ones that encode a why (deferred response / #377, padding-in-Referer, abort-on-connect-error).

madeye (changes requested):
## Summary

This PR adds a stream-one XHTTP (splithttp) transport layer, wires it into VLESS `network: xhttp` / `xhttp-opts`, and teaches `H2Stream` to abort its h2 connection driver. The stream-one happy path (POST, gRPC content-type, Referer `x_padding`, deferred response) is in the right shape and matches Xray/mihomo for the REALITY+stream-one case. The dominant risks are a connection-driver abort on `H2Stream` drop that can lose EOS/unflushed DATA (and regresses `network: h2`), host/`auto` mode fallbacks that diverge from Xray/mihomo, and padding parse that can silently disable the header Xray servers require.

## Issue counts by severity

- bugs: 4
- suggestions: 4
- nits: 0

ltsj:
Addressed all review findings in `2c3beb6`:

- `H2Stream` now sends EOS first, drains the response for up to 1 second, then aborts the driver; error paths still abort immediately. Added a loopback assertion for body + clean EOS and driver termination.
- Reject `mode: auto` instead of silently treating it as `stream-one`; omitted mode remains explicit `stream-one` while only that mode is implemented.
- Resolve an omitted host from effective `servername`, then `server`; IPv6 authority is bracketed.
- Reject unrecognized `x-padding-bytes` shapes.
- Set `:scheme` from TLS/h2c context.
- Match Xray Referer behavior by replacing the existing query with `x_padding`.
- Added assertions for scheme, authority, content type, padding length/query replacement, config fields/defaults, clean EOS, and driver termination.
- Removed restating comments.

Verification:
- XHTTP transport: 6/6
- H2 transport: 8/8
- XHTTP config integration: 5/5
- XHTTP parser field/default test: 1/1
- VLESS-over-XHTTP end-to-end: 1/1
- `cargo fmt --all -- --check`: clean
- strict clippy for `meow-transport`, `meow-config`, and `meow-proxy`, all targets/features, `-D warnings`: clean

Issue #513, opened 2026-09-07
Audit: remote DoS, traffic-policy bypass, cryptographic integrity, and provider trust-boundary defects


aimkiray:
This issue groups the findings that can terminate the process, bypass a configured egress policy, violate a cryptographic safety guarantee, or let remote provider content influence local process execution. It deliberately excludes ordinary reload consistency and performance problems, which are tracked separately.

**Audited commit:** `a663364469b3b7dc2b46d2e0f4530e50a54108ee`.

**Evidence status:** finding 1 is dynamically reproduced by a parser probe and field-boundary suite; source inspection traces HTTP rule-provider bytes to that parser. Finding 2 is dynamically reproduced through loopback proxy-chain counters. Findings 3–4 are source-confirmed static findings and are explicitly not claimed as runtime exploit reproductions. Finding 5 is dynamically confirmed from provider refresh through adapter construction and plugin launch. Finding 6 dynamically exercises runtime fallback behavior, but not a full config/API ingestion path.

P0/P1/P2 are proposed repair priorities, not CVSS ratings. P1 here includes demonstrated routing/cryptographic failures and reachable unsafe behavior; source-confirmed entries explicitly state where exploitability or deployment prerequisites remain unproven. Findings are ordered by severity within this issue.

---

### 1. [P0] mrs: a remote rule-provider can make the release process abort with a 30-byte declared-length payload

- **Symptom:** upstream MRS parsing routes both `leaves_len` and `label_bitmap_len` through `read_u64_vec`, which calls `Vec::with_capacity(len)` before checking whether required u64 words exist. A 30-byte file declaring `leaves_len = i64::MAX` triggers Rust's capacity arithmetic panic.
- **Where:** rule-provider bytes reach [`parse_upstream_ruleset_mrs`](https://github.com/meow-rs/meow-rs/blob/a663364469b3b7dc2b46d2e0f4530e50a54108ee/crates/meow-rules/src/rule_set.rs#L132-L146); [`mrs_parser.rs#L210-L219`](https://github.com/meow-rs/meow-rs/blob/a663364469b3b7dc2b46d2e0f4530e50a54108ee/crates/meow-rules/src/mrs_parser.rs#L210-L219) allocates from the declared length before proving enough input remains.
[...]

PR #516, opened 2026-09-07
fix: close five of the six audit findings in #513

aimkiray:
Fixes findings **1, 2, 3, 5 and 6** of #513, one commit each, in the issue's suggested repair order.

Finding **4** (the `from_utf8_unchecked` binary KDF contexts in VLESS encryption) is deliberately **not** in this PR — it needs a byte-oriented BLAKE3 derive-key that preserves the exact two-phase domain separation, which is a separate change. #513 should stay open for it.

## What changed

| # | Finding | Commit |
|---|---------|--------|
| 1 | **[P0]** mrs `read_u64_vec` allocates from the declared length before proving the input exists — a 30-byte remote payload aborts the release process | `a3b3dc0` |
| 2 | **[P1]** groups clone their members before `apply_dialer_proxies` rewrites the registry, so any grouped node bypasses its `dialer-proxy` chain | `593c5c7` |
| 3 | **[P1]** the VMess body `u16` record counter wraps and reuses an AEAD key/nonce pair | `ba0c805` |
| 5 | **[P1]** a remote proxy-provider's SIP003 `plugin:` reaches `Command::new` while the node is parsed | `e53648e` |
| 6 | **[P1]** unparseable proxies/groups/rules are warn-skipped while the rebuild succeeds, and a matched rule with a missing target resolves to DIRECT | `434347b` |

Two small follow-ups fix feature-gate regressions the CI clippy matrix caught in commits 3 and 5: `0b2d750` (`parse_proxy_scoped` built with an unused `origin` parameter when `ss` is off) and `e835c9e` (the new VMess nonce-budget test was not gated on `vmess`).

### Notes on the individual fixes

**1** — `read_u64_vec` now does a checked `len * size_of::<u64>()`, compares against the remaining input, and only then allocates. Covers both confirmed call sites (`leaves_len`, `label_bitmap_len`); the fields the audit showed already failing with `Truncated` are untouched.

**2** — the build order is now leaves → `apply_dialer_proxies` → groups → auto-GLOBAL, so groups clone the chained adapters.

**3** — the nonce format is unchanged and no rekey is negotiated, per the issue's constraint. The budget is enforced before seal/open and the physical connection is retired on exhaustion. Mux needs no separate case: logical streams share one body cipher and therefore one counter, so the session dies with it.

**5** — not string filtering. `parse_proxy_scoped` takes a `ProxyOrigin`, and a node whose origin is `Provider` may not name an external SIP003 plugin unless that provider sets `allow-external-plugin: true`. Default-deny, per provider, opt-in by the operator only. Built-in in-process plugins (`obfs`, `simple-obfs`, `v2ray-plugin`, `ech-tls-tunnel`) and the operator's own `proxies:` block are unaffected.

**6** — Class A per ADR-0002, following #490's precedent of turning warn into hard-fail:

- A proxy, group, or rule the parser cannot build now fails the load, and the message names the entry. ADR-0002 defers any `--strict` leniency flag to M3, so there is no flag here.
- A *matched* rule whose target the registry does not hold is refused through a non-drop `RejectAdapter`, never dialled out directly. `DIRECT` needs no registry entry — the tunnel owns a direct adapter for exactly that — and a connection no rule matches still falls through to DIRECT. The refusal is reported to the match statistics as the REJECT it now is rather than as a proxy hop that never happened (no new metric label, per ADR-0004).
- Provider documents keep the lenient behaviour: a subscription is remote input, so one unparseable node is still warn-dropped and the rest loads. Only a group left with no members at all fails.
- The provider path-containment checks (#429 / #444) move to the top of the rebuild, so a group that cannot resolve its provider slot reports the containment violation instead of its own emptiness.
- A parser rejection is the caller's fault, so the subscription and reload routes now return 400 rather than 500.

## Verification

- New: `crates/meow-config/tests/strict_config_load.rs` (6 tests), `crates/meow-tunnel/tests/rule_target_missing_refuses.rs` (5 tests), `crates/meow-proxy/tests/vmess_nonce_budget.rs`, plus per-finding regression tests in the earlier commits.
- Migrated the existing tests that pinned the old warn-skip contract (30 across `config_test`, `vless_config_test`, `config_persistence_test`, `ech_dns_test`, `api_test`) to assert on the refusal text, which names the offending entry.
- `cargo test --workspace --no-fail-fast`: 99 test binaries pass. The one failure, `meow-tunnel::raii_guard_test::close_connection_cancels_blocked_prefix_write`, is pre-existing on `a663364` and unrelated — it asserts a 16 MiB loopback prefix write stays pending, which does not hold on this host's socket buffers.
- CI gates run locally and clean: `cargo clippy --all-targets` with default features, `--all-features`, and `--no-default-features` (all `-D warnings`), `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps`, and `cargo fmt --all -- --check`.
- Docs: `website/guide/configuration.md` compatibility notes now list the hard-fail behaviour and the connection-time refusal; `website/guide/providers.md` documents `allow-external-plugin`.

### DIFF
--- a/crates/meow-api/src/routes.rs
+++ b/crates/meow-api/src/routes.rs
@@ -1047,7 +1047,14 @@ async fn apply_raw_to_tunnel(
         cache_dir,
     )
     .await
-    .map_err(|e| (StatusCode::INTERNAL_SERVER_ERROR, e))?;
+    .map_err(|e| match e {
+        // The candidate is the caller's, so a parser rejection is a bad
+        // request — the same classification `PUT /configs` uses. This covers
+        // the proxies, groups, and rules that now fail the load instead of
+        // being warn-skipped (issue #513).
+        RebuildError::Config(msg) => (StatusCode::BAD_REQUEST, msg),
+        RebuildError::Task(msg) => (StatusCode::INTERNAL_SERVER_ERROR, msg),
+    })?;
     if let Some(missing) = expected_groups
         .iter()
         .find(|name| !proxies.contains_key(name.as_str()))
@@ -1070,18 +1077,37 @@ async fn commit_raw_candidate(
     Ok(())
 }
 
+/// Why a candidate config did not turn into a route table. Callers classify
+/// the two cases differently: a rejected config is the caller's fault, a failed
+/// task is ours.
+enum RebuildError {
+    /// The parser refused the config — an unbuildable proxy, group, or rule
+    /// (issue #513), an escaping provider path, a bad rule-provider payload.
+    Config(String),
+    /// The blocking rebuild task panicked or was cancelled.
+    Task(String),
+}
+
+impl std::fmt::Display for RebuildError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Config(msg) | Self::Task(msg) => f.write_str(msg),
+        }
+    }
+}
+
 async fn rebuild_from_raw_with_resolver_async(
     raw: RawConfig,
     resolver: Arc<meow_dns::Resolver>,
     providers: HashMap<String, Arc<ProxyProvider>>,
     cache_dir: std::path::PathBuf,
-) -> Result<meow_config::RebuildResult, String> {
+) -> Result<meow_config::RebuildResult, RebuildError> {
     tokio::task::spawn_blocking(move || {
         meow_config::rebuild_from_raw_runtime(&raw, Some(resolver), &providers, Some(&cache_dir))
     })
     .await
-    .map_err(|e| format!("config rebuild task failed: {e}"))?
-    .map_err(|e| e.to_string())
+    .map_err(|e| RebuildError::Task(format!("config rebuild task failed: {e}")))?
+    .map_err(|e| RebuildError::Config(e.to_string()))
 }
 
 // ── Subscriptions ────────────────────────────────────────────────────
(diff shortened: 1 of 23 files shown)


aimkiray:
Closing in favour of #517.

Everything here was green — all eight checks passed — so this is not about the
code failing. It is about three of the five fixes changing what a working
upstream config does, which this port should not do:

- **Finding 6, first half** (fail the load on any entry it cannot build) would
  take down a whole subscription over one node mihomo skips with a warning:
  `tuic` / `wireguard` / `hysteria` v1 nodes, any protocol compiled out of the
  running binary, and above all a `RULE-SET,<name>` whose rule-provider failed
  to fetch or whose `.mrs` did not parse — a routinely transient condition.
- **Finding 5** added `allow-external-plugin` on `proxy-provider`, a key mihomo
  does not have, and without it dropped provider nodes mihomo launches fine.
- **Finding 3** retired VMess connections past 65,536 body records where
  mihomo's `wrapping_add` counter keeps them alive.

#517 keeps findings 1 and 2 unchanged, and keeps the second half of finding 6
in a reduced form: the missing-target fallback still dials DIRECT exactly as
upstream does, but it now logs a `warn!` naming the target and the rule type
instead of a `debug!`, and the match statistics report it as the DIRECT dial it
is rather than as a PROXY hop that never happened.

Findings 3, 4 and 5 remain open in #513.

Issue #489, opened 2026-09-02
dialer-proxy on provider-sourced proxies is silently ignored


aimkiray:
Found while reviewing the robustness of the recently merged PRs (#473 in particular).

## Facts

- The only consumer of `dialer-proxy` is `apply_dialer_proxies` (`crates/meow-config/src/lib.rs`), which collects `name → dialer` edges from the **static `proxies:` section of the main config only** and re-parses those entries with an injected `meow_proxy::dialer::ProxyDialer`.
- Provider payloads take a different path: `crates/meow-config/src/proxy_provider.rs:348` calls `proxy_parser::parse_proxy(raw_map, self.ipv6)` directly. That entry point hardcodes `DirectDialer` (`crates/meow-config/src/proxy_parser.rs:100`) and the provider raw-node reader has no `dialer-proxy` handling at all.
- Net effect: a subscription node carrying `dialer-proxy: some-proxy` is parsed as if the field were absent — its traffic dials **direct**, with no warning.

This is not a regression from #473 (Phase 1 explicitly scoped to static `proxies:`); it behaved identically before the merge.

## mihomo comparison

Upstream mihomo puts `DialerProxy` on `BasicOption` (`adapter/outbound/base.go`, set via `NewDialer`), the struct embedded by every outbound type, so dialer-proxy chaining applies uniformly to all outbounds — provider-sourced ones included. The dialer itself is resolved lazily by name at dial time (`component/proxydialer/byname.go`); unknown names surface as a runtime dial error rather than a parse error. meow-rs is not obligated to match that split (the static path already hard-fails unknown dialers at parse time per ADR-0002, see the post-merge review PR), but provider nodes getting *no* chaining at all leaves a real capability gap.

## Classification per ADR-0002

Class A shape: a user reading a provider subscription with `dialer-proxy` on nodes expects chained routing and silently gets direct routing — a silent misroute. However, since this is a missing feature rather than a behavior change, warn-once seems more appropriate than hard-error for the interim (hard-failing would break subscriptions that merely carry the field).

## Proposed resolution

1. **Interim hardening (cheap):** log a warn when a provider node carries `dialer-proxy`, so the silent misroute becomes visible.
2. **Phase 2 (real fix):** thread the dialer through provider parsing — collect `dialer-proxy` edges from provider payloads as well, mirroring what `apply_dialer_proxies` does for static proxies. Considerations:
[...]

PR #517, opened 2026-09-08
fix: bound mrs allocation and bind dialer-proxy by name (#513)

aimkiray:
Replaces #516. That PR closed five of the six findings in #513 and was green
on every CI check; this one carries only the parts that do not change what a
working upstream config does. Three commits, in the order #513 suggests.

## What is in

**1. `fix(rules): bound mrs reservations by the input before allocating`** (P0)

`read_u64_vec` reserved `Vec::with_capacity(len)` from a word count declared by
a remote rule-provider before proving those words existed, so `leaves_len =
i64::MAX` tripped the capacity-overflow panic — an unrecoverable termination
under the release profile's `panic = "abort"`. A 30-byte payload was enough.
The geosite `category_count` / `domain_count` reservations asked for hundreds
of gigabytes on a bogus value the same way.

Both now check the byte requirement against the remaining input with a
saturating comparison (`pos + n` can itself wrap on a hostile length) and only
then reserve. Malformed and truncated mrs return `Truncated` / `InvalidLength`
instead of aborting. This also fires on an ordinary truncated download, not
only on an attacker.

**2. `fix(config): bind dialer-proxy by name so groups cannot bypass it`** (P1)

A node reached through a proxy group silently dialled direct, past the chain
its `dialer-proxy` declared. Groups clone their members eagerly and were built
before the dialer pass replaced the registry entries, so every group kept the
pre-dialer adapter while direct rule references got the chained one — the same
node behaved differently depending on how it was selected.

The front hop is now resolved by name against a registry published when the
build finishes, the way mihomo does (`component/proxydialer/byname.go`), and
the dialer pass runs before groups are built. Late binding also makes a
group-valued `dialer-proxy` work, which eager binding could not. A cycle is
still rejected explicitly, since late binding would otherwise turn it into
unbounded recursion on the first dial.

**3. `fix(tunnel): report the DIRECT fallback a missing rule target causes`**

A matched rule whose target the registry does not hold fell back to DIRECT.
That is what mihomo does and it stays. What is fixed is that nothing said so:
the log line was a `debug!`, invisible at the default level, and the match
statistics derived their action label from the target *name* before the lookup
ran — so a rule naming `ghost-group` was counted as a PROXY hop that never
happened while the bytes left over the direct adapter. The fallback is now a
`warn!` naming the target and the rule type, and the label says DIRECT when
that is what dialled. No new label value; DIRECT / REJECT / PROXY are the
existing three. Traffic goes exactly where it went before.

## What is deliberately not in

#516 also carried findings 3, 5 and 6. All three are pulled here because each
one changes behaviour a working upstream config depends on:

- **Finding 6, first half** (fail the load on any entry it cannot build) was
  the most invasive. Upstream drops the entry with a warning and keeps loading;
  failing the whole config instead would take down a subscription over one node
  mihomo skips — including `tuic`, `wireguard` and `hysteria` v1 nodes, any
  protocol compiled out of the running binary (`--no-default-features` could
  not load a normal config at all), and, worst, a `RULE-SET,<name>` whose
  rule-provider failed to fetch or whose `.mrs` did not parse. A rule-provider
  URL being briefly unreachable is routine. Only the second half, above, is
  kept.

- **Finding 5** (keep provider content away from the SIP003 plugin launcher)
  needed `allow-external-plugin` on `proxy-provider` — a config key mihomo does
  not have — and without that key it dropped provider nodes mihomo launches
  fine. The underlying observation still stands: an external plugin becomes the
  executable handed to `Command::new` inside `ShadowsocksAdapter::new`, and
  that constructor runs while the node is parsed, on first load and on every
  provider refresh. It needs a design that does not add a field or refuse a
  node upstream accepts.

- **Finding 3** (VMess body AEAD nonce budget) is a real defect but the fix
  changed runtime behaviour: mihomo's record counter is `wrapping_add`, so a
  connection past 65,536 records keeps running there, and the budget check
  retired it here — at roughly 1 GiB per physical connection, sooner for mux,
  whose logical streams share one counter. Exploiting the reuse also needs an
  outer transport that is not encrypting. Worth fixing, but not as part of a
  compatibility-scoped PR.

Finding 4 is untouched and still open.

## Compatibility

No new config keys. No config or subscription that loads in mihomo fails here,
and no traffic mihomo would send is refused. Commit 2 adds no new config
rejection — the `dialer-proxy` self-reference, not-found and cycle errors all
predate this branch.

## Test plan

- [x] `cargo clippy --all-targets -- -D warnings` (default, `--all-features`, `--no-default-features`)
- [x] `cargo test --workspace --no-fail-fast` — one pre-existing environmental failure, `raii_guard_test::close_connection_cancels_blocked_prefix_write`, reproduces identically on a clean `a663364` checkout (a 16 MiB prefix write fits this host's loopback socket buffers); it runs in `TunnelMode::Direct` and never reaches the changed code
- [x] new `rule_target_missing_falls_back_visibly` — 6 tests pinning the DIRECT fallback, the honest counter on both resolve paths, an existing REJECT-DROP target, a registry-less `DIRECT` rule, and the uncounted no-match fall-through
- [x] `dialer_proxy_group` — a grouped node obeys the same chain a direct reference does
- [ ] real-node smoke per AGENTS.md, left to the reviewer

#513 stays open.

### DIFF
--- a/crates/meow-config/src/lib.rs
+++ b/crates/meow-config/src/lib.rs
@@ -577,13 +577,14 @@ pub fn rebuild_from_raw_with_cache_dir(
 /// a direct dial would let traffic egress from the real source path past a
 /// chain the user configured for policy/security reasons (Class A, ADR-0002).
 /// A cycle has to be caught here in particular: late binding would turn it into
-/// unbounded recursion on the first dial. The check covers cycles through group
-/// membership too — a group-valued dialer can select the chained proxy itself —
-/// not just proxy→proxy edges, which is all upstream mihomo validates.
+/// unbounded recursion on the first dial. Cycles *through group membership* are
+/// checked separately by [`reject_group_membership_cycles`], which runs after
+/// the group build where the real membership is known.
 ///
 /// Returns the applied `(proxy, dialer)` edges so the caller can re-validate
 /// dialer targets once groups have been built (a *declared* group that failed
-/// to build satisfies `dialable` but never enters the registry).
+/// to build satisfies `dialable` but never enters the registry) and run
+/// [`reject_group_membership_cycles`] against the finished registry.
 fn apply_dialer_proxies(
     proxies: &mut HashMap<SmolStr, Arc<dyn Proxy>>,
     raw_proxies: &[HashMap<String, serde_yaml::Value>],
@@ -664,77 +665,6 @@ fn apply_dialer_proxies(
         anyhow::bail!("dialer-proxy cycle detected: {}", cycle.join(", "));
     }
 
-    // A dialer may name a *group*, whose front-hop dial then selects any member
-    // — including, transitively, the proxy that declared the chain. That loop
-    // never reaches I/O: it is synchronous nested polls and exhausts the
-    // native stack on the first dial (mihomo's `validateDialerProxies` has the
-    // same blind spot — it only sees proxy→proxy edges). Reject any edge whose
-    // dialer can reach back to its source over the combined graph of dialer
-    // edges and group-membership edges.
-    //
-    // Group membership is expanded from the *declared* config: `proxies:` may
-    // name leaves or other groups (nested), `include-all-proxies` and an
-    // auto-created `GLOBAL` contain every registry name. Provider-slot members
-    // (`use:` / `include-all`) can never carry a `dialer-proxy` themselves, so
-    // they are dead ends for reachability and are ignored.
-    let all_names: std::collections::HashSet<&str> = proxies
-        .keys()
-        .map(SmolStr::as_str)
-        .chain(raw_groups.iter().map(|group| group.name.as_str()))
-        .collect();
-    let mut members_of: HashMap<&str, Vec<&str>> = HashMap::new();
-    for group in raw_groups {
-        let mut members: Vec<&str> = group
-            .proxies
-            .as_deref()
-            .unwrap_or(&[])
-            .iter()
-            .map(String::as_str)
-            .collect();
-        if group.include_all_proxies.unwrap_or(false) {
-            members.extend(all_names.iter().copied());
-        }
-        members_of.insert(group.name.as_str(), members);
-    }
-    if !raw_groups.iter().any(|group| group.name == "GLOBAL") {
-        // Auto-created GLOBAL selects over the entire registry.
-        members_of.insert("GLOBAL", all_names.iter().copied().collect());
-    }
-    let edge_map: HashMap<&str, &str> = edges
-        .iter()
-        .map(|(name, dialer)| (name.as_str(), dialer.as_str()))
-        .collect();
-    // DFS from each dialer over dialer edges and membership edges; reaching the
-    // edge's source means the first dial recurses without bound.
-    for (name, dialer) in &edges {
-        // A source that never parsed has no entry to wrap, so no chain is
-        // applied and it cannot re-enter itself — skip it rather than rejecting
-        // a config that is otherwise only missing one node.
-        if !proxies.contains_key(name) {
-            continue;
-        }
-        let mut stack: Vec<&str> = vec![dialer.as_str()];
-        let mut visited: std::collections::HashSet<&str> = std::collections::HashSet::new();
-        while let Some(node) = stack.pop() {
-            if node == name.as_str() {
-                anyhow::bail!(
-                    "proxy '{name}': dialer-proxy '{dialer}' can route back to \
-                     '{name}' through group membership, which would recurse \
-                     forever on the first dial"
-                );
-            }
-            if !visited.insert(node) {
-                continue;
-            }
-            if let Some(next) = edge_map.get(node) {
-                stack.push(next);
-            }
-            if let Some(members) = members_of.get(node) {
-                stack.extend(members.iter().copied());
-            }
-        }
-    }
-
     // Apply the edges. Order no longer matters — the front hop is resolved at
     // dial time — so nested chains need no deepest-first deferral pass.
     for (name, dialer) in &edges {
@@ -795,6 +725,101 @@ fn apply_dialer_proxies(
         }
     }
     Ok(edges)
+}
+
+/// Reject any `dialer-proxy` edge whose target can route the front-hop dial
+/// back to the chained proxy through group membership — a loop that never
+/// reaches I/O: it is synchronous nested polls and exhausts the native stack
+/// on the first dial (mihomo's `validateDialerProxies` only sees proxy→proxy
+/// edges and has the same blind spot).
+///
+/// Runs after the group build so the model matches the *built* registry:
+/// groups that failed to build contribute no membership edges, and a declared
+/// `GLOBAL` that failed to build is backstopped by the auto-created one
+/// holding every registry entry — modelling declared membership before the
+/// build would miss both.
+///
+/// Conservative over-approximations, all fail-closed: `include-all-proxies`
+/// expands to the final registry (the build itself saw a mid-pass subset);
+/// duplicate group declarations are unioned (the registry keeps the last
+/// *successful* build, not the last declaration); relay members are all
+/// treated as reachable heads (only the first member's dialer can actually
+/// fire); provider-slot members (`use:` / `include-all`) are dead ends —
+/// provider nodes never carry a `dialer-proxy`.
+fn reject_group_membership_cycles(
+    edges: &[(SmolStr, SmolStr)],
+    raw_groups: &[raw::RawProxyGroup],
+    proxies: &HashMap<SmolStr, Arc<dyn Proxy>>,
+    global_auto_created: bool,
+) -> Result<(), anyhow::Error> {
+    if edges.is_empty() {
+        return Ok(());
+    }
+    let mut members_of: HashMap<&str, Vec<&str>> = HashMap::new();
+    for group in raw_groups {
+        if !proxies.contains_key(group.name.as_str()) {
+            // Never built — its entry is absent, so it contributes no
+            // membership edges at runtime.
+            continue;
+        }
+        let mut members: Vec<&str> = group
+            .proxies
+            .as_deref()
+            .unwrap_or(&[])
+            .iter()
+            .map(String::as_str)
+            .filter(|m| proxies.contains_key(*m))
+            .collect();
+        if group.include_all_proxies.unwrap_or(false) {
+            members.extend(proxies.keys().map(SmolStr::as_str));
+        }
+        members_of
+            .entry(group.name.as_str())
+            .or_default()
+            .extend(members);
+    }
+    if global_auto_created {
+        members_of
+            .entry("GLOBAL")
+            .or_default()
+            .extend(proxies.keys().map(SmolStr::as_str));
+    }
+    // Only edges whose source actually entered the registry produce a wrapped
+    // adapter; an unparseable source can never fire its chain, so following
+    // its edge would be a phantom.
+    let edge_map: HashMap<&str, &str> = edges
+        .iter()
+        .filter(|(name, _)| proxies.contains_key(name))
+        .map(|(name, dialer)| (name.as_str(), dialer.as_str()))
+        .collect();
+    // DFS from each dialer over dialer edges and membership edges; reaching the
+    // edge's source means the first dial recurses without bound.
+    for (name, dialer) in edges {
+        if !proxies.contains_key(name) {
+            continue;
+        }
+        let mut stack: Vec<&str> = vec![dialer.as_str()];
+        let mut visited: std::collections::HashSet<&str> = std::collections::HashSet::new();
+        while let Some(node) = stack.pop() {
+            if node == name.as_str() {
+                anyhow::bail!(
+                    "proxy '{name}': dialer-proxy '{dialer}' can route back to \
+                     '{name}' through group membership, which would recurse \
+                     forever on the first dial"
+                );
+            }
+            if !visited.insert(node) {
+                continue;
+            }
+            if let Some(next) = edge_map.get(node) {
+                stack.push(next);
+            }
+            if let Some(members) = members_of.get(node) {
+                stack.extend(members.iter().copied());
+            }
+        }
+    }
+    Ok(())
 }
 
 /// Policy names resolved internally rather than declared as usable outbounds.
@@ -988,7 +1013,8 @@ fn rebuild_from_raw_impl(
     // outbound first: SelectorGroup uses its first member when no choice has
     // been stored, and sorting every registry key previously made global mode
     // default to DIRECT or an alphabetically-first quota/expiry pseudo-node.
-    if !proxies.contains_key("GLOBAL") {
+    let had_global = proxies.contains_key("GLOBAL");
+    if !had_global {
         let mut all_proxy_names: Vec<String> = proxies
             .keys()
             .map(std::string::ToString::to_string)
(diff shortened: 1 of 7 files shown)


aimkiray:
Deep review of all three commits. The redesign is correct in its mechanics — but the core commit introduces a new crash vector and a provider-fetch regression, and commit 3's upstream-parity claim is factually wrong. Details below.

## Commit 1 (`af19018`, mrs bounds): correct, but the remote-kill class is not closed

The `read_u64_vec` / `need()` / `bounded_capacity` fixes are right: both call sites now prove bytes exist before reserving, `pos + n` wrap is eliminated, `usize::try_from` covers 32-bit, and the tests exercise both rejection paths. Verified no regression for well-formed input.

However the same file still has three remote-reachable termination/amplification paths that sit *upstream* of or beside these checks:

1. **`decompress_payload` has no decompressed-size bound** (`mrs_parser.rs:132-137`). `zstd::stream::Decoder::read_to_end` grows `out` without limit — a few MB of crafted input inflates to tens of GB → alloc failure → abort. `MAX_BODY_BYTES` only caps the *compressed* wire bytes, so every new bound check validates against a buffer whose size is itself attacker-chosen. Fix: `decoder.take(MAX_DECOMPRESSED + 1).read_to_end()`, error when it exceeds the cap.
2. **`DomainSetTraversal::traverse` is unbounded recursion on remote data** (`mrs_parser.rs:248-281`). Depth ≈ input size (alternating-bitmap + equal-size labels gadget gives ~0.8×input depth). Provider fetches run on a spawned thread (`fetch_http_blocking`, ~2 MB stack) — a ~100 KB payload overflows it; larger inputs approach the 8 MB main-stack limit. Termination is guaranteed (no cycles possible in the label DAG), but depth is not bounded. Fix: explicit-stack iterative walk, or a `Result` + depth cap (legit domain sets never exceed ~127 labels).
3. **Output amplification is super-linear**: `out.push(current.clone())` per leaf × `labels.len()`-sized entries → O(input²); `DomainSetIndex::select_ones` spends one `usize` per 1-bit (64×). Bounding `out` by the header `count` plus a hard byte cap would close it.
4. Minor: `bounded_capacity` still reserves up to 8–12× remaining bytes (48 B element vs 6 B min encoding; 24 B vs 2 B) — cap by element `size_of` too, or just `Vec::new()` and let `need()` bound growth. Three `as usize` casts on i64 fields (`:174`, `:185`, `:202`) silently truncate on 32-bit — use `usize::try_from` like `read_u64_vec` does.

## Commit 2 (`11064ce`, by-name dialer): the fix works, but opens a recursion vector and a fetch regression

The per-build `ProxyRegistry` + publish-once design is clean, the peeling cycle detection is verified correct for proxy→proxy edges (including feeders — `chain_feeding_a_cycle_is_a_config_error` covers it), the fallback wrapper fails loudly, no lock is held across await, and `dialer_proxy_group.rs`'s accept-counter tests physically prove the chain. The #513 group-bypass is genuinely fixed.

**High — group-membership recursion, newly reachable.** `dialer-proxy` may now name a group, and groups are built *after* the dialer pass, so a group member is the *wrapped* adapter. If the group's transitive membership contains the proxy itself — `A` declares `dialer-proxy: G` where `G` can select `A` — the dial path `A.dial_tcp → NamedProxyDialer → G.dial_tcp → A.dial_tcp → …` never reaches I/O or yields `Pending`, so it's synchronous nested `poll` → **native stack exhaustion → abort**, not graceful OOM. Nothing catches it: the peeling loop only sees names that *declare* `dialer-proxy` (groups can't), and `relay.rs`'s `ptr_eq` only guards a group returning itself.

- The old code was accidentally immune (groups held pre-wrap members), so this is a **regression introduced by the fix**.
- `dialer-proxy: GLOBAL` is explicitly accepted (`"GLOBAL"` in `dialable`) and GLOBAL auto-contains every proxy — including the chained node. Worse, it's triggerable *after* a clean load: `PUT /proxies/GLOBAL {"name":"A"}`, a url-test/fallback member drifting onto `A`, or a health probe (`health.rs` dials through the same path).
- Mihomo has the same latent bug — upstream's `validateDialerProxies` comment admits it "cannot determine if it is caused by groups" — but that's a reason to guard, not to inherit: this PR's own comment states "a cycle has to be caught here in particular: late binding would turn it into unbounded recursion on the first dial", which the group path defeats.
- Suggested fix: in `apply_dialer_proxies`, expand each group-valued dialer's *declared* transitive membership (`raw_groups[].proxies`, nested groups recursively, `include-all*`/GLOBAL ⇒ all names) and reject when the source is in the closure — i.e., cycle detection over the combined proxy→dialer + group→member graph. Provider-slot members can't carry `dialer-proxy` so they can be ignored. A `Metadata`-carried hop budget would be a belt-and-suspenders runtime backstop.

**Medium — publish happens after in-build provider fetches.** `registry.publish` is at `lib.rs:1041`, but `prefetch_payloads` (:973) and `load_providers_prefetched` (:995) dial `download_proxy` — the *built* adapter — through `fetch_http_blocking` inside the same impl. A chained download proxy resolves its front hop against a still-unpublished registry → `dialer-proxy 'X' is not in the proxy registry` → fetch fails → cache fallback or warn-skipped provider. Bites every sync rebuild path (PUT /configs via `rebuild_from_raw_runtime`, subscription refresh, geodata rebuilds) when the first-named proxy or a provider `proxy:` override is chained; initial load is unaffected only because it prefetches with raw-parsed unchained proxies. The `proxies` map is final right after the GLOBAL auto-create (:944) — nothing inserts after — so `publish` can move up before `first_named_proxy`.

**Medium-low — `dialable` accepts declared-but-never-built group names.** A group that fails to build (all members unparseable → warn-skip at :897) still satisfies the check, so the config loads and fails at first dial instead of at parse. Cheap fix: re-verify each edge's dialer against the post-group `proxies` map (all names are known by then).

Smaller items: non-string/empty `dialer-proxy` values are silently dropped (`:604-610`) — worth a warn; `dialer-proxy: <provider-node>` is parse-rejected since provider nodes aren't in the published map (pre-existing divergence from mihomo's `ProxiesWithProviders`, fine to leave); drive-by — `load-balance` groups never receive provider `slots` (`proxy_parser.rs:2517`), so a `use:`-only load-balance group builds empty and always returns `NoProxyAvailable` — worth its own issue.

## Commit 3 (`6b0dd66`, fallback observability): mechanics verified, but the parity claim is false

Mechanics are right: `materialize_rule_match` is the single funnel for every ingress path (verified — no second resolve site), the `"DIRECT"` relabel is honest, REJECT/REJECT-DROP are real registry entries so they're not swallowed, and the six tests assert what they claim.

**But the justification is factually wrong.** mihomo's `match()` does `adapter, ok := proxies[ada]; if !ok { continue }` — it *skips* the matched rule and keeps scanning later rules; DIRECT arrives only via the no-match tail, and the reported rule is `nil`, not the ghost rule. Upstream additionally hard-fails at load on missing static rule targets (`rules[N] error: proxy [X] not found`). So "upstream mihomo falls back to DIRECT here, exactly as upstream does" in the code comment, test-file header, and assertion message codifies a divergence as if it were parity — and it can leak: `DOMAIN,ads.x,ghost-group` followed by `DOMAIN,ads.x,REJECT` is a REJECT upstream, a direct dial here. Suggest rewording to state the deviation honestly and filing a parity issue for skip-and-continue (or load-time validation), rather than letting the comment settle the question.

One more real gap: `warn!(target, rule = %…, "…")` records the ghost name as structured fields, but `LogBroadcastLayer`'s `MessageVisitor` (`log_stream.rs:102-114`) only keeps `message` — so `/logs` WebSocket consumers (the dashboard audience this fix exists for) see the warning with the target name stripped. Interpolate into the message string instead. Per-connection `warn!` without dedup matches existing precedent (per-conn logs at `tcp.rs:195/240`, per-packet at `udp.rs:148`), so non-blocking, but a once-per-`(rule_type, target)` dedup would be strictly better.

## Verdict

Commit 1: mergeable but leaves the remote-kill class open — the decompress bound is the same severity as what was fixed. Commit 2: the redesign is right, but please add the group-membership cycle guard and move `publish` before provider fetches — both are this-commit regressions. Commit 3: mergeable once the false parity claim is corrected in the comment and test docs.

aimkiray:
Pushed `51d57c1` addressing the review findings:

**Commit 2 follow-ups (this PR's own regressions):**
- `apply_dialer_proxies` now rejects any dialer edge whose target can route back to its source over the combined proxy→dialer + group→member graph — covers `A → G` where `G` (transitively, incl. nested groups, `include-all-proxies`, and the auto-created `GLOBAL`) can select `A`. That loop never reaches I/O, so the first dial was a synchronous stack overflow → abort; the pre-late-binding build was accidentally immune. mihomo's `validateDialerProxies` has the same blind spot — noted in the comment.
- `registry.publish` moved before provider prefetch/load — the map is final once GLOBAL is created, and in-build fetches dial through `download_proxy`, which may itself be a chained node whose front hop must resolve. Previously every such fetch failed on runtime rebuilds (PUT /configs, subscription refresh, geodata).
- Edges are returned and re-verified against the post-group registry, so a *declared* group that failed to build (all members unparseable) is a load error instead of a first-dial surprise.
- Malformed (non-string/empty) `dialer-proxy` values now warn instead of silently dropping.

**Commit 1 follow-ups (same remote-input class as #513):**
- `decompress_payload` bounds decompressed output at 256 MiB (`take` + one-byte overflow detection) — the HTTP cap only bounds compressed bytes, so a zstd bomb was still an abort.
- `DomainSetTraversal::traverse` rewritten iteratively (explicit stack) — depth was proportional to input size and overflowed the 2 MiB fetch-thread stack. Output additionally capped at 4M entries / 256 MiB since each leaf clones a label path (~O(input²) worst case).
- `bounded_capacity` now divides by `max(min_encoding, size_of::<T>())` so a reservation stays proportional to input bytes.
- The three `i64 as usize` casts are now `usize::try_from` → `InvalidLength` on 32-bit.

**Commit 3 follow-ups:**
- The missing-target warning interpolates the ghost target and rule type into the message — `/logs` keeps only `message`, so structured fields never reached the dashboard.
- Comments and the test-file header corrected: upstream *skips* a missing-target rule and keeps matching (`continue` in `tunnel.go`), reaching DIRECT only via the no-match tail — the current immediate-DIRECT behaviour is now documented as a deliberate deviation, not claimed as parity. Skip-and-continue is the open parity follow-up.

Tests: 7 new unit tests in `dialer_proxy_tests` (direct/nested/include-all/GLOBAL membership cycles, combined membership+dialer cycle, allowed case, malformed value), 3 new integration tests (`dialer_proxy_group.rs` — the publish-order test was verified to fail before the `publish` move), 3 new parser tests (decompression bound via `decompress_payload_bounded`, 300k-deep iterative traversal, 1M-leaf bounded fan). `cargo fmt --check`, all three clippy configs (`default` / `no-default-features` / `all-features`), `cargo doc -D warnings`, and the `meow-config`/`meow-rules`/`meow-tunnel`/`meow-proxy` suites are green.

aimkiray:
Second-pass review of `51d57c1` found two false negatives in my own group-membership check and three residual amplification surfaces in the MRS parser — all now fixed in `47252d5`.

## What the second review found

**Membership-check blind spots (now closed by moving the check post-build)**

- **Declared-but-failed `GLOBAL`**: the pre-build check skipped the auto-GLOBAL model whenever `GLOBAL` was *declared*, but a declaration that fails to build is backstopped by the auto-created all-members GLOBAL — `dialer-proxy: GLOBAL` still recursed. Verified live: `proxies: [ghost-node]` GLOBAL + `dialer-proxy: GLOBAL` now fails `-t` with the membership error.
- **Duplicate group names**: the registry keeps the last *successful* build, not the last declaration; the model now unions members across same-named declarations.
- The check now runs in `reject_group_membership_cycles` after GLOBAL creation, over the built registry (failed groups contribute no edges, member names pruned to resolvable entries, only applied edges followed). Documented conservatisms: `include-all-proxies` expands to the final registry; relay members all treated as reachable heads (only the first member's dialer can actually fire — `DialerProxyAdapter::connect_over` delegates).

**MRS amplification surfaces (now bounded)**

- `DomainSetIndex::select_ones` allocates one `usize` per bitmap 1-bit — ~64× input amplification, ~17 GiB from a sub-MB compressed payload. A well-formed trie has one terminator per node and one incoming label edge per non-root node, so `#ones ≤ labels + 1` is now enforced *before* the index is built.
- The iterative traversal stack was still unbounded (~200 M × 16 B frames from a max-size payload): added a 4 Ki depth cap (a real domain trie never exceeds ~127 labels).
- `parse_upstream_ipcidr_set` expands each 32-byte record into ~100 CIDR strings; total output is now capped like the domain set.
- `traverse` takes its bounds as parameters (`traverse_bounded`) so the caps are directly testable — new tests assert the terminator correlation and that entry/byte/depth caps actually *error* rather than amplifying.

The iterative rewrite itself was verified frame-by-frame against the recursive original (pre-order leaf emission, `current.len() == stack.len() - 1` invariant, all three early-exit mappings, cursor advance) — semantically identical.

## Real-server verification (sing-box)

Two local sing-box Shadowsocks servers (`B` front hop :18401, `A` inner :18402), meow config `A.dialer-proxy: B`, group `G` selecting `A`, http rule-provider `rp` served from a local HTTP listener:

- `curl --proxy socks5h://… http://127.0.0.1:18090/…` → returned content via `MATCH → G → A → B`; sing-box log shows `inbound/shadowsocks[B] → 127.0.0.1:18402` and `inbound/shadowsocks[A] → target` — the real two-hop chain.
- The ~20 MB `Country.mmdb`/`GeoLite2-ASN.mmdb` geodata downloads ran "via proxy 'A'" through the same chain.
- `PUT /configs` (204) triggered a rebuild whose provider prefetch produced a **new** `B → :18402` connection — the publish-before-fetch ordering is exercised end-to-end, not just in the unit test.
- `probe.example.com` (from the fetched provider) is `REJECT`ed → empty reply — provider rules are live.
- `dialer-proxy: GLOBAL` → `-t` fails with "can route back to 'A' through group membership"; `dialer-proxy: G2` where `G2` selects only `B` loads fine.

## Pre-existing notes for follow-up issues (not introduced by this PR)

- **Snapshot leak**: `ProxyRegistry` → published `Arc<HashMap>` → adapter `Arc` → `DialerTarget` → `ProxyRegistry` is a strong reference cycle — every rebuild that applies `dialer-proxy` leaks its snapshot. Bounded by rebuild frequency; fixing it needs the published map to be the same `Arc` `Config` owns, or a shared epoch registry.
- **Startup prefetch is un-chained**: `prefetch_rule_provider_payloads_async` (lib.rs:~1166) re-parses raw proxies with `DirectDialer` before any build exists, and the prefetched bytes are reused — so a rule-provider `proxy:` naming a *chained* node fetches un-chained at startup only. The publish fix cannot cover this (no registry exists yet); it is the same two-phase problem as #489.
- **Skip-and-continue parity** (from the first round) remains the deliberate divergence: meow-rs stops at the first match and dials DIRECT; mihomo `continue`s. Worth a dedicated issue.
- `load-balance` groups can't carry provider-slot members (`parse_proxy_group_inner` doesn't pass slots) — a `use:`-only load-balance group permanently reports `NoProxyAvailable`.

Verification on this branch: `fmt --check`, three-way clippy, `cargo doc -D warnings`, affected-crate tests, and the CI-mirror test list all pass; a whole-workspace `cargo test` run is green except one listener test (`client_eof_reaps…`) that flakes under parallel load on this host — passes in 0.00 s standalone and is untouched by this PR.

PR #522, opened 2026-09-10
fix(config): accept mihomo's multi-value provider header form

fengqi:
## Summary

mihomo types provider `header:` as `map[string][]string` (one field line per list entry, RFC 9110 §5.2), and both the mihomo wiki examples and subscription converters emit that list form. meow-rs typed `proxy-providers` `header` as `map[string]string`, so a mihomo-style config failed to load with `invalid type: sequence, expected a string`. Rule providers had no `header` key at all and silently ignored one.

This PR aligns both provider kinds with upstream:

- `StringOrList` untagged enum accepts both the mihomo list form and the single-string form (kept for existing meow-rs configs — a strict superset of upstream's accepted inputs, no mihomo-compatible config changes meaning).
- Headers are flattened to name-sorted `(name, value)` pairs via a **stable** sort, so same-name values keep their declared order (Go's `net/http` preserves per-name value order); a list entry repeats the field name once per value.
- Rule-provider `header` applies on initial load, prefetch, and periodic refresh; proxy-provider `header` applies at load and API-triggered refresh (proxy-providers have no periodic refresh — `interval` is parsed but not scheduled). `FetchContext` groups `download_proxy` + headers for the refresh path.
- `internal_http` no longer appends the built-in `User-Agent` when the user supplies one — previously a custom UA produced two UA field lines.
- `internal_http` **drops reserved/framing header names** (`Host`, `Connection`, `Content-Length`, `Accept-Encoding`, and the rest of the hop-by-hop set) at emission, mirroring Go `net/http`'s `reqWriteExcludeHeader` — a configured `Host:` was previously written verbatim after the request's own Host line (a duplicate-Host request-smuggling primitive), and a configured `Accept-Encoding: gzip` would return a body meow can't decode.

Rebased onto current `main` (resolves `RawProxyProvider` vs the newly added `allow_external_plugin` field).

## Deliberate divergence: single-string values still accepted (maintainer input wanted)

mihomo is **list-only**: its structure decoder rejects a single-string header value with `parse proxy provider lx error: 'header[User-Agent]' is not a slice` ([common/structure/structure.go](https://github.com/MetaCubeX/mihomo/blob/Meta/common/structure/structure.go), slice decoding).

This PR accepts **both** the list form and the single-string form (`StringOrList`). Rationale:

- meow-rs already parsed `proxy-providers` `header` as `map[string]string` before this change, so single-string configs exist in the wild; a bug-fix PR should not break them at load time.
- The accepted input is a strict superset of upstream's: every mihomo-valid config parses identically, and a single string `s` is unambiguously equivalent to `[s]` (one field line).

If maintainers prefer strict mihomo alignment (fail on single-string like upstream does), that is a small follow-up: type the field as `map<string, Vec<String>>`, drop `StringOrList`, and remove the single-string tests. Happy to switch in this PR if that is the preferred call.

## Scope check

- [x] This is a **client-side** change. It does not add server-mode features (protocol servers for remote clients, user/quota management, or anything meant to run meow-rs on the server end of a tunnel).
- [x] New config keys / behaviour follow **mihomo mainline** (same keys, formats, defaults). Upstream reference:
  - proxy-providers `Header map[string][]string`: [adapter/provider/parser.go](https://github.com/MetaCubeX/mihomo/blob/Meta/adapter/provider/parser.go)
  - rule-providers `Header map[string][]string`: [rules/provider/parse.go](https://github.com/MetaCubeX/mihomo/blob/Meta/rules/provider/parse.go)
  - default UA suppression when a user UA exists: [component/http/http.go](https://github.com/MetaCubeX/mihomo/blob/Meta/component/http/http.go)
- [x] Any deliberate divergence from mihomo is classified and documented per [ADR-0002](https://github.com/meow-rs/meow-rs/blob/main/docs/adr/0002-upstream-divergence-policy.md). The only divergence is accepting the single-string value form in addition to upstream's list-only form; it is backward compatibility for existing meow-rs configs and strictly widens accepted input (no mihomo-valid config changes meaning, no new spellings for things mihomo names). See the "Deliberate divergence" section above for the full rationale and the strict-alignment alternative.

## ADR compliance

- [x] This change does not contradict an ADR in `docs/adr/`, **or** this PR updates / supersedes the affected ADR. (Confined to `meow-config` provider parsing/fetch; no architecture touched.)
- [x] If it touches `Metadata`, `ConnectionInfo`, `UdpSession`, or the DNS cache entries: N/A — none of those types are touched.
- [x] If it touches relay code (`relay.rs`, `tcp.rs`, or their `meow-listener` call sites): N/A — no relay code touched.

## Verification

All commands run locally on macOS after the rebase + review fixes, all green:

- [x] The full "Unit + integration tests (default features)" target list from `CONTRIBUTING.md` (not just `cargo test --lib`) — 2300+ tests across the listed targets, plus `cargo test -p meow-listener --all-features --lib udp_port_53` (3 passed) and `cargo test -p meow-config` (all suites)
- [x] Protocol-specific suites for the areas touched: N/A — change is confined to `meow-config` provider config/fetch; no protocol code touched.
- [x] New behaviour has a test that fails before this change and passes after it:
  - `raw_proxy_provider_deserializes_list_header` (previously: `invalid type: sequence, expected a string`)
  - `raw_rule_provider_deserializes_list_header` / `raw_rule_provider_deserializes_single_string_header` (previously: `header` silently ignored)
  - `http_provider_new_with_multi_value_headers` (flattening + ordering)
  - `http_provider_sends_custom_headers_on_the_wire` — loopback server asserts two `User-Agent` field lines from one list entry and that the built-in default UA is not duplicated
  - `reserved_header_names_are_dropped_at_emission` — a configured `Host`/`Content-Length`/`Accept-Encoding`/`Connection` never reaches the wire; exactly one Host line with this client's own value
  - `header_values_with_crlf_are_rejected` / `header_names_with_colons_are_rejected` — the `bail!("invalid HTTP header")` guard (header-injection attempts)
  - `http_provider_sends_single_string_header_on_the_wire` — single-string form reaches the wire as one field line and replaces the built-in UA
- [x] `CHANGELOG.md` updated for user-visible changes


aimkiray (commented):
Traced the full header plumbing on this branch — the parity story checks out. `map[string][]string` is upstream's type for both provider kinds (rule-providers got theirs in mihomo #2463 / `498f81a`), the list-only rejection claim about `structure.go` is accurate, and the UA-suppression matches `component/http/http.go` post-`05fbf55`. Headers reach every fetch path: rule-provider prefetch / initial load / periodic `refresh()` / API force-refresh / PUT+subscription rebuilds, and proxy-provider load + API refresh. CRLF-in-value is already guarded at the single emitter (`internal_http.rs`), so the new surface is covered. The `StringOrList` superset is a reasonable, well-documented divergence — dropping it would break existing meow configs for zero upstream benefit.

Three small asks before merge:

1. **Filter reserved header names before emission.** Built-ins (`Host`, `Connection`, `Accept-Encoding`, `Content-Length`) are written first, then user headers verbatim — so `header: {Host: x}` emits a **duplicate `Host` line** (Go's writer drops `Host`/`Content-Length`/`Transfer-Encoding`/`Trailer`/`User-Agent` from the map write via `reqWriteExcludeHeader`). Duplicate Host is a classic request-smuggling primitive, and a configured `Accept-Encoding: gzip` yields an undecodable body. Suggest skipping hop-by-hop/reserved names at `fetch_one` (mirroring the UA replace semantics) or warn-and-skip at parse.

2. **`sort_unstable_by` → `sort_by`** in `flatten_header_map`. Go's `Header` preserves declared order for same-name values; an unstable sort doesn't contractually. One word, kills the nondeterminism.

3. **Two cheap tests**: the `bail!("invalid HTTP header")` rejection path is currently untested, and the single-string form has a parse test but no wire test.

Wording nits:
- CHANGELOG says "initial load, prefetch, and periodic refresh" — proxy-providers have no periodic refresh (`RawProxyProvider.interval` is parsed but never scheduled); the claim only holds for rule-providers.
- "One field line per list entry" isn't strictly accurate for `User-Agent` on h1: Go's request writer special-cases it to the first value only (and drops `Host`/`Content-Length`/`Trailer` from the map). meow emits all values — RFC-legal, just not h1-byte-parity.
- Narrow behavior change worth noting: on API-pushed configs, `header: {X: 123}` used to coerce to `"123"`; the untagged enum now rejects it with an opaque "did not match any variant" error. Rejection matches mihomo, so just changelog it.

Pre-existing issues surfaced while tracing (not this PR — follow-up material):
- **`RuleSetRule` holds an `Arc` snapshot**: `refresh()` replaces the provider's inner `Arc`, so refreshed ruleset content never reaches live rules. The refresh path this PR extends is currently detached from matching — worth its own issue.
- `state.rule_providers`/`state.proxy_providers` are written once at startup; PUT /configs rebuilds providers internally but never swaps the API-visible map, so manual/periodic refresh keeps using startup-era instances.
- Other upstream provider-download gaps remain: `proxy:` on proxy-providers (they always fetch direct), `size-limit`, ETag/`If-None-Match`, `global-ua`, and gzip `Accept-Encoding`.

None of the asks are large — happy to re-review quickly once they're in, ideally together with the rebase this branch already needs.

fengqi:
Thanks for the thorough trace — all three asks plus the wording nits are in 9cda554 (rebased onto current `main`, which resolves the `RawProxyProvider` conflict with the new `allow_external_plugin` field):

1. **Reserved header names filtered at emission** — `internal_http` now drops `Host`, `Connection`, `Content-Length`, `Accept-Encoding`, and the rest of the hop-by-hop set (`Transfer-Encoding`, `Trailer`, `Te`, `Upgrade`, `Keep-Alive`, `Proxy-*`) with a debug log, mirroring `reqWriteExcludeHeader`. Covered by `reserved_header_names_are_dropped_at_emission`, which asserts exactly one Host line with the client's own value and no `Content-Length: 999` on the wire.

2. **`sort_unstable_by` → `sort_by`** — done; same-name values now contractually keep their declared order.

3. **Two cheap tests** — `header_values_with_crlf_are_rejected` and `header_names_with_colons_are_rejected` cover the `bail!("invalid HTTP header")` guard, and `http_provider_sends_single_string_header_on_the_wire` covers the single-string form end-to-end (one UA line on the wire, built-in default suppressed).

Wording nits fixed in the CHANGELOG: the periodic-refresh claim is now scoped to rule-providers (proxy-provider `interval` is parsed but never scheduled), the UA multi-value emission is noted as RFC-legal but not h1-byte-parity with Go's first-value special case, and the numeric-value rejection on API-pushed configs is called out as a deliberate behavior change matching mihomo.

crates/meow-config/src/internal_http.rs
```diff
@@ -167,24 +196,41 @@ async fn fetch_one(
...
-        ua = USER_AGENT,
     );
+    if !has_custom_ua {
+        request.push_str("User-Agent: ");
+        request.push_str(USER_AGENT);
+        request.push_str("\r\n");
+    }
     for (name, value) in headers {
         if name.is_empty()
             || name.bytes().any(|b| b == b'\r' || b == b'\n' || b == b':')
```
aimkiray: **Blocking:** this validation still accepts non-token field names such as `Host `, ` Host`, and `Host\t`. Those names then miss the exact case-insensitive `is_reserved_header` check and are emitted as `Host : evil` / ` Host: evil`. RFC 9110/9112 requires a field name to be a `token` and forbids whitespace between the field name and colon; tolerant intermediaries can normalize these differently, recreating a parser-differential duplicate-Host/request-smuggling input that the new filter is intended to prevent. Go's writer similarly checks `ValidHeaderFieldName`. Please reject (do not trim/rewrite) any name containing a byte outside RFC `tchar` before reserved-name matching, and add regressions for leading/trailing SP and HTAB. The same validation will also reject the currently accepted Unicode/DEL/other separator cases. Values allowing CTLs other than HTAB is a secondary hardening opportunity.

aimkiray (changes requested):
Thanks — the follow-up addresses all of the original asks: reserved-name filtering, stable per-name value order, the rejection and single-string wire tests, and the CHANGELOG corrections. The focused `meow-config` test suite (238 tests), formatting check, and Clippy with `-D warnings` all pass locally.

There is one security-relevant bypass in the new reserved-name defense that should be closed before merge; see the inline comment. Header field names need full RFC token validation, not only CR/LF/colon rejection.

fengqi:
Good catch — fixed in 1247ac8. The name check is now the full RFC 9110 `tchar` set (mirroring `httpguts.ValidHeaderFieldName`), applied before reserved-name matching, and names are rejected as-is (never trimmed/rewritten): `"Host "`, `" Host"`, `"Host\t"`, `"Ho st"`, non-ASCII, DEL, separators, and empty all bail. The value check mirrors `ValidHeaderFieldValue` (SP/HTAB/visible/0x80+ allowed, other CTLs rejected), which closes the secondary hardening too.

Regressions added: `header_names_with_padding_whitespace_are_rejected` (leading/trailing SP, HTAB, inner SP, padded `Content-Length`), `header_names_with_non_token_bytes_are_rejected` (Unicode, DEL, `()`, backslash, `\x1f`), and `header_values_with_control_characters_are_rejected` (NUL, VT, DEL rejected; SP/HTAB round-trip as legal). `meow-config` 241 tests, fmt, three-way clippy, `cargo doc -D warnings`, and the full CI regression target list all pass locally (`smux_singbox_integration` runs in CI — no sing-box binary on this machine).

aimkiray (approved):
Approved. The follow-up now validates header names against the complete RFC 9110 `tchar` set before reserved-name matching, rejects invalid control characters in values, and adds regressions for whitespace/non-token names and CTL values. The previously requested changes are complete. Focused `meow-config` tests pass (229 tests in the current checkout), and all reported CI checks (`lint`, `test`, `features`, `msrv`, `windows`, `macos`, `tproxy`, `openwrt`) are successful. No blocking findings remain.

PR #523, opened 2026-09-11
fix: close remaining #513 findings (vmess nonce budget, vless UB, provider plugin gate, target skip)

aimkiray:
## Summary

Closes the four remaining findings from #513 after #517 merged:

- **F4 — VLESS `from_utf8_unchecked` UB** (`aead.rs`): VLESS KDF contexts are arbitrary binary (IVs, ciphertext, rekey records). `blake3::derive_key` requires `&str`, so the crate's public API cannot express the context stage over bytes — but its internal flags are documented (`DERIVE_KEY_CONTEXT`/`DERIVE_KEY_MATERIAL`). Added a local byte-oriented context-stage hash mirroring BLAKE3's chunk/tree processing, then the public hazmat `new_from_context_key` for the material stage. Output verified bit-identical to `blake3::derive_key` for valid UTF-8 at every length 0..4096 and across chunk boundaries (1023/1024/1025); invalid-UTF-8 contexts now work where the unchecked call was UB. Zero `unsafe` left.

- **F3 — VMess 16-bit nonce wrap** (`body.rs`): record nonces derive from a u16 counter that wraps to 0 after 64K records — AEAD nonce reuse under one key. Counters are now u32; `NONCE_BUDGET = 1<<16` is enforced before the wire cast, returning `io::Error` that propagates up and retires the connection. Covers both AEAD suites, both directions; mux shares the physical budget automatically. **Deliberate divergence**: mihomo/v2fly share this wrap, so a single VMess connection past ~1 GiB is severed — documented in the code and here for the changelog.

- **F5 — provider plugin trust boundary** (`raw.rs`, `proxy_provider.rs`, `proxy_parser.rs`): provider content is remote-controlled, yet an `ss` node's `plugin:` reached `Command::new`. New provider-level `allow-external-plugin` (default `false`) rejects external SIP003 plugins before adapter construction with an actionable error; explicit opt-in restores the behavior; trusted local `proxies:` unchanged; built-ins (`obfs`, `simple-obfs`, `v2ray-plugin`, feature-gated `ech-tls-tunnel`) unaffected — the classifier mirrors `ShadowsocksAdapter::new`'s dispatch exactly.

- **F6 — missing-target semantics** (`rule_ir.rs`, `match_engine.rs`, `tunnel.rs`): mihomo's `match()` does `continue` on `proxies[adapter] == nil`; meow-rs stopped at the first match and silently substituted DIRECT, so `DOMAIN,ads.x,ghost` + `DOMAIN,ads.x,REJECT` direct-dialed what upstream rejects. Both engines now take a `target_exists` predicate fed by `route.proxies` (plus the built-in `DIRECT`), warn on the skip (message-field-interpolated for /logs), and resume scanning — for the compiled IR this covers linear scans, domain-indexed scans, trie-hit tail continuation, and the lazy path's demand-driven re-entry. `materialize_rule_match` keeps a defensive fallback.

## Test plan

- [x] `cargo test -p meow-proxy` — VLESS: 6 tests (derive equivalence at every length + boundary, invalid-UTF-8, AEAD round-trip); VMess: nonce budget exhaustion test drives the 65,537th record to error in both suites and both directions, wire format unchanged
- [x] `cargo test -p meow-config` — provider gate: default rejection, opt-in reaching plugin construction, built-in plugins accepted, non-ss stray `plugin:` accepted
- [x] `cargo test -p meow-tunnel` — skip-and-continue: dead target + later REJECT → REJECT (parity case), dead target final → DIRECT tail, skipped match not counted, lazy path identical; all 82 rule-ir equivalence tests still green
- [x] Full regression bar: `cargo fmt --check`, three-way clippy `-D warnings`, `cargo doc -D warnings`, `cargo test` across the CLAUDE.md list — all green

#### Compatibility notes

- VMess connections >~1 GiB per physical connection now error at the nonce budget (upstream silently wraps). Mux'd traffic reaches it sooner; connections re-dial transparently.
- `allow-external-plugin` is a meow-rs extension — mihomo has no external SIP003 mechanism, so no upstream subscription depends on it.


aimkiray:
Second-pass review found real holes in the first cut — all fixed in `188024c`.

**IR clean-up passes were still first-match-terminates.** Under the new `continue` semantics a matched rule with a dead target no longer ends the scan, so several pruning passes became unsound — including the literal motivating example (`DOMAIN,ads.x,ghost` + `DOMAIN,ads.x,REJECT` still direct-dialed). Fixed:

- dedup, domain-shadow and covered-CIDR pruning are now keyed per adapter — a rule is dropped only when the covering rule shares its adapter (identical outcome whether the target is live or skipped);
- dead-rule truncation only applies to `MATCH,DIRECT` (the one target the match-time predicate guarantees);
- post-skip tail scans evaluate trie-owned slots directly (`EVAL_TRIE` const generic) — a second matching domain rule after a skipped hit now resolves;
- the legacy↔compiled equivalence test now runs both engines under a rejecting predicate over the indexed plan (was `|_| true` only — structurally blind to all of this).

**Subscription bypass closed.** `subscriptions:` laundered remote nodes into the trusted `proxies:` list → `parse_proxy` → `Command::new`. `parse_subscription_yaml` now drops external-SIP003 `ss` nodes before they enter the trusted path (no opt-in — a local plugin belongs in local config).

**Smaller fixes:** nonce-budget exhaustion and decrypt failures now warn in `conn.rs` (previously a silent ~1 GiB disconnect); EOF stays quiet. BLAKE3 equivalence test gained 7168/15360-byte contexts covering the ≥3-popcount final-fold branch. `allow-external-plugin` key pinned by a serde test; builtin list in docs corrected (`obfs`/`simple-obfs`, not `obfs-local`); `docs/specs/rule-ir.md` updated for the new signatures.

**Validation:** `fmt --check`, three-way clippy `-D warnings`, `cargo test -p meow-tunnel` (86 lib + all integration suites), `meow-config` (229 lib incl. new gate/subscription tests), `meow-proxy` vmess+vless suites — all green.

aimkiray:
mihomo compatibility audit (`b1ffa23`, verified against upstream `tunnel.go`/`config.go`/`shadowsocks.go` on Alpha):

**Match loop — now exact.** Upstream `match()` has *two* `continue`s: missing target (`proxies[ada] == nil`) and, for UDP, `!adapter.SupportUDP()`. The first pass implemented only the former; `target_usable` now covers both — a UDP flow matching a TCP-only target skips to the next rule instead of failing the dial. Group `support_udp` delegates to the live member, same as upstream.

**Built-in registry.** Upstream always registers DIRECT, REJECT, REJECT-DROP, COMPATIBLE, PASS, PASS-RULE (+ GLOBAL). We register the first three + auto-GLOBAL. `PASS`/`PASS-RULE`/`COMPATIBLE` stay unimplemented — a rule naming one now skips to the next rule (deterministic fall-through), the honest approximation since the Pass re-match loop isn't implemented.

**SS plugin gate vs upstream subscriptions.** mihomo implements `obfs`, `v2ray-plugin`, `gost-plugin`, `shadow-tls`, `restls`, `jls`, `kcptun` in-process; we implement `obfs`/`simple-obfs`, `v2ray-plugin`, `ech-tls-tunnel`. A subscription node carrying e.g. `shadow-tls` is rejected by the provider/subscription gate — but it could never have worked here anyway (no in-process impl → previously it attempted to spawn a nonexistent binary at dial time). The gate turns a dial failure into a load-time skip with an actionable message.

**VMess budget** — upstream wraps the u16 counter (silent nonce reuse); we sever at ~1.05 GiB/direction/connection. Documented divergence, unchanged.

**VLESS KDF** — output byte-identical by construction; zero divergence.

**`allow-external-plugin`** — meow-rs extension key; mihomo's non-strict decoder ignores it if the same config is fed upstream.

PR #536, opened 2026-09-11
fix(config): stabilize forward proxy group resolution

cedar-otter-47:
## Summary

Preserve forward proxy-group references when a referenced group is declared later and also contains unknown static members. Declared group dependencies are resolved before meow-rs falls back to its existing warn-and-skip behavior for genuinely missing members.

This also prevents a same-named leaf from being captured in place of a declared group. `include-all-proxies` now follows mihomo: it includes only name-sorted top-level `proxies:` entries, not built-ins or proxy groups, regardless of group build order.

## Scope check

- [x] This is a **client-side** change. It does not add server-mode features (protocol servers for remote clients, user/quota management, or anything meant to run meow-rs on the server end of a tunnel).
- [x] New config keys / behaviour follow **mihomo mainline** (same keys, formats, defaults). No key or format is added. See the [mihomo proxy-group documentation](https://wiki.metacubex.one/en/config/proxy-groups/#include-all-proxies).
- [x] Any deliberate divergence from mihomo is classified and documented per [ADR-0002](https://github.com/meow-rs/meow-rs/blob/main/docs/adr/0002-upstream-divergence-policy.md). This PR retains meow-rs's existing lenient handling of unknown static members; it introduces no new divergence.

## ADR compliance

- [x] This change does not contradict an ADR in `docs/adr/`, **or** this PR updates / supersedes the affected ADR.
- [x] If it touches `Metadata`, `ConnectionInfo`, `UdpSession`, or the DNS cache entries: before/after byte counts from `-Zprint-type-sizes` are in the commit body (ADR-0011). Not applicable.
- [x] If it touches relay code (`relay.rs`, `tcp.rs`, or their `meow-listener` call sites): relay buffers stay stack-allocated and hot-path allocation counts do not increase (ADR-0008). Not applicable.

## Verification

- [x] `cargo clippy -p meow-config --all-targets -- -D warnings`
- [x] `cargo clippy -p meow-config --all-targets --no-default-features -- -D warnings`
- [x] `cargo clippy -p meow-config --all-targets --all-features -- -D warnings`
- [ ] `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps`
- [ ] The full "Unit + integration tests (default features)" target list from `CONTRIBUTING.md` (not just `cargo test --lib`)
- [x] `cargo test -p meow-config`
- [x] Protocol-specific suites for the areas touched. Not applicable; this changes config-time group construction only.
- [x] New behaviour has tests that fail before this change and pass after it.
- [x] `CHANGELOG.md` updated for user-visible changes.

A manual comparison with mihomo v1.19.31 produced the same `include-all-proxies` membership for a valid control config: both included the top-level proxy and excluded the earlier proxy group.


aimkiray (commented):
## Updated code review — head `5e68a10c8f576b33cfcb4a43ce5cef46382818d6`

The dependency-aware strict/lenient resolver fixes the intended forward-reference loss. In particular, a parent waits for a declared nested group even when missing static leaf members require lenient construction. The two forward-reference tests have meaningful assertions; the same-name leaf/group test correctly uses `Arc::ptr_eq`, rather than just comparing names. `proxy_parser.rs` changes are documentation-only.

### PR-specific compatibility decision: include-all snapshot membership

One narrow behavior change remains worth making explicit. `include-all-proxies` clones `existing_proxies.values()` at build time, rather than resolving the final registry dynamically. In the new dependency-aware lenient pass (`crates/meow-config/src/lib.rs:1025-1047`), a deferred sibling may be skipped while a later include-all group builds.

Complete minimal example:

```yaml
proxies:
  - {name: leaf-d, type: socks5, server: 127.0.0.1, port: 10001}
  - {name: leaf-g, type: socks5, server: 127.0.0.1, port: 10002}
proxy-groups:
  - {name: D, type: select, proxies: [G, leaf-d]}
  - {name: X, type: select, include-all-proxies: true, proxies: [missing-x]}
  - {name: G, type: select, proxies: [leaf-g, missing-g]}
```

Source-traced outcomes:

- **Before this PR:** strict parsing stalls; the ordered final lenient pass builds D with only leaf-d, then X. X therefore includes D.
- **With this PR:** D is dependency-deferred; X builds before G and D. G then builds leniently, and D later builds with both G and leaf-d. X never gains D because it already captured its members.

D is more correct after the fix, but X loses a previously selectable sibling. Please add a test for this exact case and explicitly choose/document the snapshot semantics, or adjust scheduling if preserving that membership is required. Simply postponing every include-all group is not obviously safe: dependencies involving an include-all group can make such postponement cyclic. This needs a defined policy, not an assumed five-line fix.

I am keeping this as a compatibility decision/comment, not treating the entire forward-reference fix as incorrect.

### Test clarification

`test_missing_member_preserves_include_all_proxy_snapshot_timing` is a valid **compatibility guard** for strict-before-lenient ordering. Passing on the old implementation does **not** make it an ineffective test: preserving existing behavior is precisely its purpose. It does not promise that include-all sees the final registry, and it does not cover the deferred-sibling example above. Keep this test and add the missing scenario rather than replacing it on that basis.

Adding an ordinary acyclic forward-reference chain alone would not make the old resolver fail this timing assertion: strict passes already continue while progress is made.

### Out-of-scope follow-ups

These are existing behaviors, not newly introduced defects in this PR:

- #561 — Duplicate proxy-group names can leave a parent holding an earlier child Arc while the final registry contains a later successful duplicate. The same trace is possible with the old resolver; it is not a new `built_group_names` regression. Resolve duplicate-name policy separately.
- #562 — Pure cyclic/self-referential group declarations can be accepted after order-dependent lenient edge removal. The demonstrated issue is missing declared-graph validation, **not** proven runtime recursion or stack overflow; eagerly captured members do not by themselves establish a cyclic runtime object graph.

These replace the broader, unsupported statements that all old outcomes are preserved or that accepting a declared group cycle necessarily produces runtime recursion. Neither follow-up should block this PR as an introduced regression.

### Verification scope

This update is based on source comparison between base `0c78890e015accbfad7cce292a7e042e40cb21c2` and the head above. No new Cargo tests were executed for this update. The earlier review reported test runs, but those counts are not independently re-verified here and should not be read as fresh validation.

**Overall:** the core fix looks sound. Please make the deferred-sibling include-all behavior explicit and cover it; track duplicate-name and pure-cycle policy independently in #561 and #562. No blocking finding is justified solely by the compatibility test passing on the base revision.

cedar-otter-47:
Thanks. I made the behavior explicit in f5f901f78d054399b8ee8704d91579fb3d1a4a23.

`include-all-proxies` now follows mihomo. It includes only the top-level `proxies:` entries, sorted by name. It does not include proxy groups or built-in policies. Those must be listed explicitly in `proxies:`.

This also removes the build-order dependency. In the example above, `D` contains `[G, leaf-d]`. `X` contains `[leaf-d, leaf-g]`. Whether `D` is built before or after `X` no longer changes `X`.

The [mihomo documentation](https://wiki.metacubex.one/en/config/proxy-groups/#include-all-proxies) explicitly excludes proxy groups. Its parser does the same. It builds `AllProxies` from the top-level proxy definitions before sorting and building proxy groups.

I also compared mihomo v1.19.31 with the updated meow-rs binary. In a valid control case, both report `aggregate = [node-a]`; neither includes the earlier proxy group.

Mihomo rejects the original example because `missing-g` does not exist. meow-rs keeps its existing lenient behavior for missing members. It produces `D = [G, leaf-d]`, `G = [leaf-g]`, and `X = [leaf-d, leaf-g]`.

The new tests cover both cases.

aimkiray (approved):
Reviewed commit . The response addresses the PR-specific compatibility concern correctly:  is now materialized once from successfully parsed top-level  entries, sorted by name, and no longer depends on group build order or includes built-ins/groups. The new deferred-sibling regression test asserts  retains  while  is , which matches the stated mihomo semantics. The cycle-check path was updated to use the same top-level set, avoiding a mismatch between actual group membership and conservative validation. The existing forward-reference and same-name tests remain meaningful. I found no blocking issue in the reply commit. Out-of-scope historical items remain tracked in #561 and #562.

aimkiray:
Correction to the approval note above: shell quoting stripped Markdown backticks from that message, but the technical review is unchanged. Reviewed commit `f5f901f78d054399b8ee8704d91579fb3d1a4a23`: the response correctly materializes `include-all-proxies` once from successfully parsed top-level `proxies:` entries, sorted by name; excludes built-ins and groups; removes group-build-order dependence; updates cycle validation to use the same set; and adds the deferred-sibling regression coverage. No blocking issue found. Historical issues remain tracked in #561 and #562.

Issue #543, opened 2026-09-12
tracking: follow-ups from #514 review — provider/registry staleness, tun on-on params, refresh rebuild gaps


aimkiray:
## Summary

Follow-up findings from the third adversarial review of #544 (the `fix/514-lifecycle` branch, issue #514). None are blockers for that PR — most predate it — but they are real defects or cleanup worth tracking once it lands.

### 1. Health check: provider-sourced group members are never periodically probed — `bug`

`run_health_check_loop` (`meow-tunnel/src/health_check.rs`, member resolution ~L120) and `get_group_delay` (`meow-api/src/routes.rs` ~L1880) resolve `group.members()` names via `proxies.get(name)` against the route map. Provider-slot names (`FallbackGroup::member_names`, `fallback.rs` ~L133-146) aren't keys there → a `use:`-only group resolves zero members, wakes every interval to probe nothing, and never records `last_probed_generation`. mihomo does probe provider members.

**Fix:** resolve members through the group's `provider_slots` (or a `find_member`-style lookup) in both places.

### 2. `rule-providers` rebuilt on reload never reach the runtime registry — `bug`, `priority: medium`

`state.rule_providers` is written only at startup (`main.rs` ~L762). Consequences:

- `GET /providers/rules` reports the startup set forever.
- `PUT /providers/rules/{name}` returns 204 while refreshing an `Arc` no live matcher holds.
- Interval refresh loops (`main.rs` ~L849, now name-resolved after #544) only exist for startup-declared providers — providers added via `PUT` never get a loop.

#544 fixed the **DNS** side (`parse_dns_from_raw` swaps its loaded set into the registry). The routing side (`RuleSetRule` providers inside `rebuild_from_raw_*`) still bypasses the registry.

**Fix:** surface the rebuild's loaded provider map (e.g. on `RebuildResult`) and swap it into `state.rule_providers` on every commit; supervise interval loops for provider add/remove.

### 3. Subscription refresh rebuild loses proxy-provider and selector-store state — `bug`, `priority: medium`

`subscription_refresh::run_loop` rebuilds via `rebuild_from_raw_with_resolver` — no `proxy_providers` DashMap or `SelectorStore` like `apply_raw_to_tunnel`/`put_configs` pass. Selector groups' persisted choice snaps back on every refresh commit; provider-backed members may rebuild without provider context.

**Fix:** share the API path's rebuild entry (providers + selector store + cache dir).

[...]

Issue #535, opened 2026-09-11
anytls udp: SYNACK wait deadlocks against sing-box UoT inbound


aimkiray:
## Summary

`type: anytls` + `udp: true` deadlocks against a **sing-box** anytls inbound: every `dial_udp` hangs until the 5 s dial deadline fires, then the stream is torn down. UDP over AnyTLS never works against sing-box (works only against the vendored anytls server, which SYNACKs unconditionally).

Root cause is an ordering deadlock in `create_proxy_stream` + `AnytlsAdapter::dial_udp` (`crates/meow-proxy/src/anytls_adapter.rs:134`, `crates/meow-anytls/src/client/client.rs:72`):

1. Client opens a stream to the UoT magic destination `sp.v2.udp-over-tcp.arpa:0` and **waits for SYNACK** before returning.
2. sing-box's anytls inbound parses the destination, enters UoT mode, and **blocks reading the UoT request** before sending SYNACK.
3. Our UoT request is written lazily with the first datagram (`uot.NewLazyConn` semantics, `anytls_adapter.rs:332` `encode_uot_request`) — which only happens after `dial_udp` returns.
4. Client waits SYNACK ↔ server waits UoT request → deadlock → 5 s timeout → teardown.

## Repro

sing-box 1.13.13 server:

```json
{
  "inbounds": [{
    "type": "anytls", "listen": "127.0.0.1", "listen_port": 20006,
    "users": [{ "name": "u", "password": "pw" }],
    "tls": { "enabled": true, "server_name": "localhost",
             "certificate_path": "cert.pem", "key_path": "key.pem" }
  }],
  "outbounds": [{ "type": "direct" }]
}
```

meow config (client):

```yaml
proxies:
  - name: a
    type: anytls
    server: 127.0.0.1
    port: 20006
    password: pw
    sni: localhost
    skip-cert-verify: true
    udp: true
rules: [MATCH,a]
```

Then a SOCKS5 UDP ASSOCIATE + datagram through `mixed-port` times out on the reply read every time.

## Evidence

Client side (`anytls_rs=debug`), deterministic ~5.000 s gap between stream open and SYNACK:

```
[Client] Opened stream 4 in session, waiting for SYNACK
... (5.003 s) ...
[FrameCodec] decode: Parsed header cmd=SynAck, stream_id=4
[Session] Received SYNACK for unknown stream 4     # waiter already dropped
```

[...]

PR #544, opened 2026-09-12
fix: lifecycle and reload-consistency audit (#514)

### LINKED ISSUE #514: Audit: connection lifecycle and reload-consistency defects across encrypted relays, UDP, TUN, subscriptions, DNS, and health checks
This issue groups defects where a returned connection, background task, session, or accepted configuration can outlive the state that should own it. User-visible outcomes range from stuck UDP flows and retained sockets to dashboard configuration reporting a state that is not actually active.

**Audited commit:** `a663364469b3b7dc2b46d2e0f4530e50a54108ee`.

**Evidence status:** finding 1 is a source-confirmed lifecycle risk. Finding 2 is dynamically confirmed with serial lwIP generation probes, while the real TUN-device path remains unexercised. Finding 3 dynamically cancels a partially consumed Trojan UDP read and then reuses that connection. Finding 4 dynamically observes a SOCKS5 UDP reply-reader failure followed by continued writes. Findings 5–8 retain the dynamic control-plane evidence; finding 6 models an API-equivalent raw/routing mutation rather than issuing an actual `PUT /configs` request.

P0/P1/P2 are proposed repair priorities, not CVSS ratings. P1 lifecycle items either have a demonstrated cross-generation correctness failure or a source-confirmed unbounded-cleanup risk; P2 covers faults whose long-lived production persistence was not fully exercised. Findings are ordered by severity within this issue.

---

### 1. [P1] vmess/vless encryption: source-confirmed relay ownership gap leaves silent-upstream cleanup unbounded

- **Symptom:** both paths spawn independent read/write relays around a memory `DuplexStream`, but return only the duplex endpoint. Dropping that endpoint cannot cancel a read task blocked on a silent upstream transport.
- **Where:** VMess relay ownership is in [`vmess/conn.rs#L31-L65`](https://github.com/meow-rs/meow-rs/blob/a663364469b3b7dc2b46d2e0f4530e50a54108ee/crates/meow-proxy/src/vmess/conn.rs#L31-L65); VLESS encryption starts detached relay work at [`vless/encryption/client.rs#L382-L398`](https://github.com/meow-rs/meow-rs/blob/a663364469b3b7dc2b46d2e0f4530e50a54108ee/crates/meow-proxy/src/vless/encryption/client.rs#L382-L398).
- **Source-confirmed:** if the client/API drops the returned duplex while the upstream holds its read side open without data, FIN, or RST, the write task may exit but the read task remains blocked and retains the real transport. The outer relay's linger does not own these inner tasks.
- **Impact:** repeated connect/cancel cycles against a silent upstream can retain tasks, sockets/handles, buffers, and cryptographic state after the API connection view disappears. This is not claimed for every disconnect: FIN/RST, later data, or transport timeouts may release the task; the defect is no reliable bounded cleanup path.
- **Fix:** make the returned connection own cancellation/drop handling for both relay tasks while preserving legitimate half-close downloads. Regression: dropping it against a silent mock upstream drops the underlying transport promptly, while normal half-close downloads still complete.

### 2. [P1] tun/lwip: off→on reload can create a new lwIP generation before the old one finishes teardown

- **Symptom:** lwIP declares a one-live-stack lifecycle, but TUN shutdown aborts child pump tasks without joining them. `stop_tun` waits only for the listener parent before reload can create another `NetStack`; split halves may still retain the previous stack through their shared `BiLock`.
- **Where:** one-generation contract `meow-lwip/rust/stack.rs:18-30`, `core.rs:25-29`; split pump ownership `meow-listener/src/tun/device.rs:20-29`; child abort `tun/mod.rs:86-121`; parent-only wait `meow-tunnel/src/tunnel.rs:507-516`.
- **Dynamically confirmed:** a serial lwIP generation probe aborts split sink/stream pump tasks and observes that they are not yet reaped while the old core is still live; it can then construct a successor because no one-generation guard is enforced. A third probe sends ICMP through the old generation after a new generation is created and observes the old generation's egress delivered through the new generation's output channel, demonstrating cross-generation output-context replacement. Re-run with `cargo test -p meow-lwip --test review18_lwip_generation -- --nocapture`.
- **Impact:** the shutdown path provides no ordering guarantee between old pump/core teardown and new stack creation. Concurrent generations can race on lwIP C global netif/PCB state and output callback context; the probe directly demonstrates cross-generation egress handoff. The real TUN-device off→on path remains unexercised, and production failure timing is scheduler-dependent rather than guaranteed on every reload.
- **Fix:** abort-and-join all pump tasks, await lwIP `core_done`, and only then permit new stack creation. A regression should retain serial execution because lwIP global state is itself unsafe for parallel `NetStack::new` tests.

### 3. [P1] trojan: cancelling a partially consumed `read_packet` desyncs a subsequently reused UDP stream

- **Symptom:** `TrojanPacketConn::read_packet` consumes frame-header bytes (ATYP + addr + port + length) without storing partial-frame progress in the conn. If its future is dropped mid-frame and the connection is subsequently reused, the consumed bytes are lost and the next read starts mid-frame.
- **Where:** [`trojan.rs#L378`](https://github.com/meow-rs/meow-rs/blob/a663364469b3b7dc2b46d2e0f4530e50a54108ee/crates/meow-proxy/src/trojan.rs#L378) (`read_packet`), ATYP error at `trojan.rs:349`.
- **Reproduced:** mock TLS trojan server writes a partial header, client read cancelled by a 500 ms timeout, server completes the frame and sends a second one. Second `read_packet` errors: `Proxy error: trojan udp: unknown ATYP 0xa` (leftover payload byte of the first frame).
- **Impact:** cancelling a partially consumed read makes subsequent parsing on that connection unreliable. This matters only on paths that cancel `read_packet` and later reuse the same packet connection; relay teardown that drops the connection itself does not expose a later desync. Fleshing out the "Trojan UDP cancel-safety" item held back in #495's medium list — runtime evidence argues P1, not medium.
- **Fix:** keep a read cursor in the conn so partial header bytes survive a dropped future, or treat cancellation as fatal and close the conn.

### 4. [P2] socks5-udp: reply-task death is invisible to the NAT table — active traffic can retain a one-way session

- **Symptom:** the per-destination reply task exits silently when `read_packet` errors; nothing removes or tombstones the session. The fast path keeps writing and refreshing `last_activity_ms`, so the idle sweeper never evicts the dead session and no re-dial ever happens.
- **Where:** [`socks5_udp.rs#L233-L252`](https://github.com/meow-rs/meow-rs/blob/a663364469b3b7dc2b46d2e0f4530e50a54108ee/crates/meow-listener/src/socks5_udp.rs#L233-L252) (reply task, `while let Ok(...)`), session insert `:255-260`, fast path `:198-209`, sweeper `:130-142`.
- **Reproduced:** mock conn returns one reply then errors. After the reply task dies, 2 further client datagrams leave total dials=1 (no re-dial), total upstream writes=3 (all succeed), and no further replies reach the client.
- **Impact:** a single upstream read error leaves a one-way session that does not immediately recover. While continued client traffic refreshes `last_activity_ms`, source inspection indicates the dead session remains ineligible for idle eviction and does not redial; sends can keep succeeding while nothing returns. This is high P2 in the current evidence because the probe does not cross the sweeper/idle lifecycle; it may warrant P1 for deployments where SOCKS5 UDP/QUIC is a core workload.
- **Fix:** have the reply task remove/tombstone its session on exit (or signal death to the relay loop) so the next datagram re-dials.

### 5. [P2] subscription refresh mutates `raw_config` before rebuilding and never rolls back on failure

- **Symptom:** the refresh loop writes the fetched payload into shared `raw_config` (proxies/groups/rules replaced, `last_updated` advanced) **before** the rebuild runs; a failed rebuild is log-only. `raw_config` keeps the rejected payload while the tunnel keeps the old routing.
- **Where:** [`subscription_refresh.rs#L58-L71`](https://github.com/meow-rs/meow-rs/blob/a663364469b3b7dc2b46d2e0f4530e50a54108ee/crates/meow-app/src/subscription_refresh.rs#L58-L71) (mutation), rebuild `:74-85`, log-only failure `:95-99`.
- **Reproduced:** mock endpoint returns a payload whose rebuild fails (unknown proxy type + `dialer-proxy` to a nonexistent name — hard-fail per #490). After the failed refresh: `raw_config.proxies` contains the rejected payload, `last_updated` advanced, live routing still resolves to the pre-refresh node.
- **Impact:** `GET /configs` and the next cold start treat the rejected payload as config truth while the running tunnel disagrees — a restart silently changes behavior. This is the refresh-loop twin of #304; #314 fixed the API handlers and did not touch `subscription_refresh.rs`.
- **Fix:** apply the #314 transactional shape here — candidate on a clone, rebuild/validate first, commit raw + routing together only on success.

### 6. [P2] subscription refresh does not participate in `CONFIG_MUTATION` — silently overwrites concurrent API config changes

- **Symptom:** API mutations serialize on the static `CONFIG_MUTATION` mutex; the refresh loop takes neither that nor `config_mutation_lock`. It snapshots, awaits the fetch, then writes its payload over whatever `raw_config` holds at that moment and swaps the tunnel.
- **Where:** [`routes.rs#L94`](https://github.com/meow-rs/meow-rs/blob/a663364469b3b7dc2b46d2e0f4530e50a54108ee/crates/meow-api/src/routes.rs#L94) (`static CONFIG_MUTATION`), refresh commit `subscription_refresh.rs:58-71`.
- **Dynamically exercised:** fetch delayed 2.5 s; at +700 ms an API-equivalent mutation sets `raw_config.rules` to `MATCH,orig` and calls `update_routing`. After the refresh lands, those rules are gone from `raw_config` and routing resolves per the subscription payload again (port 443 → `DST-PORT,443,sub-node`; port 80 no longer reaches `orig`). This does not send a real `PUT /configs` request through `CONFIG_MUTATION`.
- **Impact:** refresh can overwrite concurrent raw/routing changes in both views. Source inspection shows the real API uses `CONFIG_MUTATION` while refresh does not; a real PUT-vs-refresh race test would complete that proof. #305/#314 serialized API-vs-API; refresh-vs-API remains a likely lost-update race.
- **Fix:** fetch outside the mutation lane; at commit, acquire the same lane and use a generation/version check before atomically applying subscription-owned sections plus routing. Retry or report a conflict when a concurrent update won.

### 7. [P2] PUT /configs accepts a new `dns:` section but keeps the running resolver — DNS changes need a restart

- **Symptom:** `put_configs` persists the new raw config (whole `dns:` section included) and reconciles TUN, but hands the **old** resolver to the new config; nothing rebuilds DNS.
- **Where:** [`routes.rs#L1903`](https://github.com/meow-rs/meow-rs/blob/a663364469b3b7dc2b46d2e0f4530e50a54108ee/crates/meow-api/src/routes.rs#L1903) (`Arc::clone(state.tunnel.resolver())`), `swap_config_and_reconcile_tun` `routes.rs:1782`.
- **Reproduced:** resolver pointed at mock upstream A; `PUT /configs` with `dns.nameserver` = mock upstream B → 204, `raw_config.dns.nameserver` lists B, `tunnel.resolver()` is `Arc::ptr_eq` with the pre-PUT resolver, fresh lookups still query A (A's hit count grows, B stays 0).
- **Impact:** the standard dashboard reconfiguration path (nameservers, fake-ip, sniffing, enhanced-mode) updates what `GET /configs` reports while actual resolution silently continues on the old settings until process restart — reported-vs-effective divergence.
- **Fix:** rebuild the `Resolver` from the candidate's `dns:` section inside the `CONFIG_MUTATION` lane and swap it atomically with the routing update.

### 8. [P2] health-check tasks are spawn-once — config reload neither updates, adds, nor removes them

- **Symptom:** `spawn_health_checks` runs exactly once at startup. Each loop captures its spec (url/interval/lazy) for the process lifetime and re-looks-up the group by name per tick; a vanished group just logs a skip. No reload path re-extracts specs or aborts old tasks.
- **Where:** [`main.rs#L820`](https://github.com/meow-rs/meow-rs/blob/a663364469b3b7dc2b46d2e0f4530e50a54108ee/crates/meow-app/src/main.rs#L820) (only call site), [`health_check.rs#L45-L79`](https://github.com/meow-rs/meow-rs/blob/a663364469b3b7dc2b46d2e0f4530e50a54108ee/crates/meow-app/src/health_check.rs#L45-L79) (loop + not-found skip at `:65-71`).
- **Reproduced (virtual time):** startup g1=[m-a], g3=[m-c], checks spawned for both; reload swaps in g1=[m-a2], adds g2=[m-b], removes g3. Observed: old task keeps probing recreated g1 with the **old** url/interval (m-a2 dials=4); g2 never probed (m-b dials=0); g3's task lives on skipping forever (m-c dials frozen).
- **Impact:** after any reload — dashboard PUT, subscription refresh — changed groups keep stale probe parameters, **new** url-test/fallback groups get no health checking at all (fallback then selects on alive-state nobody refreshes), and removed groups leak immortal tasks. Distinct from #485 (load-balance has no group loop) and #281 (interval:0 panic, fixed).
- **Fix:** tie health-check tasks to the config generation: abort previous tasks and re-extract/re-spawn specs on every successful routing swap (supervisor holding abort handles keyed by group name).

---

---

**Suggested repair order:** 1–2 (bounded connection/TUN teardown) → 3–4 (UDP session correctness) → 5–6 (transactional refresh and conflict handling) → 7 (DNS hot-reload contract) → 8 (health-check task supervision).

_Audited by Codex (**`gpt-6-astra`**, xhigh effort) against source at **`a663364`**; every finding is explicitly labeled as dynamically reproduced/exercised or source-confirmed above._

aimkiray:
## Summary

Closes #514 — lifecycle/reload-consistency audit, all 8 findings fixed plus three rounds of adversarial review (issue comments carry the audit trail).

**Original findings:**

- **Encrypted-relay ownership (vmess/vless)** — relay tasks now live behind `TaskedDuplex`; dropping the returned stream aborts both directions, and a read→write supervisor tears down the write loop when the read side ends. Silent-upstream drop no longer leaks transports.
- **TUN/lwIP generation gate** — a process-wide `PREVIOUS_CORE` gate plus ordered teardown (pumps aborted+joined, `TaskGroup::shutdown`, `core_done` barrier) serializes lwIP generations; the 10 s teardown timeout now **fails** the spawn instead of stacking a second core over live pcb globals. `TunHandle` carries `core_done`; `stop_tun`/`set_tun_handle` abort+await before returning.
- **Trojan UDP cancel safety** — a `read_packet` future dropped mid-read poisons the conn (length-prefix framing can no longer desync) and is re-checked after the reader mutex.
- **SOCKS5/Shadowsocks UDP dead sessions** — reply-task exit marks the session `dead`; the fast path evicts+redials, write errors evict too, and the sweeper reaps them.
- **Transactional subscription refresh** — candidate built off a clone; routing+raw commit inside `CONFIG_MUTATION`; DNS republished via shared `reconcile_dns_config`/`publish_dns`; `last_updated` stamped even on rebuild/join failure (no more 60 s re-download spin).
- **Mutation lane** — every config-mutation endpoint (PUT/PATCH, rules, groups, subscription add/delete/refresh, geodata) now goes through `CONFIG_MUTATION`; PATCH no longer races a concurrent PUT.
- **DNS hot reload** — `TunnelInner.resolver` is a shared `ResolverSlot`; `DirectAdapter`, standalone `BoundDnsServer` (`from_slot`), and every rebuilt DIRECT share it. `PUT /configs` detects dns/hosts/ipv6/geodata changes (plus proxies/groups/providers when `#tag`/`rule-set:` refs exist), parses the candidate DNS before commit (invalid dns → 400), then publishes: resolver swap + host-resolver hook + standalone server (in-place slot swap when listen addr unchanged; bind-before-teardown on rebind; failed rebind keeps the old listener **on the new generation**).
- **Health-check supervision** — `HealthCheckSupervisor` on `TunnelInner` reconciles probe tasks on every commit (startup, PUT, section endpoints, subscription refresh): added→spawned, removed→aborted, spec-changed→respawned, dead→restarted. `HealthCheckSpec` moved to meow-common, extractor to meow-config; last-wins dedup matches the builder across types.

**Review-round follow-ups folded in:**

- TUN pump pinned-read future (per-datagram `read_one`) — select! cancellation can no longer desync stream-framed UDP conns (the original Trojan poison mis-fired on every loop iteration).
- `publish_dns` keeps the old listener on rebind failure AND writes the new resolver generation into its slot.
- `rule-set:` policy keys resolve the candidate's own `rule-providers:`; loaded providers are swapped into the live registry so `PUT /providers/rules/{name}` and name-resolved interval loops stay live.
- Fake-IP input changes while `tun` stays enabled now restart the listener.
- Subscription/geodata loops capture the tunnel weakly; dropping the last `Tunnel` handle stops them.
- `expected_status` honored by periodic probes; PATCH applies fallible reload before the mode commit; `dns_inputs_equal` treats serialization failure as changed.

## Test plan

- [x] `cargo fmt --all -- --check`; `cargo clippy --all-targets` ×3 (default / no-default / all-features, `-D warnings`); `cargo doc -D warnings`
- [x] Unit+integration: api 85 (new: dns hot-swap, rebind-failure keep, rule-set candidate providers + mixed keys), tunnel 96 (health supervisor + dedup + weak-drop), listener 75, config 229, proxy 429 (relay teardown/merge tests), dns 175, common 51
- [x] Live sing-box 1.13.13 matrix (6 protocols × TCP local/gstatic h2+h1.1 + UDP echo): all pass except vmess UDP (pre-existing `NotSupported`) and anytls UDP (pre-existing deadlock → #535)
- [x] Runtime smoke: `PUT /configs` dns listen add/remove/keep, url-test group add/remove probe lifecycle, invalid `fake-ip-range` → 400

Follow-ups filed as #543; #535 is an unrelated pre-existing anytls UDP deadlock found during the smoke matrix.


aimkiray:
Superseded by the stacked split for easier review — same content, seven ordered parts:

1. #545 — outbound relay/task lifetimes (independent)
2. #546 — resolver-slot plumbing (independent, base for 3–7)
3. #548 — UDP session correctness
4. #549 — TUN lwIP core-generation lifecycle
5. #550 — mutation lane + transactional commits
6. #551 — health-check task supervision
7. #552 — DNS hot reload

Merge order: #545 and #546 can land independently; then #548 → #552 in sequence (each PR's base is the previous split branch and retargets to `main` automatically after its parent merges). The union tree is verified byte-identical to this PR's head.

PR #574, opened 2026-09-17
relay: connect_over for all TCP outbounds (vless/vmess/trojan/ss/anytls)

### LINKED ISSUE #570: relay groups can't terminate on vless/trojan/anytls/ss: connect_over unimplemented
## Summary

`relay` proxy groups cannot terminate on vless / trojan / anytls / shadowsocks (and any other protocol that doesn't implement `ProxyAdapter::connect_over`). The dial fails with:

```
relay chain failed at hop 1: <name>: connect_over not supported
```

`connect_over` is currently implemented only by `direct`, `reject`, `http`, `socks5`, and `snell` (`adapter.rs:143` default returns `NotSupported`). Since `relay` reaches every hop after the first via `connect_over` (`group/relay.rs:128-156`), a chain whose last hop is a real protocol always fails.

## Reproduction

```yaml
proxies:
  - name: front
    type: anytls            # or any dialable proxy
    server: front.example.com
    port: 443
    password: '...'
  - name: backend-vless
    type: vless
    server: backend.example.com
    port: 443
    uuid: '...'
    tls: true
proxy-groups:
  - name: CHAIN
    type: relay
    proxies: [front, backend-vless]
```

Any `/proxies/CHAIN/delay` or traffic routed to `CHAIN` fails with `connect_over not supported`.

## Expected

mihomo semantics: the last hop's full protocol handshake (TLS / VLESS header / Reality / …) runs over the stream produced by the preceding hop. `dialer-proxy` already achieves the equivalent through `NamedProxyDialer` injection (`proxy_parser.rs`): the inner proxy's own dialer dials the *inner* server address through the front proxy's stream, so `dialer-proxy: <front>` on a vless node works today — verified end-to-end.

The gap is specifically the `connect_over` trait method: adapters that got a `ProxyDialer` via injection can reuse that same machinery — `dial_stream(existing_stream)` — to implement `connect_over` without duplicating the handshake code.

## Suggested scope

Implement `connect_over` for the adapters that support `dialer-proxy` injection: vless (incl. reality/vision/encryption variants), trojan, anytls, shadowsocks, vmess, hysteria2 where applicable. Until then, configs combining `type: relay` with a real-protocol last hop are silently broken at dial time.

## Notes

- This is a long-standing capability gap since `dialer-proxy`/`relay` landed (#251 area), not a recent regression.
- Workaround: prefer `dialer-proxy` on the leaf node over `relay` groups where possible — but `relay` is still needed for N-hop chains in a single group.

aimkiray:
## Summary

Fixes relay groups (and the `dialer-proxy` fallback) so a chain can terminate on real protocol adapters, not just `http`/`socks5`/`snell`. `ProxyAdapter::connect_over` now means what mihomo's `DialContextWithDialer` means: the supplied stream already terminates at this adapter's server, so the adapter runs its **complete post-connect pipeline** — its own TLS/WS/obfs stack, then the protocol handshake — skipping only the raw socket dial.

**Semantics fix** — `connect_over` previously claimed TLS was "intentionally skipped", which was wrong for `tls: true` http/socks5 nodes in non-first positions (they sent plaintext handshakes to TLS endpoints). Both adapters now apply `wrap_tls` on the supplied stream.

**New `connect_over` impls**, each sharing its `dial_tcp` post-dial path:

- **vless** — `dial_stream` split into `dial_raw` + `wrap_stream`; `handshake_tcp` shared by `dial_tcp`/`connect_over`.
- **vmess** — `vmess_over` shared by both paths.
- **trojan** — `tls_header_over` shared.
- **shadowsocks** — `SsCore::tcp_stream_over` dispatches on `PluginKind`: plain SS, built-in obfs, v2ray-plugin, ech-tls-tunnel. External SIP003 plugins return `NotSupported` (the subprocess owns its outbound leg).
- **anytls** — adapter runs its `Arc`-shared `TlsLayer` on the supplied stream, then `Client::create_proxy_stream_on_tls` authenticates and opens the stream. The session is **unpooled** (pooling would let a later direct `dial_tcp` silently reuse the relay channel) and closed on conn drop instead of idling to the ~60s heartbeat timeout.

**Relay fixes** — `RelayGroup::connect_over` resolves group members (selector/urltest/nested relay) before invoking hops, matching `relay_tcp`; `metadata_for_proxy` types IP-literal servers as `dst_ip` + `network: Tcp` + `conn_type: Inner` instead of pseudo-domain `host`.

**Boundaries** — hysteria2 stays `NotSupported` (QUIC cannot ride a TCP stream); mux pooling is bypassed on relay-supplied streams (single-use, nothing to pool).

Closes #570.

#### Test plan

- [x] `cargo fmt --all -- --check`, clippy in default / no-default / all-features, `cargo doc` — clean
- [x] `cargo test --lib` + full CI-mirror integration list — all pass
- [x] New tests: relay member-resolution + IP-literal metadata unit tests; vmess sealed-header-over-duplex unit test; `connect_over` integration tests for vless (mock server + full `[direct → vless]` RelayGroup echo), trojan (TLS mock), ss (real ssserver, built-in obfs, external-plugin `NotSupported`), anytls (real `anytls_rs::server::Server`), http/socks5 (plain + rustls-TLS variants), hysteria2 `NotSupported`
- [x] `shadowsocks_integration` 9/9, `anytls_integration` + leak 13/13, `trojan_integration`, `vless_integration`, `smux_singbox_integration` (real sing-box) — all pass


aimkiray:
Review follow-up: three defects found in code review are fixed in the new commits (9a6654f, 870f65e, 5c7e175).

**Fixed**

- **AnyTLS unpooled session leaked on dial cancellation** — `create_proxy_stream_on_tls` only closed the session on a normal `Err`; the host's 5 s dial timeout drops the future inside the 30 s SYNACK wait, orphaning a running session + its relay leg. `SessionCloseGuard` now owns the session across the open window and spawns a detached `close()` on drop; the Err path relies on the same guard (an inline `close().await` cancelled mid-teardown would leave `is_closed` set with shutdown unfinished). The same guard covers the pool-add window in `create_new_session`.
- **`REJECT` hop caused a real outbound connection first** — `metadata_for_next_hop` skipped every empty-`addr()` member, so `[entry, REJECT]` dialed the final target before rejecting. Flattening now fails up front with `RelayHopFailed` at the member's flattened index on any non-DIRECT hop with no dialable address — before any I/O; REJECT-DROP at hop 0 no longer burns the dial timeout in its 60 s sleep.
- **Nested relay lost the global first-hop `dialer-proxy`** — the keep-wrapper guard used per-recursion `depth`/`out`, so a nested relay's first member was peeled even when globally first. The flattened output is now threaded through the recursion; a raw `DialerProxyAdapter` at the global first hop keeps its wrapper so its `dial_tcp` applies the configured front dialer.

**Coverage**: 8 new regression tests (duplex-transport cancellation EOF check; fail-closed before I/O for REJECT final/middle/nested positions, REJECT-DROP at hop 0, `[DIRECT, REJECT]`, empty-selector mid-chain; nested-relay first-hop DPA both unresolvable and working front dialer). Docs (`group-relay.md`, `dialer_proxy.rs`) synced.

**Verified**: `cargo fmt --all -- --check`; clippy `-D warnings` (meow-anytls all-features; meow-proxy default / no-default / all-features); `cargo test -p meow-proxy --lib` 453 passed; `cargo test -p meow-anytls --lib` 55 passed; `anytls_integration` 12 passed.

Known remaining limitations (all pre-existing or hand-built-graph only): UDP relay doesn't flatten (`[REJECT, x]` UDP still blackholes — REJECT advertises `support_udp`); a DPA reached *through* a resolvable group member is still unwrapped at hop 0; `resolve_proxy`/`DialerProxyAdapter::dial_tcp` recursion on cyclic hand-built graphs relies on config-side cycle rejection.

